PR #411, opened 2020-09-17
Clear cache on logout

yuval-hazaz:
Depends on #412

### DIFF
--- a/packages/amplication-client/src/User/Login.tsx
+++ b/packages/amplication-client/src/User/Login.tsx
@@ -63,24 +63,30 @@ const Login = () => {
       <span className={`${CLASS_NAME}__title`}>Sign In</span>
       <Formik initialValues={INITIAL_VALUES} onSubmit={handleSubmit}>
         <Form childrenAsBlocks>
-          <TextField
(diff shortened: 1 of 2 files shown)


iddan (approved):

packages/amplication-client/src/User/Login.tsx
```diff
@@ -63,24 +63,30 @@ const Login = () => {
...
-            type="password"
-            autoComplete="current-password"
-            minLength={8}
-          />
-          <Button type="submit">Continue</Button>{" "}
-          {REACT_APP_GITHUB_CLIENT_ID && <GitHubLoginButton />}
-          <div className={`${CLASS_NAME}__signup`}>
-            Do not have an account? <Link to="/signup">Sign up</Link>
-          </div>
+          {REACT_APP_GITHUB_CLIENT_ID ? (
```
iddan, line 66: The logic should be: show the button if `REACT_APP_GITHUB_CLIENT_ID` is defined.
Do not show login if another env var is set.

iddan (approved):

PR #412, opened 2020-09-17
Hide login text fields when using Github

yuval-hazaz:


packages/amplication-client/src/User/Login.tsx
```diff
@@ -63,24 +63,30 @@ const Login = () => {
...
-            type="password"
-            autoComplete="current-password"
-            minLength={8}
-          />
-          <Button type="submit">Continue</Button>{" "}
-          {REACT_APP_GITHUB_CLIENT_ID && <GitHubLoginButton />}
-          <div className={`${CLASS_NAME}__signup`}>
-            Do not have an account? <Link to="/signup">Sign up</Link>
-          </div>
+          {REACT_APP_GITHUB_CLIENT_ID ? (
```
iddan, line 66: The logic should be: show the button if REACT_APP_GITHUB_CLIENT_ID is defined.
Do not show login if another env var is set.

iddan:
Merging with a comment, please issue it in another PR.

iddan (approved):

packages/amplication-client/src/User/Login.tsx
yuval-hazaz: We will add support as we go, For now we only support sign in with Github and the local login is only used for dev

PR #413, opened 2020-09-17
Create sample app with new organizations

yuval-hazaz:


packages/amplication-server/src/core/organization/sampleAppData.ts
```diff
@@ -0,0 +1,244 @@
+import { Entity, EntityField } from 'src/models';
```
iddan, line 1: The name sample is misleading. Please name it default, initial, or something else

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -222,43 +234,80 @@ export class EntityService {
...
-                versionNumber: CURRENT_VERSION_NUMBER,
-                name: entity.name,
-                displayName: entity.displayName,
-                pluralDisplayName: entity.pluralDisplayName,
-                description: entity.description,
-                entityFields: {
-                  create: entity.fields
-                }
+        appId: appId
+      }
```
iddan, line 255: Add a select to only return properties you need from existing entities.

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -222,43 +234,80 @@ export class EntityService {
...
+    });
+
+    //create a map of existing entities
+    const entityMap = Object.fromEntries(
+      existingEntities.map(entity => {
+        return [`[${entity.name}]`, entity.id];
```
iddan, line 261: What is this [entityName] notation?

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -222,43 +234,80 @@ export class EntityService {
...
+
+    //replace any reference to entity (using the "[entityName]" notation) with the actual ID of the entity
+    for (const entity of entities) {
+      JSON.parse(JSON.stringify(entity), (key, value) => {
+        if (entityMap[value]) {
+          return entityMap[value];
+        } else {
+          return value;
+        }
+      });
```
iddan, line 273: I don't understand this code.

packages/amplication-server/src/core/organization/sampleAppData.ts
```diff
@@ -0,0 +1,244 @@
...
+  Entity,
+  'id' | 'createdAt' | 'updatedAt' | 'app' | 'appId' | 'fields'
+> & {
+  fields: EntityFieldData[];
+};
+
+export const SAMPLE_APP_COMMIT_MESSAGE = 'Commit Sample App Entities';
+
+//List of entities to be created in the sample app.
+//The list includes references between entities. We use [entityName] notation to allow the bulkCreate operation to map the name to the actual ID.
```
iddan, line 18: Provide an ID for user entity instead of the default created one of Prisma and use it for referencing the model.

packages/amplication-server/src/core/entity/entity.service.ts
iddan: From the context of its use I understand this complexity can be eliminated with pre-defined IDs.

yuval-hazaz: Pre-defined IDs are just "ok" for this scenario. This implementation allows us to use the "bulkCreate" in any future need without knowing the ID of the entity.

yuval-hazaz: This code stringify the object,  and runs a replace for any placeholder that exists in the map object,
After all the values were replaced it re-parses the JSON string to an object.

packages/amplication-server/src/core/organization/sampleAppData.ts
yuval-hazaz: We should leave the current implementation and not use hard-coded IDs

packages/amplication-server/src/core/entity/entity.service.ts
yuval-hazaz: The new entities can hold references to other entity using its name which is replaced with its actual ID.
It can be used for new and existing entities.
For example, if we want to reference the user entity by its ID, we can use the following code:

{ relatedEntityId: "[user]" }

iddan (approved):

PR #414, opened 2020-09-17
Center application list when less than 3

yuval-hazaz:


iddan (approved):

iddan:
Seems to me like something better be done with CSS.

PR #423, opened 2020-09-20
Password service tests

GalCegla:
Finished all tests


packages/amplication-server/src/core/account/password.service.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it('should hash a password', async () => {
+    expect(await service.hashPassword(EXAMPLE_PASSWORD)).toEqual(
+      EXAMPLE_HASHED_PASSWORD
+    );
+  });
+
+  it('should generate a password', () => {
+    expect(service.generatePassword()).toEqual('generateRandomPassword');
```
iddan, line 72: Should use a constant

iddan: If this is method is not implemented yet, you can use `it.skip`

iddan (approved):

PR #428, opened 2020-09-20
Fix Build App

iddan:
- Replace Redis queue with HTTP call to self
 - Refactor implementation


iddan:
@yuval-hazaz please review retrospectively.

PR #432, opened 2020-09-20
Pending changes icon

yuval-hazaz:


packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -36,7 +36,7 @@ export class AuthService {
     const account = await this.accountService.createAccount({
       data: {
         email,
-        firstName: '',
+        firstName: email,
```
iddan, line 39: Please remove this from this PR

PR #436, opened 2020-09-20
Server: Stack traces for unknown errors

iddan:
Correct gql resolver exceptions filter to include stacktrace for unknown errors


iddan:

PR #439, opened 2020-09-21
Fix GQL Auth

iddan:


iddan:

PR #440, opened 2020-09-21
Data Service Generator: Fix create test values

iddan:
Create null values for lookup types


iddan:

PR #443, opened 2020-09-21
Remove static users module from generated app

iddan:


iddan:

PR #446, opened 2020-09-21
Fix action service DB work

iddan:
- Await log writes to DB
 - Correct the models of the action service


iddan:

PR #450, opened 2020-09-21
Fix action service prisma select

iddan:


iddan:
@yuval-hazaz please review retrospecitvely

PR #469, opened 2020-09-22
Client: Fix update build list

yuval-hazaz:


packages/amplication-client/src/VersionControl/Build.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+const Build = ({
+  build,
+  onError,
+  open,
+}: {
+  build: models.Build;
+  onError: (error: Error) => void;
+  open: boolean;
+}) => {
```
iddan, line 33: Extract to `Props` type

iddan (commented):
Great job! One small comment

PR #496, opened 2020-09-27
Menu items style

yuval-hazaz:
- [x] MenuItem Component - Use Button instead of Primer SideNav.Link to support menu items without URLS.
- [x] Remove usage of Primer SideNav 
- [x] CommandPalette - Use MenuItem instead of custom JSX
- [x] Remove URL from signout button
- [x] Remove menu collapse to bottom

### DIFF
--- a/packages/amplication-client/src/CommandPalette/CommandPalette.tsx
+++ b/packages/amplication-client/src/CommandPalette/CommandPalette.tsx
@@ -2,11 +2,13 @@ import React, { useState, useMemo } from "react";
 // @ts-ignore
 import ReactCommandPalette from "react-command-palette";
 // @ts-ignore
-import { Icon } from "@rmwc/icon";
 import { useQuery } from "@apollo/react-hooks";
 import { gql } from "apollo-boost";
 import { History } from "history";
 import { useHistory } from "react-router-dom";
+import { SideNav } from "@primer/components";
+import MenuItem from "../Layout/MenuItem";
+
 import * as models from "../models";
 import "./CommandPalette.scss";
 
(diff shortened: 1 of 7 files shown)


iddan:
CI seems to be failing

packages/amplication-client/src/Layout/MenuItem.scss
```diff
@@ -0,0 +1,59 @@
...
+  border: none !important;
+  box-sizing: border-box;
+  transition: all var(--menu-expand-animation-duration) ease-in;
+  height: auto !important;
+
+  &__title {
+    white-space: nowrap;
+    opacity: 0;
+    transition: all var(--menu-expand-animation-duration) ease-in;
+    @include title1;
```
iddan, line 22: Include should be in the beginning of the block

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -97,7 +98,7 @@ const CommandPalette = () => {
 
   return (
     <ReactCommandPalette
-      trigger={<Icon icon="search" />}
+      trigger={<MenuItem title="Search" icon="search" />}
```
iddan, line 101: The trigger should come from props.

packages/amplication-client/src/Layout/MenuItem.tsx
```diff
@@ -1,29 +1,35 @@
...
-  const match = useRouteMatch(to);
+  const match = useRouteMatch(to || NON_URL);
+
   return (
-    <SideNav.Link
+    <Button
+      buttonStyle={EnumButtonStyle.Clear}
+      as={to ? NavLink : Button}
       onClick={onClick}
-      as={NavLink}
```
iddan, line 22: Break the component to one with the `to` prop and one without.

yuval-hazaz: I don't see many benefits from splitting this component to two.
We will have to repeat most of the code in both, unless we use two external ones that wrap the same internal component. In this case we will be left with the same internal code - Just two external interfaces. But, I don't think the current interfaces is confusing in any way...

PR #523, opened 2020-09-29
Refactor - entity version clone data

yuval-hazaz:
Extract the logic from EntityService.CreateVersion to a new function EntityService.CloneVersionData

The new function will be used in these cases:
- Commit - to copy data from version 0 to the new version
- Discard Changes - to copy data from the last version to version 0

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -16,7 +16,7 @@ import head from 'lodash.head';
 import last from 'lodash.last';
 import omit from 'lodash.omit';
 import difference from '@extra-set/difference';
-import { isEmpty } from 'lodash';
+import { isEmpty, pick } from 'lodash';
 import {
   Entity,
   EntityField,
@@ -640,6 +640,13 @@ export class EntityService {
       omit(field, ['entityVersionId', 'id'])
     );
 
+    const names = pick(sourceVersion, [
+      'name',
+      'displayName',
+      'pluralDisplayName',
+      'description'
+    ]);
+
     //update the target version with its fields, and the its parent entity
     targetVersion = await this.prisma.entityVersion.update({
       where: {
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -607,7 +576,88 @@ export class EntityService {
...
+        id: targetVersionId
+      },
+      data: {
+        entity: {
+          update: {
+            name: sourceVersion.name,
+            displayName: sourceVersion.displayName,
+            pluralDisplayName: sourceVersion.pluralDisplayName,
+            description: sourceVersion.description
+          }
```
iddan, line 655: Code can be shortened using lodash's pick

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -607,7 +576,88 @@ export class EntityService {
...
         },
+        name: sourceVersion.name,
+        displayName: sourceVersion.displayName,
+        pluralDisplayName: sourceVersion.pluralDisplayName,
+        description: sourceVersion.description,
```
iddan, line 660: Since these fields are identical to the once you are passing to entity update maybe better save them to a variable and use it twice.

iddan (changes requested):
Please update the test to reflect the logical change you've made.

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -607,7 +576,89 @@ export class EntityService {
...
+    //update the target version with its fields, and the its parent entity
+    targetVersion = await this.prisma.entityVersion.update({
+      where: {
+        id: targetVersionId
+      },
+      data: {
+        entity: {
+          update: {
+            ...names
+          }
```
iddan, line 659:
```suggestion
          update: names
```

iddan (approved):

PR #525, opened 2020-09-29
Server: Docker Build generated app

iddan:
- Create `amplication-container-builder`
   - Provide integration with Docker
   - Provide integration with Cloud Build
   - Provide integration with Nest.js 
 - Add run method to ActionService to abstract creation and completion of actions
 - Save generated code as `.tar.gz` archive for sending to docker build
 - Add a step in the build action for building the docker image using `ContainerBuilderService`
 - Cleanup build service tests
 - Create `app-node` Docker base image
 - Use `app-node` when building the generated app

### Todo
 - [x] Fix sending tarball to Cloud Build. May need to use .tar.gz instead.
 - [x] Improve performance of Cloud Build


terraform/envs/production/terraform.tfvars
```diff
@@ -33,4 +33,7 @@ google_cloudbuild_trigger_filename = "production.cloudbuild.yaml"
 google_cloudbuild_trigger_name = "version-manual-deploy"
 github_owner = "amplication"
 github_name = "amplication"
-github_tag = "v.+"
\ No newline at end of file
+github_tag = "v.+"
+
+# Apps Cloud Build
+apps_gcp_project_id = "amplication-production"
```
yuval-hazaz, line 39: I think the data on the deployment environment should be taken from a list in the DB. initially it will be a table with a single record (different between prod, staging, and dev) - but in the future we will be able to have multiple projects and select the right one dynamically.

terraform/envs/dev/terraform.tfvars
```diff
@@ -32,4 +32,7 @@ show_ui_elements = ""
 
 # Storage
 bucket = "amplication-artifacts"
-default_disk = "gcs"
\ No newline at end of file
+default_disk = "gcs"
+
+# Apps Cloud Build
+apps_gcp_project_id = "amplication"
```
yuval-hazaz, line 38: I think the data on the deployment environment should be taken from a list in the DB. initially it will be a table with a single record (different between prod, staging, and dev) - but in the future we will be able to have multiple projects and select the right one dynamically.

packages/amplication-server/package.json
```diff
@@ -36,6 +36,7 @@
   "dependencies": {
     "@codebrew/nestjs-storage": "^0.1.4",
     "@extra-set/difference": "^2.1.52",
+    "@google-cloud/cloudbuild": "^2.0.3",
```
yuval-hazaz, line 39: The deploy service should be created in a separate package consumed by amplication-server
This new service should use a new interface with different implementations for different cloud providers (currently only with GCP)

yuval-hazaz (changes requested):

terraform/envs/production/terraform.tfvars
iddan: I do not agree. This kind of configuration should come from the environment. If we would like to have multiple ones we can provide either by providing a list in the environment or multiple variables

terraform/envs/dev/terraform.tfvars
iddan: Duplicate, answered above.

packages/amplication-server/package.json
iddan: It is implemented that way. I'll extract the code to another package.

yuval-hazaz (approved):
A M A Z I N G

PR #533, opened 2020-09-30
Entity List style

yuval-hazaz:
- [x] remove inactive 'pending changes' icon
- [x] use UserAndTime instead of UserAvatar

### DIFF
--- a/packages/amplication-client/src/Entity/EntityListItem.tsx
+++ b/packages/amplication-client/src/Entity/EntityListItem.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useContext, useState, useMemo } from "react";
+import React, { useCallback, useContext, useState } from "react";
 import { gql } from "apollo-boost";
 import { useMutation } from "@apollo/react-hooks";
 import * as models from "../models";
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/Entity/EntityListItem.tsx
```diff
@@ -95,7 +84,7 @@ export const EntityListItem = ({
       />
       <DataGridRow navigateUrl={`/${applicationId}/entities/${entity.id}`}>
         <DataTableCell className="min-width">
-          <LockStatusIcon enabled={Boolean(entity.lockedByUser)} />
+          {Boolean(entity.lockedByUser) && <LockStatusIcon enabled />}
```
iddan, line 87: What does this change mean?

yuval-hazaz: The component have two states - enabled (orange) and "not enabled" (gray).
This behavior is still used in the app control toolbar when editing an entity. 

In the entity list, it was decided not to show the icons at all for entities without changes - so instead of showing the icon in the "disabled" state we are not showing it at all

PR #542, opened 2020-10-01
Permissions panel collapse

yuval-hazaz:
Disable manual collapse for permissions panel.
The panel is always open when the permission is set to Granular


packages/amplication-client/src/Components/PanelCollapsible.tsx
```diff
@@ -8,8 +8,14 @@ import "./PanelCollapsible.scss";
 
 type Props = {
   onCollapseChange?: (open: boolean) => {};
+  /**Whether the panel is initially open or not */
   initiallyOpen?: boolean;
+  /**When true the user cannot collapse manually, and the collapse button is hidden  */
+  disableManualCollapse?: boolean;
+  /**Whether the panel collapse functionality is enabled. When false, only the header of the panel is visible and the body is hidden */
+  /**By settings initiallyOpen to True, and disableManualCollapse to True, the developer can control the collapse state from the parent component by changing enableCollapse */
```
iddan, line 16: SHould be in the same doc comment.
```suggestion
  /**
   * Whether the panel collapse functionality is enabled. When false, only the header of the panel is visible and the body is hidden.
   * By settings initiallyOpen to True, and disableManualCollapse to True, the developer can control the collapse state from the parent component by changing enableCollapse
   */
```

packages/amplication-client/src/Permissions/EntityPermissionAction.tsx
```diff
@@ -192,8 +192,9 @@ export const EntityPermissionAction = ({
 
   return (
     <PanelCollapsible
-      initiallyOpen={isOpen}
+      initiallyOpen
       enableCollapse={isOpen}
```
iddan, line 196: Missed that previously but naming should be `collapseEnabled`

packages/amplication-client/src/Permissions/EntityPermissionAction.tsx
```diff
@@ -192,8 +192,9 @@ export const EntityPermissionAction = ({
...
+      disableManualCollapse
```
iddan, line 197: Rename to `manualCollapseDisabled`

iddan (commented):
Good job, a few comments.

iddan (approved):

PR #544, opened 2020-10-01
Update images

yuval-hazaz:


packages/amplication-client/src/Application/ApplicationLayout.tsx
```diff
@@ -162,7 +162,6 @@ function ApplicationLayout({ match }: Props) {
           <Switch>
             <Route exact path="/:application/" component={ApplicationHome} />
             <Route
-              exact
```
iddan, line 165: Not sure how does this relate to the PR

packages/amplication-client/src/VersionControl/PendingChanges.tsx
```diff
@@ -36,6 +38,16 @@ type Props = {
 const PendingChanges = ({ match }: Props) => {
   const { application } = match.params;
   useBreadcrumbs(match.url, "Pending Changes");
+  const history = useHistory();
+
+  const stateMatch = useRouteMatch<{ state: string }>(
```
iddan, line 43: Rename to `match`

packages/amplication-client/src/VersionControl/PendingChanges.tsx
```diff
@@ -36,6 +38,16 @@ type Props = {
...
+    "/:application/pending-changes/:state"
+  );
+
+  let state: string | null = null;
+  if (stateMatch) {
+    state = stateMatch.params.state;
+  }
```
iddan, line 50:
```suggestion
  const state = stateMatch ? stateMatch.params.state : null;
```

packages/amplication-client/src/VersionControl/PendingChanges.tsx
```diff
@@ -65,6 +77,18 @@ const PendingChanges = ({ match }: Props) => {
     setDiscardDialogOpen(!discardDialogOpen);
   }, [discardDialogOpen, setDiscardDialogOpen]);
 
+  const handleCommitDialogCompleted = useCallback(() => {
```
iddan, line 80:
```suggestion
  const handleCommitDialogComplete = useCallback(() => {
```

packages/amplication-client/src/VersionControl/PendingChanges.tsx
```diff
@@ -24,6 +25,7 @@ import useBreadcrumbs from "../Layout/use-breadcrumbs";
 
 const CLASS_NAME = "pending-changes";
 const POLL_INTERVAL = 2000;
+const STATE_COMMIT_DONE = "done";
```
iddan, line 28:
```suggestion
const COMMIT_STATE_DONE = "done";
```

iddan: Can you use a value from an enum from the server?

iddan (approved):

iddan:
@yuval-hazaz please issue my comments in another PR

packages/amplication-client/src/VersionControl/PendingChanges.tsx
yuval-hazaz: it has nothing to do with the server, It is literally a "client side" state that only describe the state of the page in order to decide what image to show when the list is empty

packages/amplication-client/src/Application/ApplicationLayout.tsx
yuval-hazaz: See below

PR #552, opened 2020-10-03
Create entity field by name

yuval-hazaz:


packages/amplication-server/src/core/entity/entity.resolver.ts
```diff
@@ -235,6 +236,17 @@ export class EntityResolver {
     return this.entityService.createField(args, user);
   }
 
+  @Mutation(() => EntityField, {
+    nullable: true,
+    description: undefined
+  })
+  async createEntityFieldByDisplayName(
```
iddan, line 243: It would be simpler to make `createEntityField()` to take everything optional except the Display Name and consolidate the two code paths.

yuval-hazaz: I think it should be very clear to the user what is about to happen when he calls the API.
By checking individuals properties for NULL it can be confusing what is expected when some are NULL while others have values. 
The new function sets all properties (required, searchable, name, data type...) and merging the code path would introduce unneeded complexity

iddan (approved):

PR #556, opened 2020-10-04
Keyboard Shortcuts

yuval-hazaz:
Use Ctrl/Command+Enter on modals to submit-
- [x] New App
- [x] New Entity
- [x] Commit
- [x] Build

Use Ctrl/Command+Shift+G to navigate to "Pending Changes"

### DIFF
--- a/packages/amplication-client/src/App.tsx
+++ b/packages/amplication-client/src/App.tsx
@@ -1,5 +1,7 @@
 import React, { useEffect } from "react";
 import { Route, Switch, useHistory } from "react-router-dom";
+import { configure } from "react-hotkeys";
+
 import ApplicationLayout from "./Application/ApplicationLayout";
 import Login from "./User/Login";
 import Signup from "./User/Signup";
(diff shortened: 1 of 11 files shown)


packages/amplication-client/src/App.tsx
```diff
@@ -1,5 +1,7 @@
 import React, { useEffect } from "react";
 import { Route, Switch, useHistory } from "react-router-dom";
+import { configure } from "react-hotkeys";
```
iddan, line 3: It would be more clear with namespace import
```suggestion
import * as reactHotkeys from "react-hotkeys";
```

packages/amplication-client/src/Application/NewApplication.tsx
```diff
@@ -47,6 +48,10 @@ const FORM_SCHEMA = {
 };
 const CLASS_NAME = "new-application";
 
+const keyMap = {
+  SUBMIT: ["ctrl+enter", "command+enter"],
+};
```
iddan, line 53: This keyMap seems to reappear in multiple files. Please share the code.

packages/amplication-client/src/Application/NewApplication.tsx
```diff
@@ -103,30 +108,35 @@ const NewApplication = () => {
         isInitialValid={false}
       >
         {(formik) => {
+          const handlers = {
+            SUBMIT: formik.submitForm,
+          };
           return (
             <Form>
```
iddan, line 115: Since this code repeats, maybe it's better to create a component that wraps Form and HotKeys together.

iddan (approved):
A few small comments, please fix retroactively

packages/amplication-client/src/Application/NewApplication.tsx
yuval-hazaz: #573

PR #561, opened 2020-10-06
Close sidebar with Esc shortcut

yuval-hazaz:


packages/amplication-client/src/Layout/SidebarHeader.tsx
```diff
@@ -11,32 +14,44 @@ type Props = {
...
-  }
+  }, [history, backUrl]);
+
+  const handlers = {
+    SUBMIT: goBack,
+  };
 
   return (
     <DrawerHeader className="side-bar__header">
+      <GlobalHotKeys keyMap={keyMap} handlers={handlers} />
```
iddan, line 42: Why GlobalHotKeys and not HotKeys here?

iddan (approved):

iddan:
@yuval-hazaz please review my comment

packages/amplication-client/src/Layout/SidebarHeader.tsx
yuval-hazaz: There is an issue when using GlobalHotKeys together with HotKeys so we will need to use GlobalHotKeys all the way. 
I created feature request #573

PR #562, opened 2020-10-06
Command palette shortcut & MenuItem tooltip

yuval-hazaz:
- [x] - remove Children from MenuItem (not in use)
- [x] - add tooltip to MenuItem
- [x] - add shortcuts to command palette


packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -107,6 +107,7 @@ const CommandPalette = ({ trigger }: Props) => {
       closeOnSelect
       showSpinnerOnSelect={false}
       theme={THEME}
+      hotKeys={["command+shift+p", "ctrl+shift+p"]}
```
iddan, line 110: Extract to constant

iddan (approved):
Great work @yuval-hazaz! Please review my comment retrospectively

PR #575, opened 2020-10-07
Added test to check Record not Found in entity service

GalCegla:
tested thrown error 'Record not Found' for deleteEntityPermissionField function, in entity.service.ts

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.spec.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.spec.ts
@@ -3,7 +3,8 @@ import {
   FindOneEntityFieldArgs,
   SortOrder,
   FindOneEntityVersionArgs,
-  EntityVersionCreateArgs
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/entity/entity.service.spec.ts
```diff
@@ -20,6 +21,8 @@ const EXAMPLE_CURRENT_ENTITY_VERSION_ID = 'currentEntityVersionId';
 const EXAMPLE_LAST_ENTITY_VERSION_ID = 'lastEntityVersionId';
 const EXAMPLE_LAST_ENTITY_VERSION_NUMBER = 4;
 
+const EXAMPLE_ACTION: EnumEntityAction = 'View';
```
iddan, line 24:
```suggestion
const EXAMPLE_ACTION = EnumEntityAction.View;
```

iddan:
Thank you!

PR #589, opened 2020-10-10
Command palette

yuval-hazaz:
- [x] new style
- [x] prioritzie current app results


packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -33,13 +43,43 @@ export class NavigationCommand implements Command {
...
+    public readonly showAppData: boolean,
+    public readonly appName?: string,
+    public readonly appColor?: string
   ) {}
   command() {
     this.history.push(this.link);
   }
 }
 
+function RenderCommand(suggestion: Command) {
```
iddan, line 58: Name of the component should not include "render"

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -33,13 +43,43 @@ export class NavigationCommand implements Command {
...
+    <>
+      {showAppData && (
+        <>
+          <ApplicationIcon name={appName || ""} color={appColor} />
+          <span className="command-palette__app-name">{appName}</span>
+        </>
+      )}
+      <Icon icon={type} />
+      {highlight && highlight[0] ? (
+        <span dangerouslySetInnerHTML={{ __html: highlight[0] }} />
```
iddan, line 71: What if you have multiple highlights?

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -109,6 +156,21 @@ const CommandPalette = ({ trigger }: Props) => {
...
+        allowTypo: true,
+        scoreFn: (item) => {
+          const command: NavigationCommand = item.obj;
+          const scoreFactor = command.isCurrentApp ? 1000 : 0;
+
+          return Math.max(
+            item[0] ? item[0].score + scoreFactor : -1000,
+            item[1] ? item[1].score + scoreFactor : -1000
+          );
+        },
```
iddan, line 171: Extract to a function at the top level and add a unit test

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
yuval-hazaz: The first highlight is for the "name" property, the second one is for the appName. I thought there is no benefit from highlighting the app name. 
I will have a session with @Leeyaacov tomorrow and if needed, I will add it.

PR #590, opened 2020-10-11
Watch build status

yuval-hazaz:
- [x] remove polling from BuildList
- [X] useBuildWatchStatus to update build while the process is still running


packages/amplication-client/src/VersionControl/use-buildWatchStatus.tsx
```diff
@@ -0,0 +1,89 @@
+import { useEffect, useMemo } from "react";
```
iddan, line 1: Filename should be either `useBuildWatchStatus.tsx` or `use-build-watch-status.tsx`

packages/amplication-client/src/VersionControl/use-buildWatchStatus.tsx
```diff
@@ -0,0 +1,89 @@
...
+      stopPolling();
+    }
+  }, [buildProcessCompleted, stopPolling]);
+
+  //cleanup polling
+  useEffect(() => {
+    return () => {
+      stopPolling();
+    };
+  }, [stopPolling]);
```
iddan, line 54: Why two calls to stopPolling?

packages/amplication-client/src/VersionControl/use-buildWatchStatus.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const buildProcessCompleted = useMemo(() => {
+    if (!build.action?.steps?.length) return false;
+    const steps = build.action.steps;
+
+    const result =
+      steps.find((step) => step.name === GENERATE_STEP_NAME) &&
+      steps.find((step) => step.name === BUILD_DOCKER_IMAGE_STEP_NAME) &&
+      steps.every((step) => step.completedAt);
+    return result;
+  }, [build]);
```
iddan, line 25: Move this logic to the backend.

yuval-hazaz: ok, will do so in the next PR since I already made changes to the logic in the next PR

packages/amplication-client/src/VersionControl/use-buildWatchStatus.tsx
yuval-hazaz: The second one is used to stop the polling using the cleanup callback.

PR #591, opened 2020-10-11
Data Service Generator: Fix Code Generation

iddan:
- Create data mapping in controller update and create methods
 - Exclude relation fields from entity DTO
 - Don't include scalar list fields in generated where input DTO
 - Add default permissions for default user entity
 - Fix UseInterceptors usage in generated controller code
 - Cleanup data service generator E2E spec


iddan:
@yuval-hazaz please review retrospectively.

PR #593, opened 2020-10-11
Build Status

yuval-hazaz:
- [x] Remove field Build.Status from DB
- [x] Calc Build.Status based on the action steps status (server-side)
- [x] Remove calculation of build.Status in UseBuildWatchStatus (client-side)
- [x] Show separate steps statuses on build component


iddan (approved):

iddan:
Ignored some design issues as this is meant to be a temporary solution.

PR #596, opened 2020-10-11
Simplify setting feature flags

iddan:
We've tried to implement feature flags naïvely with `SHOW_UI_ELEMENTS`. Even though the implementation was straightforward it required many small changes in the code. As we want to introduce additional feature flags in the following versions I've refactored the implementation to require fewer code changes for introducing new flags.
It reduced the awareness of modules to `SHOW_UI_ELEMENTS` from 10 to 2. Partially because in the new implementation if a feature flag is turned off it doesn't need to be mentioned.


yuval-hazaz (approved):
Very important! good job

PR #611, opened 2020-10-13
Client: Build deployment

yuval-hazaz:


packages/amplication-client/src/VersionControl/Build.tsx
```diff
@@ -238,6 +234,48 @@ const Build = ({ build, onError, open }: Props) => {
...
+          />
+
+          <span>{build.status}</span>
+        </>
+      )}
+      <span className="spacer" />
+      <UserAndTime account={account} time={build.createdAt} />
+    </div>
+  );
+};
```
iddan, line 277: Can you extract this to another file?

packages/amplication-server/src/core/build/build.service.spec.ts
```diff
@@ -161,6 +162,9 @@ const EXAMPLE_ACTION_STEP = {
   id: 'EXAMPLE_ACTION_STEP_ID'
 };
 
+
+const deploymentFindManyMock = jest.fn();
```
iddan, line 166: Please add a test that checks this mock.

iddan (commented):
Good job, see comments.

packages/amplication-client/src/VersionControl/Build.tsx
yuval-hazaz: This component will never be used without the build component and it is even not exported. Do we really want to have it in a separate file?

iddan: Yes, any component more than a few lines long better exist in a component inside the directory of the main component.

iddan: This is not a show stopper. Please handle whenever you have the time.

PR #613, opened 2020-10-13
Deployment status

yuval-hazaz:
I updated the list of available statuses for deployment

Waiting - for a new deployment
Completed - for when the deployment is ready and running
Failed - for when the deployment has dailed
Removed - for deployments that are no longer live, as they were replaced by a newer deployment on the same environment


iddan:
@yuval-hazaz why isn't Removed represented with the `deletedAt` property?

yuval-hazaz:
There is no deletedAt property
Also, It is not actually deleted, the history of the deployment is kept and can be used for rollback for example. 
The "Removed" status just saying - This deployment does no longer exist in the environment - It was replaced by a newer deployment

iddan:
> There is no deletedAt property

We can add it if needed.

> Also, It is not actually deleted, the history of the deployment is kept and can be used for rollback for example.

This is correct for all the soft deletes in the system.

> The "Removed" status just saying - This deployment does no longer exist in the environment - It was replaced by a newer deployment

I still fail to understand the difference between soft deleting entities and this.

yuval-hazaz:
Soft delete is a result of a DELETE action. The users decide that they no longer want the entity and ask to delete it.

In this case, the user just deploy a new version. The status can be replaced with a flag called "isCurrent" or a back link from the environment with "CurrentDeploymnet". The deployment record is not deleted and should not be deleted

iddan (approved):

PR #614, opened 2020-10-13
Deployment status - client

yuval-hazaz:


packages/amplication-client/src/VersionControl/useBuildWatchStatus.tsx
```diff
@@ -41,6 +42,24 @@ const useBuildWatchStatus = (build: models.Build) => {
...
+
+  if (
+    build.deployments?.some(
+      (deployment) => deployment.status === models.EnumDeploymentStatus.Waiting
+    )
+  ) {
+    return true;
+  }
+
+  return false;
```
iddan, line 60: Please unite this to a single boolean expression

packages/amplication-client/src/VersionControl/Build.tsx
```diff
@@ -244,17 +245,22 @@ const BuildHeader = ({ build, deployments }: BuildHeaderProps) => {
 
   const deployedClassName = `${CLASS_NAME}__header--deployed`;
 
+  const isDeployed =
+    deployments &&
+    deployments.length &&
+    deployments[0].status === models.EnumDeploymentStatus.Completed;
```
iddan, line 251: can be greatly simplified with lodash head function

packages/amplication-client/src/VersionControl/BuildDeployments.tsx
```diff
@@ -0,0 +1,58 @@
...
+import "./BuildDeployments.scss";
+
+const CLASS_NAME = "build-deployments";
+
+type Props = {
+  build: models.Build;
+};
+
+const BuildDeployments = ({ build }: Props) => {
+  if (!SHOW_DEPLOYER) return null;
```
iddan, line 15: I think it’s better to be checked outside the component

packages/amplication-client/src/VersionControl/useBuildWatchStatus.tsx
```diff
@@ -41,6 +41,19 @@ const useBuildWatchStatus = (build: models.Build) => {
...
+function shouldReload(build: models.Build | undefined): boolean {
+  return (
+    !build ||
+    build.status === models.EnumBuildStatus.Running ||
+    Boolean(
+      build.deployments?.some(
+        (deployment) =>
+          deployment.status === models.EnumDeploymentStatus.Waiting
+      )
+    )
```
iddan, line 53: I think ending || false will be better here than wrapping with Boolean as some reruns Boolean value

packages/amplication-client/src/VersionControl/BuildDeployments.tsx
yuval-hazaz: There is ofcource another check outside the component. 
I decided to add another check inside just to validate even if the dialog is used somewhere else in the future

packages/amplication-client/src/VersionControl/Build.tsx
yuval-hazaz: If we use head we should still check that the results of head are not undefined before we check the status so I don't think it will greatly simplify...

const deployment = head(deployments);
const isDeployed = deployment && deployment.status === models.EnumDeploymentStatus.Completed;

Am I missing anything?

PR #626, opened 2020-10-14
Add tests to App Resolver

GalCegla:
Add tests:

-Find one App
-Find many Entities
-Find many Builds
-Find many Environments


iddan (approved):
Perfect

PR #630, opened 2020-10-15
Environment domain

yuval-hazaz:
Return the environment domain to the client.
Currently, the domain is taken from env variable, in the future we will replace it with a dynamic value per cloud provider


iddan:
Let's expose the full URL through GraphQL. That will allow us to change the implementation internally if needed

yuval-hazaz:
> Let's expose the full URL through GraphQL. That will allow us to change the implementation internally if needed

done, please review

packages/amplication-server/src/core/environment/environment.resolver.ts
```diff
@@ -22,4 +36,18 @@ export class EnvironmentResolver {
   ): Promise<Environment | null> {
     return this.environmentService.updateEnvironment(args);
   }
+
+  @ResolveField()
+  async domain() {
+    /**@todo: connect environment to a cloud provider and return the domain dynamically  */
+    return this.configService.get(GCP_APPS_DOMAIN_VAR);
+  }
```
iddan, line 44: Remove this please.

PR #634, opened 2020-10-16
Delete LICENSE.md

alonram:


alonram:
Update instead of delete

PR #636, opened 2020-10-18
Add all tests to Account Resolver

GalCegla:
-All tests are written.
-All tests use @UserEntity, with .skip until issue resolved.


iddan:
Updated the test to use the `mockGqlAuthGuardCanActivate` test utility. You can now use it as an example for other tests.
Corrected the implementation of `updateAccount()` in account resolver (it confused user ID with account ID). If you find other incorrect code you are welcome to fix them as part of your testing PRs or raise a flag.
Simplified the implementation of `updateAccount()` in account service to wrap Prisma's API transparently. Such simplifications are welcome when appropriate.

PR #641, opened 2020-10-19
Build container async and get updates

yuval-hazaz:
Container-builder now returns the status of the build action before it actually complete and provides a function to get status updates as long as the task is running.


packages/amplication-container-builder/src/cloud-build/CloudBuildProvider.ts
```diff
@@ -17,8 +26,48 @@ export class CloudBuildProvider implements IProvider {
       projectId: this.projectId,
       build: createConfig(repository, tag, url, buildArgs),
     });
-    // Wait for build to finish
-    const [finishedCloudBuildBuild] = await cloudBuildBuild.promise();
-    return { images: finishedCloudBuildBuild.images as string[] };
+
```
iddan, line 29:
```typescript
const {
      build,
    } = (operation.metadata as unknown) as google.devtools.cloudbuild.v1.BuildOperationMetadata;
```

packages/amplication-container-builder/src/cloud-build/CloudBuildProvider.ts
```diff
@@ -17,8 +26,48 @@ export class CloudBuildProvider implements IProvider {
...
+    return {
+      status: EnumBuildStatus.Running,
+      statusQuery: { filter: `tags="${tag}"` },
```
iddan, line 32: { id: build.id }

packages/amplication-container-builder/src/cloud-build/CloudBuildProvider.ts
```diff
@@ -17,8 +26,48 @@ export class CloudBuildProvider implements IProvider {
...
+        statusQuery,
+        "Can't find the specified build in Cloud Build service"
+      );
+    }
+    switch (build.status) {
+      case BUILD_STATUS_WORKING:
+      case BUILD_STATUS_QUEUED:
+        return {
+          status: EnumBuildStatus.Running,
+          statusQuery: { filter: `build_id="${build.id}"` },
```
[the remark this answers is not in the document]
yuval-hazaz: I prefer to return the query with every result to keep all the responsibility within the provider. 
Otherwise, the build service needs to check with the return result if the query exist or not, or have a separate code path for CreateBuild and getStatus

PR #702, opened 2020-10-26
Add missing auth fields for apps created in previous versions

yuval-hazaz:


packages/amplication-data-service-generator/src/create-data-service.ts
```diff
@@ -115,6 +115,18 @@ function createUserEntityIfNotExist(entities: Entity[]): [Entity[], Entity] {
   const nextEntities = entities.map((entity) => {
     if (entity.name === USER_ENTITY_NAME) {
       userEntity = entity;
+
+      //Add any missing auth field for backward compatibility with previously created apps
+      const missingAuthFields = USER_AUTH_FIELDS.filter(
+        (authField) =>
+          !entity.fields.find(
+            (existingField) => existingField.dataType === authField.dataType
```
iddan, line 123: Search by name, not data type

Issue #608, opened 2020-10-12
Client: Create Storybook


yuval-hazaz:
Create a storybook for amplication components

PR #728, opened 2020-10-29
Design system

meeroslava:
add global scss to storybook, add onChange to TextInput
Resolves: #608


packages/amplication-design-system/.storybook/preview.js
```diff
@@ -0,0 +1,5 @@
+
+import '!style-loader!css-loader!sass-loader!../src/index.scss';
```
iddan, line 2:
```suggestion
import '../src/index.scss';
```

packages/amplication-design-system/.storybook/preview.js
```diff
@@ -0,0 +1,5 @@
+
```
iddan, line 1: Change file type to `.ts`
Extra white space.

PR #753, opened 2020-11-03
Entity get fields

### LINKED ISSUE #180: Remove version number param from entity.get[Children] methods
Entity.get[Children] methds (getFields, getPermissions...) should not ask for version number and instead it should return the children of the current version.

In addition, create new methods on EntityVersionService to get the children of a specific version

yuval-hazaz:
resolves #180


iddan (dismissed):

iddan:
Seems like tests are failing

PR #756, opened 2020-11-04
Add UserEntity related tests to AppResolver

GalCegla:
-Add should create an app test.
-Add should get a pending change test.

### DIFF
--- a/packages/amplication-server/src/core/app/app.resolver.spec.ts
+++ b/packages/amplication-server/src/core/app/app.resolver.spec.ts
@@ -388,7 +388,7 @@ const findManyEnvironmentsMock = jest.fn(() => {
 
 const mockCanActivate = jest.fn(mockGqlAuthGuardCanActivate(EXAMPLE_USER));
 
-describe('AppService', () => {
+describe('AppResolver', () => {
   let app: INestApplication;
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/app/app.resolver.spec.ts
```diff
@@ -369,9 +386,9 @@ const findManyEnvironmentsMock = jest.fn(() => {
   return [EXAMPLE_ENVIRONMENT];
 });
 
-const mockCanActivate = jest.fn(() => true);
+const mockCanActivate = jest.fn(mockGqlAuthGuardCanActivate(EXAMPLE_USER));
 
-describe('AppResolver', () => {
+describe('AppService', () => {
```
iddan, line 391: AppResolver was the correct one

iddan (commented):
Great job

packages/amplication-server/src/core/app/app.resolver.spec.ts
GalCegla: Weird, I probably branched from old master!

iddan: But you haven't fixed it 😢

PR #759, opened 2020-11-05
Swagger UI

yuval-hazaz:


iddan (dismissed):

iddan:
Good job! For CI to succeed you need to update the snapshot

PR #783, opened 2020-11-11
Update README.md prerequisite installs

yuval-hazaz:
@iddan please see that all links are making sense, or maybe you think we should add CLI command instead for any one of them?


iddan (dismissed):

iddan:
Looks good to me

PR #828, opened 2020-11-23
Scheduler

iddan:
Add a simple scheduler to the local setting that calls the server (configurable route) every minute (configurable timing).
This PR changes the setup of the server to use `npm run docker` instead of `docker:db` as the docker compose of the server includes a scheduler now.

### Next Steps
 - Add a controller to the server and update the scheduler to call it instead of the server root
 - Setup a secret key for the scheduler to use when communicating with the server
 -  Add a Google backed scheduler to Terraform


yuval-hazaz (commented):
I only see the configuration changes but can't find the source for amplicaiton/scheduler...
What am I missing?

PR #834, opened 2020-11-24
Server: Fix Remove Background Service Tests

GalCegla:
- Update test for create method in Build Service
 - Remove build method tests in accordance with the removal of the method
 - Remove any Background Service related components from Build Service Spec
 - Remove try/catch in Build Service


iddan:
Please add with another PR a substitute test for `should catch an error when trying to build` now that the build process is part of the create method.

PR #837, opened 2020-11-24
Auto deploy

yuval-hazaz:


packages/amplication-server/src/core/deployment/deployment.service.ts
```diff
@@ -98,9 +101,49 @@ export class DeploymentService {
...
+
     @Inject(WINSTON_MODULE_PROVIDER) private readonly logger: winston.Logger
   ) {}
 
+  async autoDeployToSandbox(build: Build): Promise<Deployment> {
+    const sandboxEnvironment = await this.environmentService.getDefaultEnvironment(
+      build.appId
+    );
+
+    const deployment = (await this.prisma.deployment.create({
```
iddan, line 114: Use `this.create` method instead

yuval-hazaz: The two methods Create, and AutoDeply uses a different set of parameters and each method work differenly.
In Create we are also creating a new "Action" and in "AutoDeploy" we are connecting to an existing "Action"
Also, in Create we are adding all the initial steps like "add to queue" and deploy info, in AutoDeploy we only add a step to the action

PR #842, opened 2020-11-25
Commit builds

yuval-hazaz:
- [x] commit.getBuilds
- [x] commit.getChanges


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -404,6 +404,50 @@ export class EntityService {
...
+        entity.pluralDisplayName = changedVersion.pluralDisplayName;
+      }
+
+      return {
+        resourceId: entity.id,
+        action: action,
+        resourceType: EnumPendingChangeResourceType.Entity,
+        versionNumber: changedVersion.versionNumber,
+        resource: entity
+      };
```
iddan, line 447: This mapping should be done outside the entity service.

PR #853, opened 2020-11-26
Commit page - changes list

yuval-hazaz:


packages/amplication-client/src/VersionControl/PendingChangeDiff.tsx
```diff
@@ -64,9 +77,9 @@ const PendingChangeDiff = ({ change }: Props) => {
     return getEntityVersionYAML(dataCurrentVersion);
   }, [dataCurrentVersion]);
 
-  const oldValue = useMemo(() => {
-    return getEntityVersionYAML(dataLastVersion);
-  }, [dataLastVersion]);
+  const OtherValue = useMemo(() => {
```
iddan, line 80: should be in camelCase instead of PascalCase

PR #897, opened 2020-12-03
Design system: checkbox

yuval-hazaz:
Move CheckboxField and CheckboxListField from client/components to design-system
The checkbox style and display is still not 100%, but it is currently used only in "EntityPage" which is not relevant.
When utilizing either field in a screen, the style should be updated


packages/amplication-design-system/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <CheckboxListField
+            name="checkboxListName"
+            onChange={props.onChange}
+            options={OPTIONS}
+          />
+        );
+      }}
+    </Formik>
+  );
+};
```
iddan, line 58: This should be in separate stories file

PR #919, opened 2020-12-06
DSG: admin UI navigation

yuval-hazaz:
- [x] Use Link for navigation with react-router
- [x] Generate static routes more "natively"


packages/amplication-data-service-generator/src/admin/app/create-app.ts
```diff
@@ -33,10 +33,9 @@ export async function createAppModule(
     }
   );
   interpolate(file, {
-    ROUTES: jsxElement`<Switch>
-      {loginRoute}
+    ROUTES: jsxElement`<div>
```
iddan, line 36: No need for a div, use fragment

iddan (dismissed):

PR #935, opened 2020-12-08
Admin UI : Entity select components

yuval-hazaz:
For each entity create a select component and use it when the entity is used in a lookup field of another entity

### DIFF
--- a/packages/amplication-data-service-generator/src/admin/entity/create-entity-select-components.ts
+++ b/packages/amplication-data-service-generator/src/admin/entity/create-entity-select-components.ts
@@ -13,13 +13,13 @@ export async function createEntitySelectComponents(
     await Promise.all(
       entities.map(
         async (entity): Promise<[string, EntityComponent]> => {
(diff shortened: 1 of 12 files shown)


packages/amplication-data-service-generator/src/admin/entity/create-entity-select-components.ts
```diff
@@ -0,0 +1,27 @@
...
+  entities: Entity[],
+  dtos: DTOs,
+  entityToDirectory: Record<string, string>,
+  dtoNameToPath: Record<string, string>
+): Promise<Record<string, EntityComponent>> {
+  return Object.fromEntries(
+    await Promise.all(
+      entities.map(
+        async (entity): Promise<[string, EntityComponent]> => {
+          const selectComponents = await createEntitySelectComponent(
```
iddan, line 16:
```suggestion
          const selectComponent = await createEntitySelectComponent(
```

PR #941, opened 2020-12-08
Design system: Breadcrumbs

yuval-hazaz:


packages/amplication-design-system/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import {
+  Breadcrumb as PrimerBreadcrumb,
+  BreadcrumbItemProps as PrimerBreadcrumbItemProps,
+} from "@primer/components";
+import "./Breadcrumbs.scss";
+
+export type Props = {
+  children: React.ReactNode;
+};
```
iddan, line 10: Use primer type

PR #953, opened 2020-12-08
Admin UI - Role select

### LINKED ISSUE #946: Data Service Generator: Role Select
**Is your feature request related to a problem? Please describe.**
Currently, the generated admin forms display a text input for the roles field. 

**Describe the solution you'd like**
They should display a select input of the app roles.

yuval-hazaz:
Adds `<RoleSelect />` component and uses it for fields with data type `Roles`.
Resolves #946

### DIFF
--- a/packages/amplication-data-service-generator/src/admin/static/src/user/RoleSelect.tsx
+++ b/packages/amplication-data-service-generator/src/admin/static/src/user/RoleSelect.tsx
@@ -1,11 +1,16 @@
 import React from "react";
 import { SelectField, SelectFieldProps } from "@amplication/design-system";
 //@ts-ignore
-import { EnumRoles } from "./EnumRoles";
+import { ROLES } from "./roles";
 
-const OPTIONS = Object.keys(EnumRoles).map((key) => ({
-  value: key,
-  label: key,
+declare interface Role {
+  name: string;
+  displayName: string;
+}
+
+const OPTIONS = ROLES.map((role: Role) => ({
+  value: role.name,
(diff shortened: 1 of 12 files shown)


packages/amplication-data-service-generator/src/admin/static/src/user/RoleSelect.tsx
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+import { SelectField, SelectFieldProps } from "@amplication/design-system";
+//@ts-ignore
+import { EnumRoles } from "./EnumRoles";
+
+const OPTIONS = Object.keys(EnumRoles).map((key) => ({
```
iddan, line 6: Should use `Object.entries()` and use keys as labels and values as values.

iddan: But I think it would be easier to just create an options array using the role display name

yuval-hazaz: We should be able to fetch the roles display names somehow. In the admin the developer can set both a name and a display name for each role

yuval-hazaz: > But I think it would be easier to just create an options array using the role display name

Please do so

yuval-hazaz:
@iddan, please fix the test

iddan (approved):

PR #972, opened 2020-12-10
Admin UI breadcrumbs

yuval-hazaz:
- [x] Show the title of the entity in the breadcrumbs when editing an existing entity
- [x] Show breadcrumbs on "new entity" component
- [x] Show the parent page (list) when viewing or creating a single entity. (This is temporary and should  be replaced by a proper structure of the components)


iddan:
Make sure this code is utilizing the latest updates

PR #1054, opened 2020-12-24
E2E Tester: Pending Changes

belkind27:
tests adding new entity to app and check if the change added to the pending changes list using jest-pupeteer

### DIFF
--- a/packages/amplication-e2e-tester/tests/functions.ts
+++ b/packages/amplication-e2e-tester/tests/functions.ts
@@ -7,7 +7,7 @@ export function createRandomName(): string {
 export async function signUp(
   signUpBtn: string,
   continueBtn: string
-): Promise<Response> {
+): Promise<void> {
   await (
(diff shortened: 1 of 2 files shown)


packages/amplication-e2e-tester/tests/functions.ts
```diff
@@ -34,7 +34,8 @@ export async function signUp(
   await (
     await page.waitForXPath(`//button[contains(text(),'${continueBtn}')]`)
   ).click();
-  return page.waitForNavigation();
+  await page.waitForNavigation();
+  return page.waitFor(3000);
```
iddan, line 38: Why did you add a wait?

belkind27: sometimes tests using this function randomlly fail because they cant find the xpath of the create new app element, this ensure it doesnt happen

iddan (approved):

PR #1098, opened 2021-01-04
Container Builder: Document and add a prepare command for it

iddan:
Document the container builder package and add it to the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Amplication Data Service Generator is the component that generates the server-si
 
 To use this package as a library or as a CLI with other projects, follow the instructions on the package page.
 
-### [Amplication Container Builder](./packages/amplication-data-service-generator/README.md)
+### [Amplication Container Builder](./packages/amplication-container-builder/README.md)
 
 Amplication Container Builder is a multi-provider library for building Docker containers.
 The package provides Nest.js integration out of the box.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -60,6 +60,17 @@ Amplication Data Service Generator is the component that generates the server-si
 
 To use this package as a library or as a CLI with other projects, follow the instructions on the package page.
 
+### [Amplication Design System](./packages/amplication-data-service-generator/README.md)
+
+Amplication Data Service Generator is the component that generates the server-side code (models, Prisma client, REST API, authentication, and authorization filters) of the application built with Amplication. This package is used internally by Amplication server.
+
+To use this package as a library or as a CLI with other projects, follow the instructions on the package page.
```
yuval-hazaz, line 67: This text talks about DSG instead of Design System

README.md
```diff
@@ -60,6 +60,17 @@ Amplication Data Service Generator is the component that generates the server-si
...
+
+### [Amplication Container Builder](./packages/amplication-container-builder/README.md)
```
yuval-hazaz, line 69: I think this is not "cooked" enough to be used as a separate package and maybe should be removed from the main README for now

yuval-hazaz (changes requested):

README.md
iddan: It works perfectly, I don't see why not

iddan: But I don't have a strong opinion about it.

yuval-hazaz: I suggest keeping the readme simple for the current time

yuval-hazaz (approved):

PR #1099, opened 2021-01-04
Scheduler: Add to the readme

iddan:


README.md
```diff
@@ -60,6 +60,11 @@ Amplication Data Service Generator is the component that generates the server-si
 
 To use this package as a library or as a CLI with other projects, follow the instructions on the package page.
 
+### [Amplication Scheduler](./packages/scheduler/README.md)
```
yuval-hazaz, line 63: Should we really list all the packages in the README? isn't that takes the focus from the main target?

iddan: I think it's okay at the bottom of the document so people would know what do we hold in the monorepo

README.md
```diff
@@ -60,6 +60,11 @@ Amplication Data Service Generator is the component that generates the server-si
...
+
+Amplication Scheduler is a simple HTTP cron job scheduler.
```
udanna, line 65: @iddan does this work for both HTTP and HTTPS requests? (Or is this a stupid question?...), and is it important to note HTTP? Because I was thinking of tweaking it to: Amplication Scheduler is a cron job scheduler that integrates with Docker and Docker Compose.

yuval-hazaz: HTTP as a transport protocol and not HTTP as the unsecured version of HTTPS - So, it can work both with HTTP and HTTPS but it is no important in this context

yuval-hazaz:
We've decided to leave this addition out for the moment. closing this PR

PR #1160, opened 2021-01-19
Data service worker

### LINKED ISSUE #1159: Move build process to a worker thread

yuval-hazaz:
resolves #1159


iddan:
Nice work! It doesn't make sense to me the data service generator is aware of spawning threads. I would suggest moving the thread spawning logic (basically everything done in this PR) to the server module so the data service generator will continue to expose a simple function and the server, which has the specific need for the threading, will be responsible for spawning the threads.

aniadollxo:
100 %

PR #1177, opened 2021-01-27
Add annotations to Credentials

petarvujovic98:
Add missing ApiProperty and Field decorators


yuval-hazaz (changes requested):
Great, thanks!

Please run "update-snapshot" on DataServiceGenerator to update the snapshot test with your changes and commit the updated snapshot.
image

Also, please pay attention that currently the generated application only utilizes Basic HTTP auth so the credentials need to be sent in the request header. The auth endpoints are not yet being used.

see this
https://docs.amplication.com/docs/api

Congratulation on your first PR, looking for many more to come 👍

yuval-hazaz:
@all-contributors please add @petarvujovic98  for code, bug

Issue #1196, opened 2021-02-02
Client: Unconfirmed github sync toggle-off won't return to checked state.


liyachun01:
**Describe the bug**
A 'Sync with Github' switch button, if being toggled off but cancelled, won't return back to it's visually 'checked' state (unless a page refresh is triggered).

**To Reproduce**
1. Go to an app's 'Sync with Github' setting panel, make sure its already set to synced.
2. Try toggle off the switch, it temporarily get visually 'unchecked'. Meanwhile you get promped by a confirm dialog.
3. Clicked dismiss /  close / click outside the dialog.
4. The toggle won't return 'checked', only visually, because click it will promp comfirm again.
5. Refresh the app. The toggle goes back to normal.

**Expected behavior**
The switch button should switch back to the 'checked' state once the user cancelled the toggle-off.

**Screenshots**
![uncancellable-toggle-off](https://user-images.githubusercontent.com/7907204/106632800-1a540080-65b9-11eb-804e-55b8ad0d9a35.gif)
_The bug_

![Grabación de pantalla 2021-02-03 a las 12 45 56 a  m](https://user-images.githubusercontent.com/7907204/106632975-48d1db80-65b9-11eb-8c70-305fbef94d9f.gif)
_The ideal behavior_

**Desktop (please complete the following information):**
 - OS: macOS Big Sur 11.2 (20D64)
 - Browser: Microsoft Edge
 - Version: 88.0.705.56

**Additional context**
- Was trying to look into #1127, but found this issue instead.

PR #1197, opened 2021-02-02
fix(client): github sync toggled-off unconfirm

liyachun01:
Fix #1196.

When a toggle-off is clicked, the `Toggle` component triggers an `onValueChange` and behind the scenes, an implicit `[MDCSwitchFoundation].setChecked(false)` was fired .

Now that the toggle-off is cancelled, the app must explicitly call `[MDCSwitchFoundation].setChecked(true)` in order to get back to its 'checked' state.

Grabación de pantalla 2021-02-03 a las 1 06 28 a  m

PR #1198, opened 2021-02-02
Fix/gh toggle off unconfirm

yuval-hazaz:
push #1197 to master after removing unintended additional files


liyachun01:
> push #1197 to master after removing unintended additional files

Closes #1196

Issue #1245, opened 2021-02-17
[scheduler]: [Wrong regex for schedule format is not accepting any digits, but allows letter 'd']


kpk-pl:
**Describe the bug**
There is a bug in the regex used to validate the schedule. 
The regex is "^(((d+,)+d+|(d+(\\/|-)d+)|d+|\\*) ?){5}$" and I can only assume that the intention was to allow digits with the commonly used \d. However, there is no backslash and this does not work.

The end effect is that ANY valid cron schedule is not accepted as valid, apart from trivial '* * * * *' (stars are recognized properly). 

**To Reproduce**
Use the following format for schedule '1 * * * * *'
The error is visible in logs:
scheduler_1  | Loaded configuration from: /etc/scheduler/config
scheduler_1  | Config validation errors [
scheduler_1  |   {
scheduler_1  |     keyword: 'pattern',
scheduler_1  |     dataPath: '.schedule',
scheduler_1  |     schemaPath: '#/properties/schedule/pattern',
scheduler_1  |     params: { pattern: '^(((d+,)+d+|(d+(\\/|-)d+)|d+|\\*) ?){5}$' },
scheduler_1  |     message: 'should match pattern "^(((d+,)+d+|(d+(\\/|-)d+)|d+|\\*) ?){5}$"'
scheduler_1  |   }
scheduler_1  | ]
scheduler_1  | Invalid config, see errors

Just to validate, the regex in the current form allows literal letter 'd' which is accepted by the regex by obviously wrong:
scheduler_1  | Loaded configuration from: /etc/scheduler/config
scheduler_1  | Will make a PUT call to http://web-gui/quotes at d minutes past the hour
scheduler_1  | Unknown alias: d

**Expected behavior**
Regex needs to be corrrected to allow digits. Replace the 'd' with either properly escaped version or just use [0-9]

**Desktop (please complete the following information):**
any, latest, with docker-compose

kpk-pl:
@iddan I saw your post on SO and really decided to use this in my project because Ofelia does not work for me at this point. However this issue makes the whole scheduler unusable with json config files.

yuval-hazaz:
hi @kpk-pl 
I can confirm this is a bug in the regex pattern. Thanks for catching it and reporting it.
In case you want to send a PR with a fix, you are more than welcome. 
Otherwise, we will probably be able to push a fix soon

The schema file is here
https://github.com/amplication/amplication/blob/master/packages/amplication-scheduler/src/config.schema.json

[...]

PR #1246, opened 2021-02-17
Fix scheduler regex for schedule format

kpk-pl:
This addresses https://github.com/amplication/amplication/issues/1245


yuval-hazaz (approved):

yuval-hazaz:
@all-contributors please add @kpk-pl  for code

PR #1276, opened 2021-03-12
Implement Helm Chart: amplication

TimDurward:
### What is this?
This introduces a way to deploy Amplication for a kubernetes cluster using Helm.

### Prerequisites

 Best way to test this is using a Kubernetes cluster installed on your machine. I recommend Dockers Version, if you have it installed on your mac you can enable Kube by going to Dockers settings.
Screen Shot 2021-03-12 at 4 10 48 PM

There's a few other options that you can choose, and the instructions to install them are located [here](https://kubernetes.io/docs/tasks/tools).

You'll also need to install Helm - Instructions found [here](https://helm.sh/docs/intro/install/)

### Installing Amplication Helm Chart
Check out this branch and go to `amplication/chart/amplication`

The amplication chart is shipped with a set of defaults that should get you running without any additional configuration.
Run `helm install amplication .  --namespace amplication --create-namespace`

Helm will apply your manifests to kube and say it has finished installing.

It will install:
1. bitnami/postgresql server
2. Migration job for amp server - waits for a postgres connection before attempting.
3. Installs a Docker Registry and Docker builder using https://hub.docker.com/_/docker. (there has been discussion of removing this piece) and a volume for the registry (docker image storage)
4. installs the amplication server and a volume for `/artifacts`
5. installs the amplication sheduler

You can monitor the whole list of resources by running the following:
```sh
$ kubectl get all --namespace amplication

# Example output when all resources are completed and ready

NAME                                         READY   STATUS      RESTARTS   AGE
pod/amplication-app-5cb4f9d576-ldsvz         1/1     Running     0          53s
pod/amplication-app-migrate-f6xp4            0/1     Completed   0          53s
pod/amplication-docker-dd5bbd98f-5mrtj       1/1     Running     0          53s
pod/amplication-registry-5f969c6554-c4cbs    1/1     Running     0          53s
pod/amplication-scheduler-7b9557cdc5-dmjv2   1/1     Running     0          53s
pod/localpostgres-0                          1/1     Running     0          53s

NAME                             TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)    AGE
service/amplication-app          ClusterIP   xx.xx.xx.xx     <none>        3000/TCP   53s
service/amplication-docker       ClusterIP   xx.xx.xx.xx   <none>        2375/TCP   53s
service/amplication-registry     ClusterIP   xx.xx.xx.xx   <none>        5000/TCP   53s
[...]

PR #1282, opened 2021-03-19
Implement Helm Chart: amplication

yuval-hazaz:
merge #1276 into master


TimDurward:
@yuval-hazaz 
Transferring the conversation over to this PR, @yuval-hazaz asked the [following questions](https://github.com/amplication/amplication/pull/1276#discussion_r597766465).

> Few questions that I still trying to figure:
> 
> 1. Is there a reason to use the chart for development?
> 2. What is required in order to use the chart for our cloud deployment (currently on GCP, planning to move to AWS... it may be a great opportunity )
> 3. Are there any open items that still require attention?
> 

You can definitely use this Helm Chart for development. There's some really great tools out there to do just that. In fact you do all your development pointing at a remote cluster if you really wanted! 
1. https://garden.io/
2. https://skaffold.dev/

If that is something you're interested in, perhaps we can make a ticket for this scope of work.

This Helm Chart is ready to be deployed to all cloud providers! It's abstracted enough that it shouldn't have any problems.

I could think of a few improvements, more specifically to production use-cases. 
1. Using [Service Accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/) for the Amplication pods (so they don't have access to the clusters API, or at least a reduced scope). Or at least a parameter that allows for people to insert their own Service Accounts. 
2. Scoping the privilege given to the Container Builder service, specifically the Docker-in-Docker part. I recommend reading up on https://kurtmadel.com/posts/native-kubernetes-continuous-delivery/building-container-images-with-kubernetes/ and look for the _Docker-in-Docker_ parts. There are other options out there that run Daemonless Docker processes inside a Container. Specifically: https://github.com/GoogleContainerTools/kaniko. But this would require Amplication Server to support a new _Container Builder_.

yuval-hazaz:
> 2\. _Docker-in-Docker_ parts. There are other options out there that run Daemonless Docker processes inside a Cont

Thanks!
I will get into the details of the links you provided.

If you don't mind, let's move the conversation to discord - it would be great to get to know you better.
https://discord.gg/KSJCZ24vj2

PR #1297, opened 2021-03-28
Fix: validate entity name pluralize

almogbhl:


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -534,6 +542,12 @@ export class EntityService {
 
     const entity = await this.acquireLock(args, user);
 
+    if (args.data.name?.toLowerCase() === args.data.pluralDisplayName?.toLowerCase()) {
+      throw new AmplicationError(
+        `The entity name must be in the singular form.`
+      );
+    }
```
yuval-hazaz, line 549: The user may choose to update only a single field (e.g. only the plural display name) so instead of checking against args. data you should also validate against the saved values

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -214,6 +215,13 @@ export class EntityService {
     args: CreateOneEntityArgs,
     user: User
   ): Promise<Entity> {
+
+    if (args.data?.name?.toLowerCase() === args.data?.pluralDisplayName?.toLowerCase()) { 
+      throw new AmplicationError(
+        `The entity name must be in the singular form.`
```
yuval-hazaz, line 221: Use this message instead "the entity name and plural display name cannot be the same"

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -214,6 +215,13 @@ export class EntityService {
     args: CreateOneEntityArgs,
     user: User
   ): Promise<Entity> {
+
+    if (args.data?.name?.toLowerCase() === args.data?.pluralDisplayName?.toLowerCase()) {
```
yuval-hazaz, line 219: Since the value of plural display name may include spaces, you should remove white spaces before do the compare

yuval-hazaz (changes requested):
Well done, 
Please see some comments

also, please run prettier

packages/amplication-server/src/core/entity/entity.service.ts
almogbhl: np, where can I get access to the saved values?

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -527,6 +537,18 @@ export class EntityService {
...
+      ?.toLowerCase()
+      .trim();
+    const currPluralDisplayName = entity?.pluralDisplayName;
+
+    if (newName === (newPluralDisplayName || currPluralDisplayName)) {
+      throw new AmplicationError(
+        `The entity name and plural display name cannot be the same.`
+      );
+    }
+
```
yuval-hazaz, line 551: the new just like the pluralDisplayName can have a new value or a current value
you probably better do something like this:

```
    const newName = args.data.name?.toLowerCase().?trim() || entity.name.toLowerCase().trim();
    const newPluralDisplayName = args.data.pluralDisplayName
      ?.toLowerCase()
      ?.trim() || entity.pluralDisplayName.toLowerCase().trim(); 

    if (newName === newPluralDisplayName) {
      throw new AmplicationError(
        `The entity name and plural display name cannot be the same.`
      );
    }
```

yuval-hazaz (changes requested):

yuval-hazaz (approved):

yuval-hazaz:
@almogbhl well done!

PR #1505, opened 2021-07-05
Fix/GitHub button url

yuval-hazaz:
Fix GitHub login default controller URL


yuval-hazaz:
@all-contributors add @yuval-hazaz  as a contributor for code.

PR #1514, opened 2021-07-05
feat(server): Soft Delete App

yonantan:


packages/amplication-server/src/core/app/app.service.ts
```diff
@@ -321,11 +322,24 @@ export class AppService {
   }
```
yuval-hazaz, line 322: do not return deleted apps

packages/amplication-server/src/core/app/app.service.ts
```diff
@@ -321,11 +322,24 @@ export class AppService {
...
+    return this.prisma.app.update({
+      where: args.where,
+      data: {
+        name: prepareDeletedItemName(app.name, app.id),
+        deletedAt: new Date()
+      }
+    });
   }
 
   async updateApp(args: UpdateOneAppArgs): Promise<App | null> {
```
yuval-hazaz, line 345: do not update deleted apps

yuval-hazaz (changes requested):
besides the couple of comments in app.service.ts please have a look at this file
packages/amplication-server/src/core/permissions/permissions.service.ts

We should prevent access to any resource in deleted apps - the permissions service is a single gateway used by all endpoints. 

We may have few more specific services that need to be updated. I will run a search to pinpoint them.

yuval-hazaz (approved):

yuval-hazaz:
@all-contributors please add @yonantan for code

PR #1599, opened 2021-07-12
fix(deps): dsg mui version

yuval-hazaz:
Revert changes made by Renovate that caused the DSG e2e test to fail

- DSG e2e is not included in the CI before merge to decrease the CI run time. We should include the e2e test on PRs created by Renovate


yuval-hazaz:
@all-contributors please add @yuval-hazaz  for  code

Issue #1312, opened 2021-04-09
Client: Build errors visibilty


yuval-hazaz:
**Describe the bug**
When a build failed the detailed errors can only be found in the build log but it is not always trivial to open and read the build log.

**Expected behavior**
The user should view a detailed error in a more obvious place or at least be directed to the build log to look for the detailed error.

yuval-hazaz:
also, when the build failed, the BUILD ID in the footer is not updated with that latest build - so clicking on the BUILD ID takes the user to the previous build instead of the last failed build - and it causes more confusion

PR #1683, opened 2021-07-24
fix(client): Build errors visibility

lalit8347:
#1312 
Updating buildID on build fail fixed.
(This is not an optimal solution, but I have come up with a new solution, will push in new PR soon)


lalit8347:
I tried to solve the bug without contextAPI, but there was nothing to keep track of errors between Commit.tsx and LastCommit.tsx,
so in need to exchange data between sibling components I had to use it.

yuval-hazaz (approved):

PR #1710, opened 2021-08-27
upload the ERD

tupe12334:


yuval-hazaz:
@tupe12334 
I prefer to have a single source of truth for the DB schema - and as of now, it is the Prisma schema file. 
This kind of static ERD requires frequent updates and it will probably be left un-updated 
Also, a static image with ERD is not really usable (as opposed to a dynamic "live" ERD)

PR #1714, opened 2021-08-30
Secrets manager

tupe12334:


packages/amplication-data-service-generator/src/server/static/src/providers/secrets/secretsManager.service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable } from "@nestjs/common";
+import { ConfigService } from "@nestjs/config";
+import { SecretsManagerServiceBase } from "./base/secretsManager.service.base";
+
```
yuval-hazaz, line 4: add a comment here to explain to the user to replace the default behavior with a link to docs

packages/amplication-data-service-generator/src/server/static/src/providers/secrets/base/secretsManager.service.base.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class SecretsManagerServiceBase {
+  constructor(protected readonly configService: ConfigService) {}
+  getSecret<T>(key: string): T {
+    const value = this.configService.get(key);
+    return value;
+  }
+  constants = {
+    JWT_SECRET: "JWT_SECRET",
+  };
```
yuval-hazaz, line 11: it this being used?

yuval-hazaz (changes requested):
@tupe12334 
Good work. 
Please remove all the "auth provider" changes from this PR and keep only the relevant changes so we can push this to master separately.
see more comments inline

PR #1717, opened 2021-08-30
secretsManager files

tupe12334:


packages/amplication-data-service-generator/src/server/static/src/providers/secrets/base/secretsManager.service.base.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class SecretsManagerServiceBase {
+  constructor(protected readonly configService: ConfigService) {}
+  getSecret<T>(key: string): T {
+    const value = this.configService.get(key);
+    return value;
+  }
+  constants = {
+    JWT_SECRET: "JWT_SECRET",
+  };
```
yuval-hazaz, line 11: why do we need it?

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/server/static/src/providers/secrets/base/secretsManager.service.base.ts
tupe12334: for pre defined must have secrets

yuval-hazaz: this is not the responsibility of the secret manager - it should not manage the keys. and, this code is not being used anyway at this PR.

yuval-hazaz (approved):

PR #1744, opened 2021-09-13
pref(dsg): update docker file to install production only npm

tupe12334:


tupe12334:
The test is passed do you want to enter this change?
@yuval-hazaz

tupe12334:
in the testing I got an error because the nestjs/cli was not in the dependencies and failed to do nest build

PR #1746, opened 2021-09-14
fix/(client): adding auth selection

tupe12334:


packages/amplication-client/src/Application/ApplicationAuthSettingForm.tsx
```diff
@@ -105,19 +105,19 @@ function ApplicationAuthSettingForm({ match }: Props) {
                 </p>
 
                 <div className={`${CLASS_NAME}__space`}>
-                  <ToggleField
-                    name="Http"
-                    label="Basic HTTP"
-                    disabled={!false}
+                  <SelectField
+                    label="Auth provider"
```
yuval-hazaz, line 109:
```suggestion
                    label="Authentication provider"
```

packages/amplication-client/src/Application/ApplicationAuthSettingForm.tsx
```diff
@@ -105,19 +105,19 @@ function ApplicationAuthSettingForm({ match }: Props) {
                 </p>
```
yuval-hazaz, line 105: use this text instead:

Select the authentication method to be used in the generated app.

yuval-hazaz (changes requested):

yuval-hazaz (approved):

PR #1756, opened 2021-10-03
Fix/dynamic swagger auth

tupe12334:
Adding the dynamic auth function to the swagger of the generated server

### DIFF
--- a/packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
+++ b/packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
@@ -55,7 +55,10 @@ export async function createControllerModules(
   const controllerId = createControllerId(entityType);
   const controllerBaseId = createControllerBaseId(entityType);
   const serviceId = createServiceId(entityType);
-
+  const swaggerAuthFunction =
+    authProvider === EnumAuthProviderType.Http
+      ? builders.identifier("ApiBasicAuth")
+      : builders.identifier("ApiBearerAuth");
   const mapping = {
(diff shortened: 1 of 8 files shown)


packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
```diff
@@ -101,7 +105,14 @@ export async function createControllerModules(
       entityServiceModule,
       entity,
       dtos,
-      mapping,
+      {
+        ...mapping,
+        SWAGGER_API_AUTH_FUNCTION:
+          authProvider === EnumAuthProviderType.Http
+            ? builders.identifier("ApiBasicAuth")
+            : builders.identifier("ApiBearerAuth"),
```
yuval-hazaz, line 113: move this code to be part of the original mapping declaration

yuval-hazaz (changes requested):
Well done @tupe12334 
Just one small fix and we can merge

packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
```diff
@@ -101,7 +110,10 @@ export async function createControllerModules(
       entityServiceModule,
       entity,
       dtos,
-      mapping,
+      {
+        ...mapping,
+      },
```
yuval-hazaz, line 115: you no longer need the spread operator

yuval-hazaz (changes requested):

yuval-hazaz (approved):

PR #1765, opened 2021-10-04
feat: client demo app

thucpn:
I create a sample blog built by Amplication and NextJs.
Live demo: https://amplication-client-demo.vercel.app


thucpn:
This is dirty PR so I close it.

PR #1797, opened 2021-10-16
enhancement(data-service-generator): graphql schema file

yonantan:
Added basic support for integrating libs like graphql-codegen, graphql-tools, etc.

Changes:
- update create-app-module.ts
- export generated graphql schema to file `schema.graphqls`

### DIFF
--- a/packages/amplication-data-service-generator/src/server/app-module/create-app-module.ts
+++ b/packages/amplication-data-service-generator/src/server/app-module/create-app-module.ts
@@ -66,7 +66,9 @@ export async function createAppModule(
         const playground = configService.get("GRAPHQL_PLAYGROUND");
         const introspection = configService.get("GRAPHQL_INTROSPECTION");
         return {
-          autoSchemaFile: true,
+          // @todo: allow some documented env variable to override this (e.g. GQL_SCHEMA_EXPORT_PATH)
+          autoSchemaFile: "schema.graphqls",
+          sortSchema: true,
           playground,
           introspection: playground || introspection
         }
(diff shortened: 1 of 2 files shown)


packages/amplication-data-service-generator/src/server/app-module/create-app-module.ts
```diff
@@ -66,7 +66,9 @@ export async function createAppModule(
         const playground = configService.get("GRAPHQL_PLAYGROUND");
         const introspection = configService.get("GRAPHQL_INTROSPECTION");
         return {
-          autoSchemaFile: true,
+          // @todo: allow some documented env variable to override this (e.g. GQL_SCHEMA_EXPORT_PATH)
+          autoSchemaFile: "schema.graphqls",
```
tupe12334, line 70:
```suggestion
          autoSchemaFile: "schema.graphql",
```
by the convention of 
- https://docs.nestjs.com/graphql/quick-start
- https://www.graphql-code-generator.com/

yonantan: as far as I understand, those are just some defaults and examples, not conventions.

fyi, some projects contain multiple services/backends and generate multiple schemas and use schema stitching, federation, etc. to combine them.

from my experience, different extensions make it easier to distinguish between _schema_ (SDL) files and _query/mutation/fragment_ files, and make it easier to work with such tools as graphql-let and graphql-codegen, especially in mono-repos.

p.s., here are some links that I was able to find in 10 seconds to other projects that use the `graphqls` extension, so you know I'm not making this up 😄 :
https://github.com/piglovesyou/graphql-let
https://devdocs.magento.com/guides/v2.4/graphql/develop/create-graphqls-file.html
https://github.com/howtographql/graphql-java/blob/master/src/main/resources/schema.graphqls

yonantan:
hi @mshidlov ,
why was this closed?
any downsides that I'm not aware of?

mshidlov:
@yonantan Sorry, this was supposed to be part of milestone 0.10.3
the PR target branch is supposed to bo release/0.10.3 and not release/0.10.2.
release/0.10.2 was released ant this caused the PR to be locked.

Issue #729, opened 2020-10-29
Client: Compare shows false changes on entity permissions because of inconsistent order


yuval-hazaz:
Changes are falsely detected on entity permissions because of inconsistent ordering  

![image](https://user-images.githubusercontent.com/43705455/97566320-1aa79300-19ef-11eb-96f1-f0aba17552b4.png)

PR #1849, opened 2021-10-27
Feat/make pending changes accurate by "diffing" current and previous versions

### LINKED ISSUE #761: Server: Changing and then changing back to original state is still counted as change
Steps to reproduce:
 1. Change data type of field
 2. Chagne back to original data type
 3. Still shows pending changes

g-traub:
**WIP**
Fixes  #761

## Why?
"Changing and then changing back to original state is still counted as change" as stated in #761.
Pending changes are based on element (Entity or Block) locking (writing a property in its data), which happens for every action (create, update, delete). So once an element is locked, an action cannot unlock it. It can only be done by discarding all pending changes or by committing changes.

## How?
We add the ability to unlock an element by comparing the current and previous versions to know if there are actual changes and update the lock based on the result.

### New locking flow
*(action starts)*
│
▼
lock
│
▼
*(action done)* changes are made
│
▼
***new:* diff current and previous versions to know if there are actual changes**
├──► *(no changes)* ──► **unlock it**
└──► *(changes)* ──► keep it locked

**action = create, update or delete*
**element = Entity or Block*

### Nested records in versions had unpredictable order (related to #729).
- 1st solution (tried) : using `jsondiffpatch`, which can diff two objects and ignore array element that move
- 2nd solution (chosen): sorting nested records when getting versions (in `findMany`) 

## Possible improvements
- The sorting of nested elements to fix the textual diff happening in the frontend (commit 2e8cae4a46cf111027d7f8f9b9958e483b42c648) is done in the method used by GraphQL to resolve entity version permissions. It is done every time a query includes entity version permissions, but It is necessary only for comparing versions. It would be a performance gain to only do it when necessary.
- Refactor locking related methods in a locking service. I couldn't find a type-safe way to make this work. It's both for Entity and Block, and their methods signatures aren't compatible.

## To do:
- [X] Entity changes
- [X] Block changes
- [X] Tests
- [X] Documentation
- [X] Refactor


yuval-hazaz:
@g-traub 
First, I must say I am impressed by the level of understanding of the internal of Amplication. Well done.
I think your approach is very smart and you can move forward with it. 
I still need to make sure I understand correctly the entire flow for each resource (e.g. entity, entity field, entity field permission, etc.) but the overall idea is great. 

Please move forward and let us know if you need anything from our side.

yuval-hazaz:
@g-traub please post a link to this PR on our hacktoberfest channel on Discord and I would love to give you the credits for an implication T-shirt

g-traub:
@yuval-hazaz 
Great! I'm glad you like the approach.
I'll continue working on it !
Thanks for the T-shirt, I appreciate it :grin: I'll post it on discord as soon as it's done

yuval-hazaz:
@all-contributors please add @g-traub for code

yuval-hazaz:
@g-traub excellent work
i sent you a message on Discord, probably you are not connected. 
send me a message any time, I would love to talk.

g-traub:
Thank you @yuval-hazaz ! I'm all done :) I'll go on discord

packages/amplication-server/package.json
```diff
@@ -84,8 +84,10 @@
     "get-stream": "6.0.1",
     "graphql": "14.6.0",
     "graphql-type-json": "0.3.2",
+    "jsondiffpatch": "^0.4.1",
```
mshidlov, line 87: Please anchor the dependency version

packages/amplication-server/package.json
```diff
@@ -84,8 +84,10 @@
...
     "lodash.get": "4.4.2",
     "lodash.head": "4.0.1",
+    "lodash.isempty": "^4.4.0",
```
mshidlov, line 90: Please anchor the dependency version

packages/amplication-server/package.json
```diff
@@ -113,6 +115,7 @@
   "devDependencies": {
     "@nestjs/cli": "7.6.0",
     "@nestjs/testing": "7.6.18",
+    "@types/lodash.isempty": "^4.4.6",
```
mshidlov, line 118:
```suggestion
    "@types/lodash.isempty": "4.4.6",
```

mshidlov:
Hey @g-traub,
Thank you 🙏. 

I will look into it and update 🤙

g-traub:
Thank you !

mshidlov:
@tupe12334 please validate on staging

PR #2041, opened 2021-12-08
feat(dsg): user data give http implementestion

### LINKED ISSUE #1722: Can custom guards and basic strategy
I can't access to user infos in the resquest object after the guard. I would like to use request.user in routes (to return self infos by example)

I Would like to be able to add custom code in the auth strategy like
```basic.strategy.ts
async validate(username: string, password: string): Promise<UserInfo> {
    const user = await this.authService.validateUser(username, password);
    if (!user) {
      throw new UnauthorizedException();
    }
    request.user = user; 👈
    return user;
  }
```

But that will be overwritten in the next sync with amplication app

### LINKED ISSUE #1723: Generated App: Add User object to the request in REST API endpoints
**Is your feature request related to a problem? Please describe.**
The User object is available using the @UserData decorators in GraphQL. Also, it is available as part of the request object. 
This is not the current case with REST API - but it should. 

**Describe the solution you'd like**
Make sure the User object is available on the request, also consider adding a @userData decorator for REST API.

based on the use case described in #1722

tupe12334:
resolve #1722 and resolve #1723


tupe12334:
working by the factory that nestjs show https://docs.nestjs.com/fundamentals/execution-context#current-application-context

PR #2101, opened 2021-12-30
Fix conflicting flags

### LINKED ISSUE #2100: 🐛 Bug Report: Conflicting options in Amplication's CLI
### What happened?

There is a conflicting option in the Amplication's CLI (`entities:fields:update`)

The option for setting field and format has the same shortcut (`-f`).

```
$ amp entities:fields:update --displayName "Creation Date" -f ckxt0rbjj8294951bs645fkdmr1
 ›   Error: Expected --format=ckxt0rbjj8294951bs645fkdmr1 to be one of: JSON, styledJSON, table
 ›   See more help with --help
```

The `-f` flag was supposed to be treated as `field`

### How to reproduce

1. Open a terminal
2. Type `amp entities:fields:update --displayName "When was it created" -f ckxt0rbjj8294951bs645fkdmr1`
3. See the error (The `-f` flag was meant for field but it is treated as for `--format`)0.

### Amplication version

0.1.4

### Environment

-

### Anything else

_No response_

### Code of Conduct

0xsapphir3:
Closes #2100 

I have currently set the flag to `-z`. If you want me to change it then comment and I will make the edit ;)


yuval-hazaz:
Thanks for catching this! 
I don't think Z is the correct selection for this param. 

I suggest that instead of -z we use `-i` 

It can be related to the first letter in "ID" and the second letter in "field"

AsianCat54x:
Ah totally makes sense. I should replace `-z` with `-i`

yuval-hazaz (approved):

mshidlov:
@yuval-hazaz is this can be merged? do we need docs update for that?

yuval-hazaz:
You can merge
docs should be updated automatically before deploying to npm

PR #2183, opened 2022-01-12
Deps/dockerode

tupe12334:
replace #2143


tupe12334:
Its not finish the test because in src/docker/DockerProvider -> DockerProvider class -> build function the oprtions that returns from createBuildImageOptions have t as a string[] instead of string

PR #2184, opened 2022-01-12
deps(server): update @octokit/rest

tupe12334:
replace #2134


tupe12334:
finish tests got green light

Issue #2165, opened 2022-01-09
Add @amplication/server docker file 


mshidlov:
Create a dedicated docker file to @amplication/server.

Using Node 16.3.1 alpine base image, and build best practices:
- Optimize docker layers
- Image size
- Use pinned specific versions

Refer root docker file in master branch, and root docker file branch [refactor/npm-prune](https://github.com/amplication/amplication/tree/refactor/npm-prune)

timothy-amplication:
@mshidlov 
The error below occurs when you only build `amplication/server` and excluding `amplication/client` (`npm run build -- --scope @amplication/server`)
```sh
# Browser Client @ localhost:3000

{"statusCode":404,"message":"ENOENT: no such file or directory, stat '/Users/oz/amplication/packages/amplication-client/build/index.html'"}
```

1. `amplication/server` is using the [NestJS Serve-Static Package](https://docs.nestjs.com/recipes/serve-static) which is serving up the `amplication/client` compiled assets. If we want to separate the Server and Client into their own container images we will most likely need to rip out the *`ServeStaticModule` from [amplication-server/src/app.module.ts](https://github.com/amplication/amplication/blob/master/packages/amplication-server/src/app.module.ts#L21-L30).*
    1. Are there any concerns with this approach?
    2. Are there other areas in `amplication/server` that use/depend on `amplication/client` which may be subject to similar remediation tasks like the above?
2. Are there any ping or health-checks in `amplication/server` I can use to validate that the server is listening and accepting requests (nice to have when testing the server in isolation; and for sending traffic to healthy services)? I didn't see any NestJS specific Health-checks, perhaps it'd be worthwhile to implement [HttpHealthIndicator](https://docs.nestjs.com/recipes/terminus#healthchecks-terminus).

mshidlov:
@timothy-amplication

1. You are definitely right about the relation between `@amplication/server` and `@amplication/client`. if we will break that relationship now, it will cause the root docker file to break. As of today, this is the Dockerfile we are deploying.
Let's do it in steps. For now, keep and ignore `@amplication/client` in the server Docker file.

2. You are right. Readiness and liveliness are essential when making infrastructure aware of the application's availability. I will open up an issue regarding this matter. Would you like me to assign you?

timothy-amplication:
@mshidlov I gone ahead and set up the initial task for Health Checks for `@amplication/server` - https://github.com/amplication/amplication/issues/2192. Feel free to add anything else i may be forgetting.

PR #2195, opened 2022-01-14
feat(server): Add @amplication/server Dockerfile

timothy-amplication:
GH Issue: https://github.com/amplication/amplication/issues/2165

This PR separates `amplication`/server` into its own container image. The following improvements have been made...
1. Reduces image size from 2.45G+ to 984 MB
2. Reduces build time from 550+ Seconds to 330 Seconds.

Report from [Dive](https://github.com/wagoodman/dive)
```
Image name: 5d1f79b669c278d2a14a0b9805f5d21ffc972a3df9834f03fe8f6a0fca275f4f
Total Image size: 984 MB
Potential wasted space: 17 MB
Image efficiency score: 99 %
```

How to build the container image...
1. Go to root dir of amplication
2. run ` docker build -f packages/amplication-server/Dockerfile .` (This is telling Docker to set to the current context `.` and specifying the location of the dockerfile) 
3. run `docker run -p 3000:3000 <docker image sha>` (sha should be outputted to stdout after `docker build` is complete.

### DIFF
--- a/packages/amplication-server/Dockerfile
+++ b/packages/amplication-server/Dockerfile
@@ -1,0 +1,62 @@
+ARG ALPINE_VERSION=alpine3.14
+ARG NODE_VERSION=16.13.1
+
+FROM node:$NODE_VERSION-$ALPINE_VERSION AS base
+ARG NPM_VERSION=8.1.2
+ENV NPM_CONFIG_LOGLEVEL=silent
+ENV OPENCOLLECTIVE_HIDE=1
+
+RUN npm install --global npm@$NPM_VERSION
+RUN npm config set fund false
+
+WORKDIR /app
+COPY lerna.json /app
+COPY package*.json /app
+RUN npm ci --production
+
+FROM base AS build
+WORKDIR /app
+# Copy amplication/server and the its dependent packages
+COPY packages/amplication-server packages/amplication-server
+COPY packages/amplication-container-builder packages/amplication-container-builder
+COPY packages/amplication-data packages/amplication-data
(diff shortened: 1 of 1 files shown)


mshidlov:
It is pretty much like the original Dockerfile.
The original Dockerfile generates a very large image (+2GB) and it takes a long time to build.

I want you to try to optimize the image size and the image build time as much as possible.

For example, the build time can be reduced by using `npm prune` command instead of running lerna bootstrap twice (see [Dockerfile](https://github.com/amplication/amplication/blob/refactor/npm-prune/Dockerfile) in branch `refactor/npm-prune`).

Another way to optimize the build is by leveraging the Docker build is to use the docker layers correctly to move commands and create a build script that maximizes the use of caching.

packages/amplication-server/Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+COPY --from=build /app/packages /app/packages
+# Remove any node_modules that include dev-dependencies from @amplication/* packages.
+RUN npm run clean -- --yes
+RUN npm run bootstrap -- -- --production --loglevel=silent --scope @amplication/server --scope @amplication/client --include-dependencies
+RUN npm run prisma:generate
+COPY docker-entrypoint.sh /entrypoint.sh
+RUN chmod 755 /entrypoint.sh
+
+EXPOSE 3000
+ENTRYPOINT [ "/entrypoint.sh" ]
```
mshidlov, line 37: If we ignore the UI in this image do we need it here

Issue #2080, opened 2021-12-22
🐛 Bug Report: No warning shown when the user renames the app's name as same as that of the other app


0xsapphir3:
### 📜 Description

Let's say we have two apps created, one with the name of `A_App` and the second one with `B_App`. Now, If I create a new app with the name of `A_App`, it seems like our app is renamed since there are no warning/errors shown. But when we go to home `app.amplication.com`, It shows the default name.

### 👟 Reproduction steps

1. Create two new apps
2. Rename them as `A`  and `B`
3. Create a new app 
4. Rename it as `A`
5. Got to home (`app.amplication.com`)
6. See that the name is left unchanged 

### 👍 Expected behavior

There should be a warning/error message showing that the app with the same name exists

### 👎 Actual Behavior

It does not show any warnings 

### 💻 Device

Desktop

### 💻 Operating system

Linux

### 🌍 Browser

Chrome

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/amplication/amplication/blob/master/code_of_conduct.md)

mshidlov:
@0xsapphir3 thank you for reporting this issue.

abrl91:
@mshidlov 
Can I take this issue?

mshidlov:
Hey @abrl91 sure I'm assigning you 👍

mshidlov:
Merged in PR #2203 and will be released in `0.11.2`

souravjain540:
@all-contributors please add @0xsapphir3  for bug

PR #2203, opened 2022-01-19
bug(apmlication-client): no error message rename to an existing app name

abrl91:
No warning is shown when the user renames the app's name as same as that of the other app
https://github.com/amplication/amplication/issues/2080


mshidlov:
@abrl91 Hey Amit, thank you for submitting your PR.
I'm testing it to see that it solves the issue after that I will merge it

mshidlov:
@abrl91 Merged!!! Thank you :)

Issue #2201, opened 2022-01-17
🐛 Bug Report: position of  the error message when trying to login with a user that doesn't exist


abrl91:
### What happened?

When I tried to sign in with a user that does not exist, I got an error in the terminal. 
![image](https://user-images.githubusercontent.com/39680385/149796038-a6a37a9e-fdc5-4566-8c3b-4e713ae9e812.png)

I also got an error message presented by a toaster at the left corner of the screen, which I also think is a bad user experience because It was for a short time, and I did not expect to see an error there.

![image](https://user-images.githubusercontent.com/39680385/150208034-4e24e465-575a-4096-a1cf-da2df7bd34fb.png)

### What you expected to happen

I expect that the server will catch the error, and the client will show it to the user while the terminal will be clean.
I also expect to see the error message in the form area for an extended period or even persistent until the user fixes the error.

### How to reproduce

Run the dev environment and open http://localhost:3001/login
open the web console
try to log in with a user that you didn't register with
you will see the error on the terminal

### Amplication version

0.11.1

### Environment

Node version: 14.17.3
npm version: 8.3.1
docker version: 20.10.8
browser: chrom version 97

### Are you willing to submit PR?

Yes I am willing to submit a PR!

### Code of Conduct

I agree to follow this project's Code of Conduct

yuval-hazaz:
Thanks @abrl91 
I assigned the issue to you, so you can go ahead and fix it

yuval-hazaz:
@abrl91 
please use this design for displaying the message.

![image](https://user-images.githubusercontent.com/43705455/150300806-84c3412b-c705-446f-a405-854443df84d5.png)


There is already a code on the page that show other types of errors in the suggested way - so, what is best to do:

1. Create a new component in ../amplication-client/src/components/ErrorMessage.tsx
2. Replace the existing code in login.tsx to use the new component
3. use the new component to show the login error message

PR #2214, opened 2022-01-24
Fix/2201/login error msg

abrl91:
#2201

### DIFF
--- a/packages/amplication-client/src/Components/ErrorMessage.tsx
+++ b/packages/amplication-client/src/Components/ErrorMessage.tsx
@@ -1,0 +1,18 @@
+import {Icon} from "@rmwc/icon";
+import React from "react";
+
+type Props = {
+    error: string | string[] | undefined | null;
+    className: string;
+}
+
+export const LoginErrorMessage = ({error, className}: Props) => {
+    return (
+        <div>
+            {error && (<div className={`${className}__login-error`}>
+                <Icon icon="alert_circle" />
+                {error}
+            </div>)}
+        </div>
+    )
+}
--- a/packages/amplication-client/src/Components/ErrorMessage.tsx
+++ b/packages/amplication-client/src/Components/ErrorMessage.tsx
@@ -1,18 +1,21 @@
 import {Icon} from "@rmwc/icon";
 import React from "react";
+import classNames from "classnames";
+
+import './ErrorMessage.scss';
 
 type Props = {
-    error: string | string[] | undefined | null;
-    className: string;
+    errorMessage: string | string[] | undefined | null;
+    className?: string;
 }
 
-export const LoginErrorMessage = ({error, className}: Props) => {
+const CLASS_NAME = 'amp-error-message';
+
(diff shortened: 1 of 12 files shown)


packages/amplication-client/src/Components/ErrorMessage.tsx
```diff
@@ -0,0 +1,18 @@
+import {Icon} from "@rmwc/icon";
+import React from "react";
+
+type Props = {
+    error: string | string[] | undefined | null;
+    className: string;
+}
+
+export const LoginErrorMessage = ({error, className}: Props) => {
```
yuval-hazaz, line 9:
```suggestion
export const ErrorMessage = ({error, className}: Props) => {
```

packages/amplication-client/src/Components/ErrorMessage.tsx
```diff
@@ -0,0 +1,18 @@
...
+    return (
+        <div>
+            {error && (<div className={`${className}__login-error`}>
```
yuval-hazaz, line 12: Please add a unique class name for the component (e.g. amp-error-message)
the prop className should only be used to extend the styling but the component should come with its win style.

packages/amplication-client/src/User/Login.tsx
```diff
@@ -87,12 +86,8 @@ const Login = () => {
...
       <Formik initialValues={INITIAL_VALUES} onSubmit={handleSubmit}>
         <Form childrenAsBlocks>
-          {urlError && (
-            <div className={`${CLASS_NAME}__login-error`}>
-              <Icon icon="alert_circle" />
-              {urlError}
-            </div>
-          )}
+
+          <LoginErrorMessage error={urlError} className={CLASS_NAME}/>
```
yuval-hazaz, line 90: I think it is better to add the condition on the page instead of trusing the component to use a condition. The inner condition will add the outer div for no reason

packages/amplication-client/src/User/Login.tsx
```diff
@@ -110,6 +105,7 @@ const Login = () => {
             </>
           ) : (
             <>
+              <LoginErrorMessage error={errorMessage} className={CLASS_NAME}/>
```
yuval-hazaz, line 108: same here about the condition

yuval-hazaz (changes requested):
@abrl91 great job, thanks
I added a few comments. Please review and we can discus o discord in needed

yuval-hazaz (approved):
@mshidlov Please merge with the correct branch to remove the differences in the workflow files. 
other than that, the PR is approved.
🔥
great job @abrl91

tupe12334:
We need to check how to auto tests for fork repos, this pr enter the release without the push validation and its enter a bug as i reported in #2229 
@mshidlov

tupe12334:
i started to work on fixing #2229 and i saw its not formated too

mshidlov:
@tupe12334 this is already fixed by @yuval-hazaz

PR #2221, opened 2022-01-26
Feat/2220 update issues template

mshidlov:


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -88,13 +77,9 @@ body:
       multiple: true
       options:
         - "Yes I am willing to submit a PR!"
-  - type: dropdown
+  - type: markdown
```
yuval-hazaz, line 80: Is there a reason you are using two markdown items where the first only include <br>? can't you use a multiline markdown?

yuval-hazaz (approved):
looks good

I added one question for my knowledge but we can proceed as it is

.github/ISSUE_TEMPLATE/bug.yaml
mshidlov: Its a none mandatory selection - instead of a check box that translates to task
image

mshidlov: @yuval-hazaz

PR #2254, opened 2022-02-02
fix(dsg): fix many to many data type

### LINKED ISSUE #2224: 🐛 Bug Report: incorrect data type used for many-to-many relation on REST API controller
### What happened?

The data type used in the generated app for many-to-many relations on the REST API controller is incorrect

```typescript
 @common.UseInterceptors(nestMorgan.MorganInterceptor("combined"))
  @common.UseGuards(
    defaultAuthGuard.DefaultAuthGuard,
    nestAccessControl.ACGuard
  )
  @common.Patch("/:id/tags")
  @nestAccessControl.UseRoles({
    resource: "Post",
    action: "update",
    possession: "any",
  })
  async updateTags(
    @common.Param() params: PostWhereUniqueInput,
    @common.Body() body: PostWhereUniqueInput[],
    @nestAccessControl.UserRoles() userRoles: string[]
  ): Promise<void> {
```

In this example, when we update tags of a post, the body should be of type `TagWhereUniqueInput` and not `PostWhereUniqueInput`

```typescript
    @common.Body() body: TagWhereUniqueInput[],
```

Fix the generated app to use the correct data type

### How to reproduce

see above

### Amplication version

0.11.1

### Environment

n/a

### Anything else

This bug does not have any actual impact on the API, since the type anyway include only a single property which is the same for all types

```typescript
class PostWhereUniqueInput {
  @ApiProperty({
    required: true,
    type: String,
  })
  @IsString()
  @Field(() => String)
  id!: string;
}
```

_No response_

### Code of Conduct

tupe12334:
resolve #2224


yuval-hazaz (approved):

yuval-hazaz:
@mshidlov should this go to 11.3 or to 12.0?

Issue #2204, opened 2022-01-19
🐛 Bug Report: Broken query sent to Prisma


0xsapphir3:
### What happened?

On having a `Relation to Entity` datatype for any of the field, it seems like I cannot get back the response for that specific field; it throws the following error:

```
Invalid `prisma.extension.findUnique()` invocation:

{
  where: {
    id: 'ckylksems0000aqmd4onoct2j'
  },
  select: {
    extensions: {
      where: {
        equals: 'true',
        ~~~~~~
        in: 'string',
        ~~
        notIn: 'string',
        ~~~~~
        lt: 'string',
        ~~
        lte: 'string',
        ~~~
        gt: 'string',
        ~~
        gte: 'string',
        ~~~
        contains: 'string',
        ~~~~~~~~
        startsWith: 'string',
        ~~~~~~~~~~
        endsWith: 'string',
        ~~~~~~~~
        mode: 'Default',
        ~~~~
        not: 'true',
        ~~~
        id: 'string'
      },
      select: {
        apiEndPoint: true,
        createdAt: true,
        icon: true,
        id: true,
        title: true,
        updatedAt: true,
        user: {
          select: {
            id: true
          }
        },
        verfied: true
      }
    }
  }
}
```

As Matan said, It looks like the query sent to the Prisma is broken.

### What you expected to happen

-

### How to reproduce

1. Create a new app
2. Create a new entity (say `X`)
3. Create a new field in User entity with the datatype as `Relation to Entity` and the entity related to be `X`.
4. Build the app
5. Sent a request to `https://www.<server_url>.com/api/users/{userID}/X`; be it a GET or POST or any other
6. See the error raised

### Amplication version

0.1.4

### Environment

Node version: v14.17.0
NPM version: 6.14.13
Docker Version: Docker version 20.10.12, build e91ed5707e
Cloud Provider: -
OS Details: 
```
NAME=EndeavourOS
PRETTY_NAME=EndeavourOS
ID=endeavouros
ID_LIKE=arch
BUILD_ID=2021.12.17
ANSI_COLOR="38;2;23;147;209"
HOME_URL='https://endeavouros.com'
DOCUMENTATION_URL='https://discovery.endeavouros.com'
SUPPORT_URL='https://forum.endeavouros.com'
BUG_REPORT_URL='https://forum.endeavouros.com/c/arch-based-related-questions/bug-reports'
LOGO=endeavouros
[...]

PR #2256, opened 2022-02-03
Fix/2204 prisma broken query

### LINKED ISSUE #2230: 🐛 Bug Report: Generated App - query for /api/entities/{id}/sub-entities is missing pagination
### What happened?

for example, we have two entities - posts, and tags. Where each post can have multiple tags.
We have two different endpoints to get a list of tags

/api/tags
/api/posts/{id}/tags

the first endpoint is using `TagFindManyArgs` to filter the tags which includes `where`, `orderBy`, `Skip`, and `Take`
the first endpoint is using `TagWhereInput` to filter the tags which only includes the inner properties for `where`

They both should use `TagFindManyArgs`

The two endpoints should use the same set of query parameters

### How to reproduce

create an app with two entities with relation and open the generated REST API code

### Amplication version

0.11.2

### Environment

n/a

### Anything else

_No response_

_No response_

### Code of Conduct

abrl91:
#2204 
This pr would also fix #2230 
The issue on nestjs/swagger repo: https://github.com/nestjs/swagger/issues/1803

A bug in nestjs/swagger generated the swagger UI in an incorrect way that cause the API request to fail. 

[[ @abrl91 please add a link to the issue you opened on the nestjs repo ]]

To fix this issue a new decorator (@ApiNestedQuery) was created to generate the nested properties of the query type as expected. 

In case this issue will be resolved in the future by the Nestjs team, we can remove the local implementation of ApiNestedQuery


packages/amplication-data-service-generator/src/server/resource/controller/to-many.template.ts
```diff
@@ -7,16 +7,21 @@ import * as defaultAuthGuard from "../auth/defaultAuth.guard";
...
+// @ts-ignore
+import {ApiNestedQuery} from "../../decorators/api-nested-query.decorator";
+import {plainToClass} from "class-transformer";
 
 declare interface WHERE_UNIQUE_INPUT {
   id: string;
 }
 declare interface RELATED_ENTITY_WHERE_UNIQUE_INPUT {}
-declare interface RELATED_ENTITY_WHERE_INPUT {}
+declare class RELATED_ENTITY_WHERE_INPUT {}
```
yuval-hazaz, line 18: This class is not being used and can be removed

yuval-hazaz (changes requested):
Great job Amit!!!

please see the comment for a minor change

yuval-hazaz:
@abrl91 please add a link to the issue you opened on the nestjs repo in the description of the PR above

packages/amplication-data-service-generator/src/server/resource/controller/to-many.template.ts
abrl91: This is weird. It's class on my file
image

@yuval-hazaz

abrl91: What should I do with line 28?
image

yuval-hazaz: you are correct, my mistake 👍

yuval-hazaz (approved):

Issue #2257, opened 2022-02-03
🐛 Bug Report: Generated app - no way to filter entity list by a relation


yuval-hazaz:
### What happened?

In the generated app, there is no way to filter a list of entities by a value of a related entity - 

for example - 
Relation between a Post and Tags. one Post can be related to many Tags. 

The user wants to get all post with tag X, but the class PostWhereInput does not have a property for Tags.

### What you expected to happen

The user should be able to filter a list by values on related entities. 

in Prisma they are using a class names "EntityListRelationFilter"

```typescript
export type PostWhereInput = {
    ...
    ...
    tags?: TagListRelationFilter
    ...
  }
```

```typescript
 export type TagListRelationFilter = {
    every?: TagWhereInput
    some?: TagWhereInput
    none?: TagWhereInput
  }
```

### How to reproduce

see example above

### Amplication version

0.11.2

### Environment

n/a

### Code of Conduct

I agree to follow this project's Code of Conduct

yuval-hazaz:
see this commit for more details on the implementation done to fix this issue on the blog-server 
https://github.com/amplication/blog-server/commit/f47c75f2bffc26734eae8db66f2843523f5ccc15

PR #2295, opened 2022-02-14
Fix/2257 filter entity by relation

abrl91:
#2257 

Added the ability to filter an entity by relation, for example: 
all the users with posts that have less than 50 likes

```
prisma.user.findMany({
  where: {
    posts: {
      some: {
        likes: {
          lte: 50,
        },
      },
    },
  },
})
```

@yuval-hazaz 
I made some changes that I am not sure are needed:
- to-many.template.ts
- create-resolver.ts
- create-controller.ts


packages/amplication-data-service-generator/src/server/resource/create-dtos.ts
```diff
@@ -31,6 +32,7 @@ type EntityDTOs = {
   createArgs?: NamedClassDeclaration;
   updateArgs?: NamedClassDeclaration;
   orderByInput: NamedClassDeclaration;
+  entityListRelationFilter: NamedClassDeclaration;
```
yuval-hazaz, line 35: rename to `listRelationFilter`

packages/amplication-data-service-generator/src/server/resource/dto/create-field-class-property.ts
```diff
@@ -266,7 +272,9 @@ export function createFieldClassProperty(
         TYPE_ID,
         builders.arrowFunctionExpression(
           [],
-          prismaField.isList ? builders.arrayExpression([typeName]) : typeName
+          prismaField.isList && isToManyRelationField(field) && !isQuery
```
yuval-hazaz, line 275: This seems incorrect - see the comment on the snapshot about "favoriteColors" - I think this is the cause for the unexpected behavior

packages/amplication-data-service-generator/src/tests/__snapshots__/create-data-service.spec.ts.snap
```diff
@@ -3456,7 +3505,7 @@ class Customer {
     each: true,
   })
   @IsOptional()
-  @Field(() => [EnumCustomerFavoriteColors], {
+  @Field(() => EnumCustomerFavoriteColors, {
```
yuval-hazaz, line 3508: This should not change
See the comment in packages/amplication-data-service-generator/src/server/resource/dto/create-field-class-property.ts

packages/amplication-data-service-generator/src/server/resource/dto/create-field-class-property.ts
```diff
@@ -340,7 +348,7 @@ function createGraphQLFieldType(
   entity: Entity,
   isQuery: boolean
 ): namedTypes.Identifier | namedTypes.ArrayExpression {
-  if (prismaField.isList) {
+  if (prismaField.isList && isToManyRelationField(field) && !isQuery) {
```
yuval-hazaz, line 351: This is incorrect, the change incorrectly exclude fields when isList && !isRelation

yuval-hazaz (changes requested):
@abrl91 impressive work here! well done

Please see my comments about the conditions 
Once we fix it, I will review again all other parts

packages/amplication-data-service-generator/src/server/resource/dto/create-field-class-property.ts
abrl91: we need that when it is prisma field list and to-many relation and not query that the type would not be inside an array,
otherwise, this condition is true. Then, on line 390 we have the following condition:

```
if (isToManyRelationField(field)) {
    return createEntityListRelationFilterID(prismaField.type);
  }
```

And as a result we got [EntityListRelationField]

packages/amplication-data-service-generator/src/tests/__snapshots__/create-data-service.spec.ts.snap
abrl91: I think I fixed it. I hope it didn't break something else. Every time I change a condition, it breaks another part.
If nobody wants to refactor this file, I would love to :)

packages/amplication-data-service-generator/src/server/resource/dto/create-field-class-property.ts
abrl91: this block code determines the type of the 'ApiProperty'

packages/amplication-data-service-generator/src/server/resource/dto/graphql/entity-list-relation-filter/create-entity-list-relation.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+describe("createEntityListRelation", () => {
+  test("create list relation", async () => {
+    expect(
+      print(
+        await createEntityListRelationFilter(
+          EXAMPLE_ENTITY,
+          EXAMPLE_WHERE_INPUT
+        )
+      ).code
+    ).toEqual(`@InputType()
```
tupe12334, line 32: Way not to use snapshot inline test instead?

tupe12334: https://jestjs.io/docs/snapshot-testing#inline-snapshots

abrl91: I saw all the other Input has test files, so I added one too

tupe12334: Yes, all the other inputs have a test that tests there AST object.
You made a good test (my preferred way ;) ) that checked the generation of the dto, i just suggest you a better way (in my opinion) to do a string compressing test, a snapshot test

abrl91: I am not sure why they say on the docs:
"Snapshot tests are a very useful tool whenever you want to make sure your UI does not change unexpectedly."
A template creates EntityListRelationFilter, and it has no reason to change unless someone changes the template.
Correct me if I am wrong

abrl91: I changed it to snapshot test. Thanks :)

packages/amplication-data-service-generator/src/server/seed/seed.template.ts
```diff
@@ -41,7 +41,7 @@ async function seed(bcryptSalt: Salt) {
     update: {},
     create: data,
   });
-  client.$disconnect();
+  void client.$disconnect();
```
yuval-hazaz, line 44: why did you add this?

abrl91: the lint screamed at me and asked me to add this 😅
look on [17004e5](https://github.com/amplication/amplication/pull/2295/commits/17004e528fad61d6124bdb5ce4c72facedeea14c)

yuval-hazaz (approved):
Great work @abrl91

Issue #2323, opened 2022-02-19
🐛 Bug Report: Tooltip text is not visible in light theme mode


molaycule:
### What happened?

In light theme mode, when hovering on a component that has tooltip, the tooltip text is not visible to the user.

### What you expected to happen

Tooltip text should be visible even in light theme mode.

![Screenshot (407)](https://user-images.githubusercontent.com/20028628/154815207-b3fddf07-e331-4256-afd7-e6b5fa2cd923.png)

### How to reproduce

Switch to light theme mode, then hover on any sidebar item that has tooltip, you would notice that the tooltip text is not visible.

### Amplication version

0.11.2

### Are you willing to submit PR?

Yes I am willing to submit a PR!

mshidlov:
Thank you @molaycule for reporting this issue and submitting a PR, we will review the PR soon

PR #2324, opened 2022-02-19
fix(client): tooltip text not visible in light mode

molaycule:
Issue Number: #2323

## PR Details

Added tooltip text color so that tooltip text can be visible in light theme mode

### DIFF
--- a/packages/amplication-design-system/src/components/Tooltip/Tooltip.scss
+++ b/packages/amplication-design-system/src/components/Tooltip/Tooltip.scss
@@ -5,7 +5,7 @@ $width-short: 100px;
 .amp-tooltip {
   &:after {
     background-color: var(--tooltip-background);
-    color: var(--tooltip-text-color);
+    color: var(--static-white);
   }
 
   &:before,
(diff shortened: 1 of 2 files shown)


packages/amplication-design-system/src/components/Tooltip/Tooltip.scss
```diff
@@ -5,6 +5,7 @@ $width-short: 100px;
 .amp-tooltip {
   &:after {
     background-color: var(--tooltip-background);
+    color: var(--tooltip-text-color);
```
yuval-hazaz, line 8:
```suggestion
    color: var(--static-white);
```

packages/amplication-design-system/src/style/css-variables.scss
```diff
@@ -44,6 +44,7 @@
   --negative: #e93c51;
 
   --tooltip-background: #1b1f23;
+  --tooltip-text-color: #ffffff;
```
yuval-hazaz, line 47:
```suggestion
```

yuval-hazaz (changes requested):
@molaycule thanks for the PR!

we have a CSS variable named --static-white for a fixed white color both on light and dark mode. please use it

yuval-hazaz (approved):

yuval-hazaz:
thanks @molaycule

yuval-hazaz:
https://github.com/all-contributors please add @molaycule for code, bug fixes

Issue #2268, opened 2022-02-04
🐛 Bug Report: Generated app - gqlUserRoles.decorator.ts fails for public queries 


yuval-hazaz:
### What happened?

The decorator `gqlUserRoles.decorator.ts` does not check for the existence of the user on the request and fails when there is no user in context

### What you expected to happen

The decorator should return null or an empty array when there is no user in context

### How to reproduce

See #2006 

### Amplication version

0.11.2

tupe12334:
This solution build from 2 steps

- [x]  : update our use of the user roles for use only nest access control package

(i did that in the #2325 pr)

- [ ]  update the nest access control package with the logic the prevent crashes

(i did that in https://github.com/nestjsx/nest-access-control/pull/71 pr, wait for approval)

PR #2325, opened 2022-02-20
Feat/update user roles

tupe12334:
Part one of the solution for #2268

## PR Checklist


packages/amplication-data-service-generator/src/server/static/src/auth/gqlUserRoles.decorator.ts
```diff
@@ -1,16 +0,0 @@
-import { createParamDecorator, ExecutionContext } from "@nestjs/common";
-import { GqlExecutionContext } from "@nestjs/graphql";
```
yuval-hazaz, line 2: Please explain why do you think we do not need this file and what is the alternative

packages/amplication-data-service-generator/src/server/resource/resolver/resolver.base.template.ts
```diff
@@ -1,19 +1,17 @@
 import * as common from "@nestjs/common";
 import * as graphql from "@nestjs/graphql";
 import * as apollo from "apollo-server-express";
-import * as nestAccessControl from "nest-access-control";
```
yuval-hazaz, line 4: Can you please explain what is the reason behind changing the syntax here?

PR #2344, opened 2022-02-21
Feat/prevent roles crash

tupe12334:
This pr is temporary resolving #2268


yuval-hazaz:
@tupe12334 why do you think this fix is temporary?

tupe12334:
@yuval-hazaz i fix that inside the lib we use nestjs-access-control, but we cant update the package version

Issue #2202, opened 2022-01-19
Converting an OAuth App to a GitHub App


mshidlov:
### Feature description

GitHub Apps are the officially recommended way to integrate with GitHub because they offer many advantages over a pure OAuth-based integration.

1. Create Amplication GitHub App in Amplication GitHub Organization.
2. Update Sync with GitHub to Create GitHub App instead of Auth App in the user-selected organization
3. Create in GitHub service ocktokit client that is in sync with Amplication GitHub App using the App secret key (Amplication ocktokit client).
4. change all GitHub API calls to created by octokit client created from Amplication ocktokit client using the users GitHub App id.

Instead of using the user Auth key, we will use the app id. each sync app gets its unique identifier, combined with the Amplication App secret key GitHub grants access to APIs as Amplication and not as the users themselves.

### Use case

* Fine-grained permissions target the specific information a GitHub App can access, allowing the app to be more widely used by people and organizations with security policies than OAuth Apps, which cannot be limited by permissions. 
* Short-lived tokens provide a more secure authentication method over OAuth tokens. An OAuth token does not expire until the person who authorized the OAuth App revokes the token. GitHub Apps use tokens that expire quickly, creating a much smaller window of time for compromised tokens to be in use. 
* Built-in, centralized webhooks receive events for all repositories and organizations the app can access. Conversely, OAuth Apps require configuring a webhook for each repository and organization accessible to the user. 
* Bot accounts don't consume a GitHub seat and remain installed even when the person who initially installed the app leaves the organization. 
* Built-in support for OAuth is still available to GitHub Apps using user-to-server endpoints. 

### Code of Conduct

I agree to follow this project's Code of Conduct

yuval-hazaz:
Related to #1187

mshidlov:
Relevant Resources:
[Authenticating as a GitHub App](https://docs.github.com/en/developers/apps/building-github-apps/authenticating-with-github-apps#authenticating-as-a-github-app)
[Direct users to install your GitHub App on repositories](https://docs.github.com/en/developers/apps/getting-started-with-apps/migrating-oauth-apps-to-github-apps#direct-users-to-install-your-github-app-on-repositories)


GitHub App installation callback URL example:
http://localhost:3001/github-auth-app/callback/%7BappId%7D?code=506bcf082b5d363c12f5&installation_id=22120743&setup_action=install

mshidlov:
Hey @morhag90 
These are the new tables for git sync data.
[...]

PR #2362, opened 2022-02-25
Feat (server): Converting an oauth app to GitHub app

yuval-hazaz:
Issue Number: #2202 

## PR Checklist


packages/amplication-client/src/Application/git/GitOrganizations/GitOrganizations.tsx
```diff
@@ -0,0 +1,66 @@
...
+  setSelectedGitOrganization:(gitOrganizationId:string)=>void
+};
+
+function GitOrganizations({ workspaceId ,setSelectedGitOrganization}: Props) {
+     const {
+    refetch,
+    gitOrganizations,
+    loading: loadingRepos,
+    networkStatus,
+  } = useGetGitOrganizations({
```
yuval-hazaz, line 24: I don't think it makes sense to create a hook for each server query. 
- There is no reuse of code here since it will probably not be used from anywhere else.
- The amount of code in this file is almost identical with or without the hook.. (instead of use ueGitOrganization, you just use useQuery) - the only code that was removed from this file is the actual gql...
- The name is not clear enough to allow anyone to ever re-use this hook
- putting gql in generic hooks breaks the benefit of GraphQL where each page can ask only for the data it needs. 
for example, a page need just the appId, and appName, and another page also need related entities with the app - so using a hook called "useApp" will eventually end up with a big query to cover all use cases where someone need an app with all fields and all relations.

packages/amplication-client/src/Application/git/dialogs/GitDialogsContainer.tsx
```diff
@@ -51,7 +53,8 @@ export default function GitDialogsContainer({
         onDismiss={handleSelectRepoDialogDismiss}
       >
         <GitRepos
-          applicationId={app.id}
+          applicationId = {app.id}
+          gitOrganizationId={gitOrganizationId}
           onCompleted={handleSelectRepoDialogDismiss}
           sourceControlService={sourceControlService}
```
yuval-hazaz, line 59: I think this should be called "gitProvider" but if it is too complicated to make the change now we may leave it as it is

packages/amplication-client/src/Application/git/hooks/useGetGitOrganizations.tsx
```diff
@@ -0,0 +1,37 @@
+import { EnumSourceControlService } from "@amplication/data/dist/models";
+import { gql, useQuery } from "@apollo/client";
+import { GitOrganization } from "../../../models";
```
yuval-hazaz, line 3: see my previous comment on hooks for server queries

packages/amplication-client/src/Application/git/hooks/useGetReposOfUser.tsx
```diff
@@ -2,19 +2,19 @@ import { gql, useQuery } from "@apollo/client";
 import { EnumSourceControlService, GitRepo } from "../../../models";
 
 type Props = {
-  appId: string;
```
yuval-hazaz, line 5: see my previous comment about hooks for server queries

packages/amplication-client/src/Application/git/hooks/useGitCreate.tsx
```diff
@@ -8,14 +8,16 @@ import {
 import { useTracking } from "../../../util/analytics";
 
 type Props = {
+  gitOrganizationId: string;
```
yuval-hazaz, line 11: see my previous comment about hooks for server queries

packages/amplication-client/src/Application/git/hooks/useGitUserName.tsx
```diff
@@ -2,22 +2,22 @@ import { gql, useQuery } from "@apollo/client";
 import { EnumSourceControlService } from "../../../models";
 
 type Props = {
-  appId: string;
+  gitOrganizationId: string;
```
yuval-hazaz, line 5: see my previous comment about hooks for server queries

packages/amplication-server/prisma/migrations/20220221165146_21022022_change_enum_type/migration.sql
```diff
@@ -0,0 +1,12 @@
+/*
+  Warnings:
```
yuval-hazaz, line 2: I recommend deleting all the migrations and create a single migration for all the changes in this PR - unless there is a real need for stepped migration

packages/amplication-server/prisma/schema.prisma
```diff
@@ -8,17 +8,17 @@ datasource db {
...
-  updatedAt     DateTime @updatedAt
-  email         String   @unique(map: "Account.email_unique")
-  firstName     String
-  lastName      String
-  password      String
-  currentUserId String?  @unique(map: "Account.currentUserId_unique")
-  githubId      String?
-  currentUser   User?    @relation(fields: [currentUserId], references: [id], onDelete: SetNull, onUpdate: Cascade)
-  users         User[]   @relation("AccountOnUser")
+  id               String            @id @default(cuid())
```
yuval-hazaz, line 11: we need to make sure we all use the same formating - it looks like there is no changes in this model and still there are new extra spaces

packages/amplication-server/prisma/schema.prisma
```diff
@@ -28,6 +28,29 @@ model Workspace {
...
   apps      App[]
   users     User[]
+  gitOrganizations GitOrganization[]
+}
+
+model GitOrganization {
+  id             String          @id @default(cuid())
+  provider       EnumSourceControlService
+  name           String
+  installationId Int
```
yuval-hazaz, line 38: is that really an integer? I think it is safer to use a string to be able to work with different providers

packages/amplication-server/prisma/schema.prisma
```diff
@@ -28,6 +28,29 @@ model Workspace {
...
+  createdAt      DateTime        @default(now())
+  updatedAt      DateTime        @updatedAt
+  gitRepositories  GitRepository[]
+  workspaceId    String
+  workspace      Workspace     @relation(fields: [workspaceId], references: [id], onDelete: SetNull, onUpdate: Cascade)
```
yuval-hazaz, line 43: onDelete: SetNull - this does not make any sense...

packages/amplication-server/prisma/schema.prisma
```diff
@@ -28,6 +28,29 @@ model Workspace {
...
+}
+
+model GitRepository {
+  id                String           @id @default(cuid())
+  name              String
+  app               App[]
+  createdAt         DateTime         @default(now())
+  updatedAt         DateTime         @updatedAt
+  gitOrganization   GitOrganization? @relation(fields: [gitOrganizationId], references: [id])
+  gitOrganizationId String?
```
yuval-hazaz, line 53: This should not be optional
```suggestion
  gitOrganizationId String
```

packages/amplication-server/prisma/schema.prisma
```diff
@@ -385,6 +410,10 @@ enum EnumEntityPermissionType {
   Disabled
 }
 
+enum EnumProvider {
```
yuval-hazaz, line 413: is this being used?

packages/amplication-server/prisma/schema.prisma
```diff
@@ -442,3 +471,7 @@ enum EnumDeploymentStatus {
   Failed
   Removed
 }
+
+enum EnumSourceControlService {
```
yuval-hazaz, line 475: consider changing the name to EnumGitProvider

packages/amplication-server/src/core/app/app.service.ts
```diff
@@ -577,39 +575,8 @@ export class AppService {
...
-      where: args.where,
-      data: {
-        githubToken: token,
-        githubTokenCreatedDate: new Date()
-      }
-    });
-  }
-
   async removeAuthorizeAppWithGithub(args: FindOneArgs): Promise<App> {
+    //todo:: need to remove
```
yuval-hazaz, line 579: please update the comment with more clear instrcutions or remove it

packages/amplication-server/src/core/app/app.service.ts
```diff
@@ -628,7 +595,7 @@ export class AppService {
     return this.prisma.app.update({
       where: args.where,
       data: {
-        githubToken: null,
+        //githubToken: null,
```
yuval-hazaz, line 598: remove this comment

packages/amplication-server/src/core/app/dto/CompleteAuthorizeAppWithGithubArgs.ts
```diff
@@ -1,12 +1,9 @@
 import { ArgsType, Field } from '@nestjs/graphql';
 import { CompleteAuthorizeAppWithGithubInput } from './CompleteAuthorizeAppWithGithubInput';
-import { WhereUniqueInput } from 'src/dto';
+//import { WhereUniqueInput } from 'src/dto';
```
yuval-hazaz, line 3: remove this comment

packages/amplication-server/src/core/git/dto/args/BaseGitArgs.ts
```diff
@@ -4,7 +4,9 @@ import { EnumSourceControlService } from '../enums/EnumSourceControlService';
 @ArgsType()
 export class BaseGitArgs {
```
yuval-hazaz, line 5: I don't think this class is helpful, and it is probably better to use the fields when needed directly in the class

packages/amplication-server/src/core/git/dto/args/BaseGitArgs.ts
```diff
@@ -4,7 +4,9 @@ import { EnumSourceControlService } from '../enums/EnumSourceControlService';
...
   @Field(() => String, { nullable: false })
-  appId;
+  gitOrganizationId;
+  @Field(() => String, { nullable: true })
+  workspaceId?;
```
yuval-hazaz, line 9: workspaceId should never be provided from the client

packages/amplication-server/src/core/git/dto/inputs/GetGitOrganizationsInput.ts
```diff
@@ -0,0 +1,14 @@
+import { ArgsType, Field, InputType } from '@nestjs/graphql';
+import { EnumSourceControlService } from '../enums/EnumSourceControlService';
+
+@InputType({
+  isAbstract: true
+})
+@ArgsType()
+export class GetGitOrganizationsInput {
+  @Field(() => String, { nullable: false })
```
yuval-hazaz, line 9: workspace should not be provided by the client

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
   async createRepoInOrg(@Args() args: CreateRepoArgs): Promise<GitRepo> {
     return this.gitService.createRepo(args);
   }
+
+  @Mutation(() => GitOrganization)
+  @AuthorizeContext(
+    AuthorizableResourceParameter.GitOrganizationId,
+    'gitOrganizationId'
+  )
+  async getOrganization(@Args() args: BaseGitArgs): Promise<GitOrganization> {
```
yuval-hazaz, line 37:
```suggestion
  async getGitOrganization(@Args() args: BaseGitArgs): Promise<GitOrganization> {
```

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
+    return this.gitService.getGitOrganization(args);
+  }
+
+  @Mutation(() => GitOrganization)
+  //@AuthorizeContext(AuthorizableResourceParameter.WorkspaceId, 'workspaceId')
+  async createOrganization(
```
yuval-hazaz, line 43:
```suggestion
  async createGitOrganization(
```

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
+  ): Promise<GitOrganization> {
+    return await this.gitService.createGitOrganization(args);
+  }
+
+  @Mutation(() => Boolean)
+  @AuthorizeContext(
+    AuthorizableResourceParameter.GitOrganizationId,
+    'gitOrganizationId'
+  )
+  async deleteOrganization(@Args() args: BaseGitArgs): Promise<boolean> {
```
yuval-hazaz, line 54:
```suggestion
  async deleteGitOrganization(@Args() args: BaseGitArgs): Promise<boolean> {
```

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
+    return this.gitService.deleteGitOrganization(args);
+  }
+
+  @Mutation(() => AuthorizeAppWithGithubResult)
+  //@AuthorizeContext(AuthorizableResourceParameter.WorkspaceId, 'workspaceId')
+  async getGithubAppInstallationUrl(
```
yuval-hazaz, line 60: This function should not get any parameters - since it is called "GitHub" it should use hardcoded value for "GitHub"

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
+
   //#endregion
   //#region Query
   @Query(() => [GitRepo])
-  @AuthorizeContext(AuthorizableResourceParameter.AppId, 'appId')
+  @AuthorizeContext(
+    AuthorizableResourceParameter.GitOrganizationId,
+    'gitOrganizationId'
+  )
   async getReposOfUser(@Args() args: GetReposListArgs): Promise<GitRepo[]> {
```
yuval-hazaz, line 75: Either change the name as suggested below - or better - use a generic query called FindGitRepositories which accepts parameters from the client
also, the name of the args should be the same as the name of the query

also, can a user get a list of repos that are not belong to them? why to call it "OfUser"?

```suggestion
  async getGitReposOfUser(@Args() args: GetReposListArgs): Promise<GitRepo[]> {
```

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -16,21 +21,82 @@ export class GitResolver {
...
     return this.gitService.getReposOfUser(args);
   }
+
+  @Query(() => [GitOrganization])
+  //@AuthorizeContext(AuthorizableResourceParameter.WorkspaceId, 'workspaceId')
```
yuval-hazaz, line 80: use InjectContext to get WorkspaceID

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -17,11 +15,24 @@ import { GitUser } from '../git/dto/objects/GitUser';
 import { GithubFile } from './dto/githubFile';
 import { GithubRepo } from './dto/githubRepo';
 import { GithubTokenExtractor } from './utils/tokenExtractor/githubTokenExtractor';
+import { CreateGitOrganizationArgs } from '../git/dto/args/CreateGitOrganizationArgs';
+import { PrismaService } from 'nestjs-prisma';
```
yuval-hazaz, line 19: github service (provider) should not be aware of Prisma or the DB.

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -35,8 +46,103 @@ export class GithubService implements IGitClient {
...
+    public readonly tokenExtractor: GithubTokenExtractor,
+    private readonly prisma: PrismaService
   ) {}
+  async getGitOrganizations(workspaceId: string): Promise<GitOrganization[]> {
+    return await this.prisma.gitOrganization.findMany({
+      where: {
+        workspaceId: workspaceId
+      }
+    });
+  }
```
yuval-hazaz, line 58: this shoud be part of git.service

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -35,8 +46,103 @@ export class GithubService implements IGitClient {
...
+
+    const gitOrganization = await this.getGitOrganization(gitOrganizationId);
+    if (isEmpty(gitOrganization)) {
+      throw new Error('INVALID_GITHUB_APP');
+    }
+
+    await this.prisma.gitOrganization.delete({
+      where: {
+        id: gitOrganization.id
+      }
```
yuval-hazaz, line 85: This should be part of git.service

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -35,8 +46,103 @@ export class GithubService implements IGitClient {
...
+    );
+    gitInstallationUrl = gitInstallationUrl.replace('{state}', workspaceId);
+    return gitInstallationUrl;
+  }
+
+  async getGitOrganization(
+    gitOrganizationId: string
+  ): Promise<GitOrganization> {
+    const gitOrganization = new GitOrganization();
+    const res = await this.prisma.gitOrganization
```
yuval-hazaz, line 103: this should be part of git.service

yuval-hazaz (commented):
I did the review until and not included packages/amplication-server/src/core/github/github.service.ts

I still need to continue but please review the comments I added so far

packages/amplication-server/src/core/github/github.service.ts
morhag90: Done

packages/amplication-server/src/core/github/github.service.ts
morhag90: Done

yuval-hazaz (commented):
please see comments

packages/amplication-server/src/core/git/git.service.ts
```diff
@@ -38,13 +38,18 @@ export class GitService {
...
-
-    return app;
+    if (remoteRepository) {
+      return await this.connectAppGitRepository({ ...args });
+    }
+    return await this.prisma.app.findUnique({
+      where: {
+        id: args.appId
+      }
+    });
```
yuval-hazaz, line 52: I think this should throw an error

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -102,13 +102,22 @@ export class GithubService implements IGitClient {
       parseInt(data.installationId)
     );
 
+    const installation = await octokit.rest.apps.getInstallation({
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      installation_id: parseInt(data.installationId)
+    });
+
+    if (installation.data.account.type.toLowerCase() === 'user') {
```
yuval-hazaz, line 110: use const GITHUB_ACCONT_TYPE_USER

packages/amplication-server/src/core/github/github.service.ts
```diff
@@ -102,13 +102,22 @@ export class GithubService implements IGitClient {
...
+      return null;
```
yuval-hazaz, line 111: I think we should throw an "Unsupported Operation Exception" or something similar 
Or - at least add a comment that creating repo for personal account is not support with a link to a new GitHub Issue

yuval-hazaz (commented):
@morhag90 please see my comments

Issue #2370, opened 2022-03-01
🐛 Bug Report: Roles can't be deleted


molaycule:
### What happened?

Roles added can't be deleted

### What you expected to happen

I should be able to delete a role that is no longer needed or mistakenly created

### How to reproduce

When user add a role, there is no delete button to remove that role.

### Amplication version

0.11.4

### Are you willing to submit PR?

Yes I am willing to submit a PR!

molaycule:
@yuval-hazaz is there a way to know the roles or keep track of the roles that are actively used for entity permission, so that such roles can't be deleted only roles created by mistake or inactive roles can be deleted.

yuval-hazaz:
this issue still requires a design - deleting a role has impacts that need to be carefully considered. 
We should keep this issue open for future reference

PR #2419, opened 2022-03-10
fix(client): roles can be deleted

molaycule:
Issue Number: #2370

## PR Details

Roles created on the client app can now be deleted

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Roles/DeleteRoleField.tsx
+++ b/packages/amplication-client/src/Roles/DeleteRoleField.tsx
@@ -1,0 +1,83 @@
+import React, { useState, useCallback } from "react";
+import { gql, useMutation } from "@apollo/client";
+import { Button, EnumButtonStyle } from "../Components/Button";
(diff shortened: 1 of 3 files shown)


yuval-hazaz:
@molaycule we still need to think over this PR and decide how we want to handle the deletion of roles. 
I will keep you updated soon, we can keep the convo going on discord until then

tupe12334:
@yuval-hazaz Can you please provide more info for @molaycule ?
I want to clean up some unused or irrelevant PR's

yuval-hazaz:
I didn't see much feedback from the community about asking for this ability, so I suggest that we close this PR. 
Seleting roles has a much bigger impact than just deleting the record, and for now, I think we better keep the focus on other areas instead of dealing with it.

yuval-hazaz:
@molaycule thank you very much for your efforts. Unfortunately this time we cannot accept this PR

Issue #2412, opened 2022-03-09
🐛 Bug Report: The dropdown icon is shown incorrectly


alexbass86:
### What happened?

The arrow icon in our dropdown field is in the wrong place. 
It should be on the right side.

Currently, it is on the left side like this:
<img width="346" alt="image" src="https://user-images.githubusercontent.com/96179897/157485876-0d294e8f-94b9-4e1a-85d2-311cbc7cec42.png">

### What you expected to happen

It should be on the right side:
<img width="297" alt="image" src="https://user-images.githubusercontent.com/96179897/157486058-fabc3a49-40fa-44e6-815c-542eed838c2b.png">

Figma link
https://www.figma.com/file/tIfcueoxU4DtYtCHmmLuE7/Community---Task-Assets?node-id=1%3A4515

### How to reproduce

Click on the GitHub icon on the side menu, and it will probably happen in all places where we have a dropdown/select button.

### Amplication version

0.11.2

### Environment

.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

jainpawan21:
Hi @mshidlov,
I would like to work on this, Can you please assign this to me?

yuval-hazaz:
@jainpawan21 yes you can. 
Please note that this control is in @amplication/designs-system so before you make the change please remember that it may impact multiple places. 

best way to approach would be to add an optional parameter with the position of the icon (left/right) with the default in current place, and use the other value for the specific location where we need to make the change currently

PR #2454, opened 2022-03-18
fix: add icon position in button

jainpawan21:
Issue Number: #2412 

## PR Details

1. Add iconPosition props with the right position as default value
2. Add EnumIconPosition with values Left and Right

## PR Checklist


packages/amplication-design-system/src/components/Button/Button.tsx
```diff
@@ -65,7 +74,7 @@ export const Button = ({
           className="amp-button__icon"
         />
       )}
-      {children}
+      {iconPosition === EnumIconPosition.Left && children}
```
yuval-hazaz, line 77: if Right is the default (to keep the current behavior) then I would expect to see no change when Right is used.
I think the current options are in reverse
Isn't that the correct option here?
```suggestion
      {iconPosition === EnumIconPosition.Right && children}
```

yuval-hazaz (commented):
Thanks!
please see my question in the review

packages/amplication-design-system/src/components/Button/Button.tsx
jainpawan21: if iconPosition === "left" then icon will should be in left and {children} should be in right direction
as line no. 77 is after icon so the icon will be rendered in the left direction and text(children) will be in right direction;

yuval-hazaz: you are correct. I thought that the icon is currently on the right but it is on the left... all good!

package-lock.json
```diff
@@ -4,7 +4,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "amplication",
```
tupe12334, line 7: Why did you deleted the name in the package-lock?

jainpawan21:
@mshidlov I reverted back package-lock.json changes.
Please review again.

mshidlov:
@jainpawan21 thank you 👍

yuval-hazaz (approved):

packages/amplication-design-system/src/components/Button/Button.tsx
tupe12334, line 77: What do you do if you don't have icon at all?

tupe12334 (approved):

PR #2485, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
## PR Details

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## PR Checklist


tupe12334 (approved):

mshidlov:
@CommanderRoot Thank you for submitting this PR 
purple-heart

Issue #2486, opened 2022-03-22
`@amplication/server` Phone-Home


mshidlov:
### Feature description

To help the Amplicaion team understand and be more aware of `@amplication/server` usage a notification event should be added on server load.

The loading event should not include any private personal data.
The event should include only:

- Build version
- Server version
- Runtime environments (OS, node version, ...)

### Use case

When `@amplication/server` is loaded it should send a loading notification message.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

MoFoGo:
@mshidlov - please supply details of how to deactivate the feature.

PR #2488, opened 2022-03-22
feat(server): adding phone-home functionality: update terraform and G…

mshidlov:
Issue Number: #2486

## PR Details

adding phone-home functionality to amplication-server: update terraform and GitHub actions

## PR Checklist


packages/amplication-server/src/util/sendServerLoadEvent.ts
```diff
@@ -0,0 +1,58 @@
...
+        nodeVersion: NODE_VERSION,
+        osName: OS_NAME,
+        osVersion: OS_VERSION,
+        hostName: HOST_NAME,
+        timezone: TIMEZONE
+      },
+      timestamp: new Date()
+    };
+
+    console.dir(data);
```
yuval-hazaz, line 42: please remove

packages/amplication-server/src/util/sendServerLoadEvent.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const sendServerLoadEvent = (): void => {
+  void getServerId().then(runtimeId => {
+    const data = {
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      api_key: POSTHOG_ID,
+      event: 'server-load-event',
+      properties: {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        distinct_id: runtimeId,
```
yuval-hazaz, line 30: This value is not shown in posthog

terraform/modules/env_baseline/main.tf
```diff
@@ -305,6 +305,11 @@ resource "google_cloud_run_service" "default" {
           value = var.github_app_installation_url
         }
 
+        env {
+          name  = "DISABLE_EVENT_TRACKING"
+          value = 0
```
yuval-hazaz, line 310: I think we should disable this event for our deployments - not to be mixed with real data

yuval-hazaz (changes requested):
please see my comments

terraform/modules/env_baseline/main.tf
mshidlov: sure this was for testing before I will merge I will change it to 1

packages/amplication-server/src/util/sendServerLoadEvent.ts
mshidlov: Thanks

PR #2506, opened 2022-03-25
hotfix(dsg) react admin ui fix source attribute

### LINKED ISSUE #2505: 🐛 Bug Report: React Admin many to many relation - source attribute get wrong value
### What happened?

Data of m2m relation is sometimes lost. It depends on the relation entity name now, but it does not need to be affected.

A user can call the entity's field however he wants and still get m2m relation. Then on react admin UI when he set the relation data, the values should be saved.

### How to reproduce

After setting m2m relation on Amplication, generate an app and run it. On react admin, save the data on both sides of the related entities. Depending on how you called your fields (and this is the bug) - the data might not be saved.

### Amplication version

0.12.2

### Environment

_No response_

abrl91:
fix #2505 

changed the source attribute so that it will not depend on the relation entity name, but on the field name.

image

## PR Checklist


abrl91:
@mshidlov hotfix to master for react admin bug many to many, please merge it ASAP

yuval-hazaz (approved):

yuval-hazaz:
@abrl91 please upload to staging and confirm that it works as expected there

PR #2511, opened 2022-03-27
Refactoring generic helm-chart 

shaharblanksela:
This helm-chart will replace the existing chart on ./chart/amplication
The approach is to be generic as possible. 
As you can see the templates are generic and have conditions for each resource.
The installation of the helm-chart will be specified by the values.yaml for each service. (lookup in folder ./values)
helm install --create-namespace --namespace [NAMESPACE] -f [PATH_TO_SERVICE_VALUES_FILE] [RELEASE] [CHART]

For the environment, we will use '--set'  to override values. (There is 'environment' value for each service)
Example: helm install --namespace [NAMESPACE] --set environment=test -f [PATH_TO_SERVICE_VALUES_FILE] [RELEASE] [CHART]


chart/traefik/templates/ingress.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.traefik.ingress.alb.enabled }}
```
mshidlov, line 1: @shaharblanksela 
@germanilia 
@ItzhakLevi 

Hey Shachar I know we talked and you told me that you are using the existing traefik ingress controller we created. please notice that this is an ALB and we discussed on changing it to ELB.

mshidlov: What about local development ingress??? If I'm not part of Amplication and I want to run the charts? or part of Amplication but just want to run locally

Issue #2556, opened 2022-04-08
DSG: Add util methods to handle decorators 


yuval-hazaz:
### Feature description

To be able to manipulate decorators in the DSG we add two new util methods 

findFirstDecoratorByName 
removeObjectPropertyByName

PR #2557, opened 2022-04-08
feat(dsg): add util methods to ast.ts

yuval-hazaz:
Issue Number: #2556 

## PR Details

This PR adds two new util methods to help to manipulate decorators in the DSG

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/util/ast.ts
+++ b/packages/amplication-data-service-generator/src/util/ast.ts
@@ -769,3 +769,70 @@ export function createGenericArray(
     builders.tsTypeParameterInstantiation([itemType])
   );
 }
+
+/**
+ * Returns the first decorator with a specific name from the given AST
+ * @param ast the AST to return the decorator from
+ */
+export function findFirstDecoratorByName(
+  node: ASTNode,
+  decoratorName: string
+): namedTypes.Decorator {
+  let decorator: namedTypes.ClassDeclaration | null = null;
+  recast.visit(node, {
+    visitDecorator(path) {
+      const callee = path.get("expression", "callee");
+      if (callee.value && callee.value.name === decoratorName) {
+        decorator = path.value;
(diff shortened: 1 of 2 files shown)


yuval-hazaz:
@tupe12334 @mshidlov 
Please see if you can suggest any improvement to the implementation of the new functions 
They will be used in another PR, but I included them in a separate PR so we can review it from a generic perspective first

packages/amplication-data-service-generator/src/util/ast.ts
```diff
@@ -769,3 +769,70 @@ export function createGenericArray(
...
+      return this.traverse(path);
+    },
+    // Recast has a bug of traversing class property decorators
+    // This method fixes it
+    visitClassProperty(path) {
+      const childPath = path.get("decorators");
+      if (childPath.value) {
+        this.traverse(childPath);
+      }
+      this.traverse(path);
```
abrl91, line 807: don't need to return here?

yuval-hazaz: The reason we implement visitClassProperty and visitClassDeclaration is to make the visit method traverse into decorators - there is a bug in the package.
So the idea is just to check if there are any decorators to the class or property and traverse into it

tupe12334:
@yuval-hazaz is it not better to export those functions to separated files for more readability?

yuval-hazaz:
> @yuval-hazaz is it not better to export those functions to separated files for more readability?

Thanks Ofek, it may be better - we can consider extracting all functions to separate files in a dedicated folder. Actually, in the following PR #2558 you can see that one of the function was already moved to another PR

tupe12334:
@yuval-hazaz Please check if you need the branch and if not please delete it to prevent piling up branches

yuval-hazaz:
thansk @tupe12334, done

Issue #2563, opened 2022-04-10
Click to copy token to clipboard


molaycule:
### Feature description

Have a button to copy new token to the user clipboard for easy convenience.

### Use case

Presently, user have to select the whole token text and right click to copy or press `Ctrl + C`. But with a button to copy the token text to clipboard, it offers better user experience.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #2564, opened 2022-04-10
feat(client): button to copy new token to clipboard

molaycule:
Issue Number: #2563

## PR Details

Added a button to copy new token created to clipboard.


yuval-hazaz (commented):
Thanks! looks good

yuval-hazaz (approved):

Issue #2006, opened 2021-11-17
🚀 Feature: Creating models without auth guard


mikenussbaumer:
### 🔖 Feature description

As a user of amplication it should be possible to create models where no auth guards are needed for specific request types (GET, POST, PATCH, DELETE, ...) It should be possible to give every request type a different level of authorization. 

### 🎤 Pitch

This feature is extremely useful if data is accessible (GET request) to everybody in a frontend application but should only be edited from a user with permissions on the backend.

Example for a news app:
GET /api/news - should be accessible for everyone without a token
POST, PATCH, DELETE /api/news - should only be accessible by authorized users

### ✌️ Solution

Add the option to remove guards for the request type.

### 🔄️ Alternative

You can customize this in code but as @mshidlov told me:

"The override is possible, it's like any other override. But to override and bypass the authorization gaurd it a bit tricky. This is definitely possible but it needs a little bit work."

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/amplication/amplication/blob/master/code_of_conduct.md)

- [x]  Add Public to to EnumEntityPermissionType model, enum and Prisma + migrate. 
Add the 'Public' option to Amplication client and make sure that clicking on Public sends the request to the server with  the correct action(view/create/update...) and type (public)
- [x] POC on the generated app: try to create an interceptor for permissions and clean the implementation from the function's body. (different interceptor / implementation for 'create' entity)
https://github.com/amplication/amplication/pull/2636
https://github.com/amplication/amplication/pull/2638

another issue for reference: #2652 

- [x] On dsg: `create-controlller' - make sure that when the '@Public' decorator is added, and remove:
- UseInterceptors
- nestAccessControl.UseRoles
- [x] #2686
- UseInterceptors
- nestAccessControl.UseRoles

- [x] #2687
- [x] refactor and clean up implementaion

mikenussbaumer:
# Temporary solution 

I found a temporary solution for the problem with @ofek 

First I created the @Public decorator file  `public.decorator.ts` in the auth folder:

```
import { SetMetadata } from '@nestjs/common';
export const Public = () => SetMetadata('isPublic', true);
```

Then I edited the `defaultAuth.guard.ts` file to catch all routes where the @Public decorator is used:

```
import { ExecutionContext } from "@nestjs/common";
[...]

PR #2565, opened 2022-04-11
feat(amplication-client+server) added Public to EnumEntityPermissionT…

abrl91:
Issue Number: #2006 

The first part of this task:
Make sure the client is showing the option of public routes for every action and sending the correct values to the server.
The server has to recognize those values (in this case:  the new type - public). In order for it to happen, I added 'Public' to EnumEntityPermissionType enum and model and also to Prisma + migrate.

image

image

## PR Checklist


packages/amplication-client/src/Permissions/EntityPermissionAction.tsx
```diff
@@ -230,6 +231,10 @@ export const EntityPermissionAction = ({
                 <span className={`${CLASS_NAME}__action-summary`}>
                   {selectedRoleIds.size} roles selected
                 </span>
+              ) : permission.type === models.EnumEntityPermissionType.Public ? (
+                <span className={`${CLASS_NAME}__action-summary`}>
+                  Public endpoint selected
```
yuval-hazaz, line 236:
```suggestion
                  This action is public 
```

yuval-hazaz (changes requested):
Beside a small update in the display text, looks perfect!

packages/amplication-data-service-generator/src/server/create-grants.ts
```diff
@@ -92,6 +93,18 @@ export function createGrants(entities: Entity[], roles: Role[]): Grant[] {
           }
           break;
         }
+        case EnumEntityPermissionType.Public: {
+          for (const role of roles) {
+            grants.push({
+              role: role.name,
+              resource: entity.name,
+              action: actionToACLAction[permission.action],
+              /** @todo */
```
tupe12334, line 102: What is this TODO?

abrl91: This is a copy & paste from the AllRoles case for now until we know how to handle this case. When I only `break` on this case the, I am getting errors on react admin after login

abrl91:
will open a new branch and a new pr to be more organized.

Issue #2595, opened 2022-04-18
`@amplication/server` - Update build flow to commit only changed files


mshidlov:
Update build flow that only files that were changed from the previews build will be included in the new build pull request.
To accomplish that please write all generated files to a drive (EFS - will be managed in the helm chart).
Compare the generated files in the new folder with the last build files and include in the pull request only changed files.

As part of this task create a new service `@amplication/pull-request-service`.
the new service will receive the 2 folders will compare them for changes and will create a pull-request with the changed files.

![image](https://user-images.githubusercontent.com/91742238/163792887-09a48079-afbf-4533-96ad-fb832d67308e.png)

PR #2596, opened 2022-04-18
Feat/pr service

tupe12334:
Fix: #2595 

## PR Details

This pr is making a new service that responsible on the pr opening.

## PR Checklist
- [X] Tests for the changes have been added
- [X] `npm test` doesn't throw any error


packages/amplication-server/src/core/build/utils/previousBuild/previousBuild.ts
```diff
@@ -0,0 +1,24 @@
...
+      skip: 0,
+      where: {
+        appId: appId,
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        NOT: { id: newBuildId },
+        createdAt: { lt: buildDate }
+      },
+      orderBy: { createdAt: 'desc' }
+    })
+  ).shift();
```
mshidlov, line 22: `skip:0` is the default value so you don't have to use it.
New build-id created-at should be equal to build-date,
so newBuild.createdAt < buildDate will always be false.
So the `NOT: { id: newBuildId }` is not mandatory.

Issue #2597, opened 2022-04-18
🐛 Bug Report: In-app settings, all screens should be in same structure


alexbass86:
### What happened?

On some pages, we have a different inner page layout. To be consistent and allow our users to have smoother interaction, we need to make sure we won't have any strange jumps between the section, for example:
**App Settings** here we have an inner padding 
![image](https://user-images.githubusercontent.com/96179897/163811488-30533201-d18c-4e60-9596-6e8a6d06f3c6.png)

While in DB Settings, we dont have any padding
![image](https://user-images.githubusercontent.com/96179897/163811694-21788492-299b-4d07-8be9-74287a7b4eab.png)

The problematic screen currently is DB Settings

### What you expected to happen

We need to make sure all pages in the settings have the same pending as in "App Settings"

### How to reproduce

Just go to App Settings

### Amplication version

0.11.2

### Environment

.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

molaycule:
@alexbass86 can I submit a PR?

souravjain540:
Assigned @molaycule, let us know if you need some help :)

best of luck!

molaycule:
Hi @souravjain540 thanks. I just submitted a PR that might close this issue.

PR #2612, opened 2022-04-20
Client - Fixed database settings form padding

molaycule:
Issue Number: #2597

## PR Details

Fixed app db settings form padding to match structure of other app settings forms


tupe12334:
@molaycule I will check it today in the night or tomorrow afternoon, Just a tip, part of the naming convention for new branches is to add the issue number after the topic for example <topic>/<number>-<title>
You are the best! Many thanks!

molaycule:
Hi @tupe12334 thanks for the tip, do you mean something like this `fix/2597-db-settings-form-padding` going forward?

tupe12334 (approved):

Issue #2614, opened 2022-04-21
🐛 Bug Report: Typescript error for urlError on Login.tsx line 91


rolwin100:
### What happened?

Line 91 on Login.tsx is throwing lint error fro type safety.

### What you expected to happen

urlError should be typecast to string

### How to reproduce

Check line 91 with eslint configured on VS code.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #2615, opened 2022-04-21
fix(client): added type safety for urlError

rolwin100:
Issue Number: #2614

## PR Details

Fixed issue regarding typescript error on Login.tsx at line 91


souravjain540:
Thanks for the PR :)

tupe12334:
@rolwin100 Thanks for the pr, just for you to know, this doesn't type safety, you are not checking if it's a string, you declare it as a string.
And please check my review of the branch name to @molaycule in https://github.com/amplication/amplication/pull/2612#issuecomment-1104186444

rolwin100:
@tupe12334 can you pls tell me what can be done about this.

tupe12334:
The root of the problem is from lines 65-70 when you generate the urlError, we don't check that the ```params.error``` is a string and only string so the declare type is ```string | (string | null)[] | null``` a bit confusing but it means string or null or array of string or null, so we need to check if the error is string there and if not return a null
@rolwin100

Issue #2623, opened 2022-04-24
🐛 Bug Report:  React Admin one to many relation - source attribute get wrong value


Juicpt:
### What happened?

Data of o2m relation cannot be saved in entity multi-word names

### What you expected to happen

Data of o2m relation can be saved in entity multi-word names

### How to reproduce

After setting o2m relation on Amplication, generate an app and run it. Data cannot be saved , when the entity name is more than one word

### Amplication version

0.12.4

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #2624, opened 2022-04-24
fix:react admin ui fix source attribute

Juicpt:
Issue Number: #2623 

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/admin/entity/create-field-input.ts
+++ b/packages/amplication-data-service-generator/src/admin/entity/create-field-input.ts
@@ -55,7 +55,7 @@ const DATA_TYPE_TO_FIELD_INPUT: {
         <SelectArrayInput optionText={${relatedEntity.name}Title} />
       </ReferenceArrayInput>`;
     }
-    return jsxElement`<ReferenceInput source="${relatedEntity.name.toLowerCase()}.id" reference="${
+    return jsxElement`<ReferenceInput source="${field.name}.id" reference="${
       relatedEntity.name
     }" label="${field.displayName}">
       <SelectInput optionText={${relatedEntity.name}Title} />
--- a/packages/amplication-data-service-generator/src/admin/entity/create-field-value.ts
+++ b/packages/amplication-data-service-generator/src/admin/entity/create-field-value.ts
@@ -21,7 +21,7 @@ export function createFieldValue(
       const { relatedEntity } = field.properties as LookupResolvedProperties;
       return jsxElement`<ReferenceField label="${
         field.displayName
-      }" source="${relatedEntity.name.toLowerCase()}.id" reference="${
+      }" source="${field.name}.id" reference="${
         relatedEntity.name
       }">
             <TextField source={${relatedEntity.name.toUpperCase()}_TITLE_FIELD} /> 
--- a/packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
+++ b/packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
@@ -140,8 +140,8 @@ function createToManyReferenceField(
   const element = jsxElement` 
   <ReferenceManyField
     reference="${field.properties.relatedEntity.name}"
-    target="${entity.name}Id"
-    label="${field.properties.relatedEntity.pluralDisplayName}"
+    target="${field.properties.relatedField.name}Id"
+    label="${field.displayName}"
   >
     <Datagrid rowClick="show">
       ${fields.map((field) => jsxElement`${createFieldValue(field)}`)}
--- a/packages/amplication-data-service-generator/src/tests/__snapshots__/create-data-service.spec.ts.snap
+++ b/packages/amplication-data-service-generator/src/tests/__snapshots__/create-data-service.spec.ts.snap
@@ -1203,7 +1203,7 @@ export const CustomerCreate = (props: CreateProps): React.ReactElement => {
           <SelectInput optionText={OrganizationTitle} />
         </ReferenceInput>
         <ReferenceInput
-          source=\\"organization.id\\"
+          source=\\"vipOrganization.id\\"
           reference=\\"Organization\\"
           label=\\"VIP Organization\\"
         >
@@ -1298,7 +1298,7 @@ export const CustomerEdit = (props: EditProps): React.ReactElement => {
           <SelectInput optionText={OrganizationTitle} />
         </ReferenceInput>
         <ReferenceInput
-          source=\\"organization.id\\"
+          source=\\"vipOrganization.id\\"
           reference=\\"Organization\\"
           label=\\"VIP Organization\\"
         >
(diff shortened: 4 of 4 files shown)


yuval-hazaz:
@Juicpt thanks for this PR. 
@abrl91 can you please have a look at this?

abrl91:
@yuval-hazaz sure, I will check it.
@Juicpt thank you :)

abrl91:
@Juicpt Greate work!

@mshidlov I checked it locally. works fine. can be merged to release 0.12.6

abrl91 (dismissed):

packages/amplication-data-service-generator/src/admin/entity/create-field-input.ts
```diff
@@ -55,9 +55,7 @@ const DATA_TYPE_TO_FIELD_INPUT: {
         <SelectArrayInput optionText={${relatedEntity.name}Title} />
       </ReferenceArrayInput>`;
     }
-    return jsxElement`<ReferenceInput source="${relatedEntity.name.toLowerCase()}.id" reference="${
-      relatedEntity.name
-    }" label="${field.displayName}">
+    return jsxElement`<ReferenceInput source="${field.name}.id" reference="${relatedEntity.name}" label="${field.displayName}">
```
yuval-hazaz, line 58: are you sure the `.id` is required here?

```suggestion
    return jsxElement`<ReferenceInput source="${field.name}" reference="${relatedEntity.name}" label="${field.displayName}">
```

packages/amplication-data-service-generator/src/admin/entity/create-field-value.ts
```diff
@@ -19,11 +19,9 @@ export function createFieldValue(
...
     case EnumDataType.Lookup:
       const { relatedEntity } = field.properties as LookupResolvedProperties;
-      return jsxElement`<ReferenceField label="${
-        field.displayName
-      }" source="${relatedEntity.name.toLowerCase()}.id" reference="${
-        relatedEntity.name
-      }">
+      return jsxElement`<ReferenceField label="${field.displayName}" source="${
+        field.name
+      }.id" reference="${relatedEntity.name}">
```
yuval-hazaz, line 24: same here
```suggestion
      }" reference="${relatedEntity.name}">
```

packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
```diff
@@ -140,8 +140,8 @@ function createToManyReferenceField(
   const element = jsxElement` 
   <ReferenceManyField
     reference="${field.properties.relatedEntity.name}"
-    target="${entity.name}Id"
-    label="${field.properties.relatedEntity.pluralDisplayName}"
+    target="${field.properties.relatedField.name}Id"
```
yuval-hazaz, line 143: we need to double-check if this concatenation of the related field name and "Id" will always work.

yuval-hazaz (changes requested):
@Juicpt & @abrl91 can you please review my comments and see if they are making any sense?

packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
abrl91: I will check again

packages/amplication-data-service-generator/src/admin/entity/create-field-input.ts
abrl91: @yuval-hazaz why do you think that the `.id' can be removed?

packages/amplication-data-service-generator/src/admin/entity/create-field-value.ts
abrl91: @yuval-hazaz same here about the id?

packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
Juicpt: The id here are handled at https://github.com/amplication/ra-data-graphql-amplication/blob/5852c1dfe5a95da7e2776fa441ca3243206ca96d/src/buildVariables.ts#L150.

Juicpt: id is required elsewhere because field.name represents an object

packages/amplication-data-service-generator/src/admin/entity/create-field-input.ts
yuval-hazaz: 1. the source property on all other types is only using {field.name} without ID. 
2. field name can be "VIPCustomer" while the entity name can be "Customer" - so I am not sure if the usage of VIPCustomer.Id is correct. 

I am suggesting testing it carefully and making sure whether it is correct or not.

packages/amplication-data-service-generator/src/admin/entity/entity-show-component/create-entity-show-component.ts
yuval-hazaz: my concern is about the concatenation of "realtedField.name" with "Id" - because if I understand correctly - the related field name should already include the Id suffix. I could be wrong - I just want to make sure we verify that

yuval-hazaz:
@abrl91 following our discussion we agreed that this PR still need to be worked on.
I suggest that you record all your finding (about the special use cases) on the issue itself.

in case @Juicpt wants to follow up on that, we can keep the PR open and flag it as draft, until he makes progress. 
Otherwise, we can close this PR and leave the issue open until we prioritize it to be worked on

abrl91:
@Juicpt @yuval-hazaz 
According to what I found, when you set one-to-many relation (for example one author can have many books), react admin uses: `<ReferenceManyField >` tag.
This tag uses two attributes: `reference` and `target`. The `reference` tag is related to the 'many' side of the relation, and the `target` to the 'one' side (and the current record, meaning on admin-ui the show component of the author entity) of the relation, but as a foreign key field.

Therefore, the code should be something like: `<ReferenceManyField reference="books" target="authorId">`

But, it depends on the field name that we set for the foreign key on Amplication

image


Screen Shot 2022-05-10 at 15 13 31

In addition, inside `<ReferenceManyField>` can take two kinds of fields: TextField / DataField which takes the `source` attribute. This source is represented (again) by the **field** that we set on Amplication.

image

All works fine in we follow this convention of foreign key,  but we cannot be sure that our users follow this convention. If a user decides to set his  foreign key field to `author aaa` for example, this logic breaks.

We should find a solution for this part (`<ReferenceManyField>`)  specifically.
On the snapshot test of this PR we can see that the changes affected other parts (like `<ReferenceField>`) and I think it shouldn't.

Just a last few words to be clear, as much as I understood, the problem is not the name of the entity, meaning it can be two words / one word. The problem is the name of the field the user set as a foreign key.

tupe12334:
@Juicpt @yuval-hazaz @abrl91 Please resolve all the reviews, I will not merge this pr for the 0.12.8 release for now

yuval-hazaz:
@tupe12334 I don't think you meant to close this PR

Issue #2613, opened 2022-04-21
create permissions interceptor to handle the permissions 


abrl91:
### Feature description

We are handling the permissions now in a way that makes the route handlers massive and less readable.
By creating a permissions interceptor, we can handle the permissions stuff outside the function body and manipulate the returned value according to the user's permission right before sending the result to the client.

https://docs.nestjs.com/interceptors

![image](https://user-images.githubusercontent.com/39680385/164460958-2eb77dbe-fed6-4bbf-b3d0-6630b084f136.png)

### Use case

By using a permission interceptor, we would not only achieve cleaner architecture but also could handle public routes in the feature easily.

As a first step, we will implement a filter result interceptor which would be responsible to read and search operations in REST and Query operations in graphql.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

tupe12334:
@abrl91 Many thanks for posting this image, I learned a lot from it ❤️.

PR #2636, opened 2022-04-26
Feat/2613 permission interceptor

abrl91:
Issue Number: #2613 

## PR Details

- from read / query operations on REST and gql delete the parma decorator named: 
'@nestAccessControl.UserRoles() userRoles: string[]'
and the permission handling part from the body of the functions

- add `accessContorlPermissionsInterceptor` to handle this filter operation after the route and before sending the request to the user.

## PR Checklist


mshidlov (approved):

packages/amplication-data-service-generator/src/server/static/src/interceptors/filterResult.interceptor.ts
```diff
@@ -0,0 +1,42 @@
+import {
```
yuval-hazaz, line 1: I suggest that we call this interceptor "FilterResponse" and the other one that we are working on "ValidateRequest"

yuval-hazaz (commented):
@abrl91 please see my suggestion

Issue #2637, opened 2022-04-26
create validate input interceptor to handle the permissions


abrl91:
### Feature description

Following https://github.com/amplication/amplication/issues/2613

We are handling the permissions now in a way that makes the route handlers massive and less readable.
By creating a permissions interceptor, we can handle the permissions stuff outside the function body and manipulate the returned value according to the user's permission right before sending the result to the client.

https://docs.nestjs.com/interceptors

![image](https://user-images.githubusercontent.com/39680385/164460958-2eb77dbe-fed6-4bbf-b3d0-6630b084f136.png)

### Use case

By using a permission interceptor, we would not only achieve cleaner architecture but also could handle public routes in the feature easily.

As a second step, we will implement a validate input interceptor which would be responsible of validate the input of mutations actions such as create, update and delete in REST and Mutation in qraphql.
On the request body of those requests, we have to validate every attribute - if the user with the role that tries for example to update the customer address, has permission not only to mutate the customer but specifically to mutate his address attribute.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #2638, opened 2022-04-26
add validate input interceptor and update controller and resolver

abrl91:
Issue Number: #2637 

## PR Details

- deleted '@nestAccessControl.UserRoles() userRoles: string[]' from mutation operations on REST and gql 
- delete the permission handling part from the body of those functions

add `ValidateInputInterceptor` to handle this filter operation before the route, meaning before doing some mutations operations to prevent mutation of the data from roles that don't have permissions to do such mutation.

## PR Checklist


yuval-hazaz:
for future reference, it is important to mention that we decided to remove the logic of validating the request completely from the endpoints of "to-many" relations since anyway the client does not change the values of any fields of the object, but only creates a link between the objects.

PR #2682, opened 2022-05-04
Feat/2006 public routes 

abrl91:
Issue Number: #2006 

## PR Details

still in process:
what has already been done?
- handle the client-side
- add Public to the `EnumEntityPermissionType` model and Prisma Enum
- check isPublic on default guard in order to determine if the route should be public or not
- add logic for public permission of the entity, on every action - only for the controller (resolver on a different PR).

what is still missing?
- if all the methods in controller/resolver were set to public - remove unused imports such as AclFilterResposeInterceptor and AclValidateRequestInterceptor

 Important To Do:
 - open a pr from release 0.13 and change there the relevant Enum on Prisma schema and models

## PR Checklist


packages/amplication-data-service-generator/src/util/ast.ts
```diff
@@ -769,3 +779,61 @@ export function createGenericArray(
...
+  if (!decorator) {
+    throw new Error(
+      `Could not find class decorator with the name ${decoratorName} in provided AST node`
+    );
+  }
+
+  return decorator;
+}
+
+export function setDecoratorState(
```
yuval-hazaz, line 808: We better use a more specific version of this function instead of this generic version.
I think we need to create a function named "removeIdentifierFromUseInterceptorDecorator" in "nestjs-code-generation.ts" 

Different decorators use different constructors and accept different parameters, so this version will not fit other use cases. 

Also, you are expecting a format of a @common.useInterceptor and any other use case where the decorator will not use a namespace, it will not work - so specific version of the function will resolve this issue

packages/amplication-data-service-generator/src/util/set-endpoint-permission.ts
```diff
@@ -0,0 +1,60 @@
...
+        assert(classMethod);
+
+        if (
+          action === EnumEntityAction.Search ||
+          action === EnumEntityAction.View
+        ) {
+          setDecoratorState(
+            classMethod,
+            USE_INTERCEPTORS_DECORATOR_NAME,
+            "AclFilterResponseInterceptor"
```
yuval-hazaz, line 35: use const or Identifier

packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
```diff
@@ -173,6 +175,52 @@ async function createControllerModule(
...
+        DISCONNECT: builders.identifier(camelCase(`disconnect ${field.name}`)),
+        UPDATE: builders.identifier(camelCase(`update ${field.name}`)),
+      };
+
+      return setToManyEndpointPermission(
+        classDeclaration,
+        toManyFunctionMapping,
+        entity
+      );
+    });
```
yuval-hazaz, line 222: I think this part should be included in `createToManyRelationMethods` - it will save another iteration on all the relation fields and creation of all the identifiers

packages/amplication-data-service-generator/src/util/ast.ts
```diff
@@ -769,3 +779,78 @@ export function createGenericArray(
     builders.tsTypeParameterInstantiation([itemType])
   );
 }
+
+export function removeDecoratorByName(
```
yuval-hazaz, line 783: why do we still need this function?

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
```diff
@@ -36,3 +37,37 @@ export function addInjectableDependency(
     })
   );
 }
+
+export function removeIdentifierFromUseInterceptorDecorator(
+  node: ASTNode,
+  decoratorName: string,
```
yuval-hazaz, line 43: the decorator name should be const in the function and not a parameter

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
```diff
@@ -36,3 +37,37 @@ export function addInjectableDependency(
...
+  argToDelete: string /** @todo: change to string[] */
```
yuval-hazaz, line 44: change to NamedTyped.Identifier

yuval-hazaz (changes requested):
see comments

packages/amplication-data-service-generator/src/util/ast.ts
abrl91: Maybe I got you wrong. Your comment was something like "create a specific version of this function to interceptors", so I thought you mean to leave as a function for decorator without namespace and  create a new specific one for interceptors

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
abrl91: why? UseInterceptors() can get arguments (identifier[]) and you want that the caller of this function sends the argument that needs to be removed

packages/amplication-data-service-generator/src/util/ast.ts
abrl91: removeDecoratorByName is used to remove UseRoles decorator

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
abrl91: when I call this function I send the decorator name as a hardcoded string and in the if statement on the function above we have the following check:
```
if (callee.value && callee.value.property?.name === decoratorName
```
and `name` on callee.value.property is a string too

packages/amplication-data-service-generator/src/server/resource/controller/create-controller.ts
```diff
@@ -247,7 +276,47 @@ async function createToManyRelationMethods(
...
+    EnumEntityAction.Update
+  );
+  toManyMethodsIdsActionPairs.set(
+    toManyMapping["CONNECT"] as namedTypes.Identifier,
+    EnumEntityAction.Update
+  );
+  toManyMethodsIdsActionPairs.set(
+    toManyMapping["DISCONNECT"] as namedTypes.Identifier,
+    EnumEntityAction.Delete
+  );
```
yuval-hazaz, line 310: why are you using Map here and object in the parent use cases? 
if you decide to go with Map, please use Map< EnumEntityAction, namedTypes.Identifier>
if you decide to use an object, you should also consider using a type for the object instead of using `as`

packages/amplication-data-service-generator/src/util/ast.ts
yuval-hazaz, line 783:
```suggestion
// This function removes all instances of a decorator by its name. in case the same decorator exists multiple times, it will be removed from all locations
export function removeDecoratorByName(
```

packages/amplication-data-service-generator/src/util/ast.ts
```diff
@@ -769,3 +779,78 @@ export function createGenericArray(
...
+  });
+
+  if (!decorator) {
+    return false;
+  }
+
+  return true;
+}
+
+export function setDecoratorState(
```
yuval-hazaz, line 824: are we still using this function? if not, please remove it.

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
yuval-hazaz: I am not talking about the identifier, I am talking about the decorator name and a not about the identifier

packages/amplication-data-service-generator/src/util/nestjs-code-generation.ts
yuval-hazaz: in this case, I am talking about the identifier and not about the name of the decorator

packages/amplication-data-service-generator/src/util/set-endpoint-permission.ts
```diff
@@ -0,0 +1,60 @@
...
+const ACL_FILTER_RESPONSE_INTERCEPTOR_NAME = "AclFilterResponseInterceptor";
+const ACL_VALIDATE_REQUEST_INTERCEPTOR_NAME = "AclValidateRequestInterceptor";
+
+export function setEndpointPermissions(
+  classDeclaration: namedTypes.ClassDeclaration,
+  methodId: namedTypes.Identifier,
+  action: EnumEntityAction,
+  entity: Entity
+): void {
+  entity.permissions.find((entityPermission) => {
```
yuval-hazaz, line 22: I dont think this is what you meant to write... 
if you are using entity.permissions.find() then I would expect you just to find the element and exit, and only then use the element you found to continue with the logic

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/util/ast.ts
abrl91: added the comment

packages/amplication-data-service-generator/src/server/resource/resolver/resolver.base.template.ts
```diff
@@ -15,6 +15,8 @@ import { isRecordNotFoundError } from "../../prisma.util";
 // @ts-ignore
 import { MetaQueryPayload } from "../../util/MetaQueryPayload";
 // @ts-ignore
+import { Public } from "../../decorators/public.decorator";
+// @ts-ignore
```
yuval-hazaz, line 19: since we are adding this conditionally from the code, should we remove this?

yuval-hazaz (approved):
@abrl91 great work!!
I approved all changes, but I think we should update the snapshot test to include at least one entity with public endpoints so we can see the impact on the snapshot. 
It will give us more confidence that the results are as expected.

packages/amplication-data-service-generator/src/server/resource/resolver/resolver.base.template.ts
abrl91: @yuval-hazaz the implementation of adding an import to 'Public' decorator is only in the create-controller for now. I decided to wait until you approve it and then merge it and use it in the create-resolver file. I will add the implementation now and remove this line

abrl91:
> @abrl91 great work!! I approved all changes, but I think we should update the snapshot test to include at least one entity with public endpoints so we can see the impact on the snapshot. It will give us more confidence that the results are as expected.

@yuval-hazaz 
OK. I think we also need to check if all the methods in the controller or the resolver are public and if so, remove the imports of AclFilterResponseInreceptor and AclValidateRequestInterceptor.

mshidlov:
@abrl91 lets merge this and not wait for that cleanup. we can work on it separately

yuval-hazaz:
@abrl91 one last thing before merging - I think we should test all actions and not just view and search.

Issue #2687, opened 2022-05-05
create-resolver toMany and toOne: add logic for public


abrl91:
handle to-many and to-one for public queries and mutations

PR #2689, opened 2022-05-05
feat(dsg) add logic for handling public queries and mutation for reso…

abrl91:
Issue Number: #2687 

## PR Details

add logic for handling public queries and mutation for resolvers to-many and to-one relations

## PR Checklist


yuval-hazaz (changes requested):
@abrl91 I see that you added the call to setEndpointPermissions for the relations but you also need to add the call to setEndpointPermissions for the main class (line 160).

Also, I think we should change at least one entity in the test file for the snapshot to use public endpoint sin order to see the final results in the snapshot.

yuval-hazaz (approved):

abrl91:
> setEndpointPermissions

If I understood you right, this call to `setEndpointPermissions` in on a different PR

PR #2699, opened 2022-05-08
Fix - navigating to the correct entity from a related field

### LINKED ISSUE #2693: 🐛 Bug Report: when navigating from entity page to anther the general settings in not selected automatically
### What happened?

from the entities page, when you navigate to an entity, the general setting tab is selected and you can see data,
but when from the entity itself, when you want to navigate to a related entity, the related entity state is not set on the general settings tab.

I expect that the general setting tab would be selected while navigating from one entity to its related entity.

### How to reproduce

see the video

https://user-images.githubusercontent.com/39680385/167179005-7dd97c21-1035-43ee-b51c-aec7529f805a.mov

### Amplication version

0.12.6

### Environment

_No response_

_No response_

sebabratakundu:
Issue Number: #2693 

## PR Details

- User will redirect to general tab of an selected entity while navigating from one entity to its related entity.

## PR Checklist

Closes #2693


packages/amplication-client/src/Entity/RelatedEntityFieldField.tsx
```diff
@@ -37,7 +44,7 @@ const RelatedEntityFieldField = ({ entityDisplayName }: Props) => {
         <EntityRelationFieldsChart
           fixInPlace={false}
           applicationId={data.entity.appId}
-          entityId={data.entity.id}
+          entityId={entity}
```
yuval-hazaz, line 47: I dont think this is the right fix for this issue. 
I expected to see the fix in the component `amplication-client/src/Entity/EntityRelationFieldsChart.tsx`
The component includes 4 links, only the first pair is not working correctly - the second one is fine. I do not expect to see a change of the component props, rather just a change in how the component builds the links internally

yuval-hazaz (changes requested):

packages/amplication-client/src/Entity/RelatedEntityFieldField.tsx
sebabratakundu: Ok, will do the changes

yuval-hazaz (approved):

yuval-hazaz:
@tupe12334 please merge to next release

Issue #2700, opened 2022-05-08
🐛 Bug Report: in to-many relation Public decorator is added to related entity while supposed to be added to the entity 


abrl91:
### What happened?

in create-controller, where we manage the to-many relation, calling to setEntityPermission should be called with the entity itself when the method id is `update/connect/disconnect`

### What you expected to happen

update, connect, and disconnect are methods that mutate the entity itself and not the related entity. Therefore the entity we pass to setEntityPermission should be the parent entity and not the child (related)is

### How to reproduce

-

### Amplication version

0.12.7

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #2701, opened 2022-05-08
Fix/2700 public to many

abrl91:
Issue Number: #2700 

## PR Details

fix adding of the public decorator to a related entity for update connect and disconnect

## PR Checklist


abrl91:
@yuval-hazaz Do you want me to change the Map to the array approach in the root file also?

yuval-hazaz (approved):

Issue #2639, opened 2022-04-26
Create `@amplication/code-store-service` - generated code access API


mshidlov:
create a new service that will give access to the generated code. 
the service should include two APIs:
* **list:** lists the contents of the current app-commit directory.
* **concatenate:** returns the content of a file.

The service is public and can be accessed by the outside world, as such it will include the JWT authorization layer and will check resources access permissions.

the authorization flow should be:
1. verify the JWT access-token
2. send resource access permissions validation message over Kafka topic `authorization.command.can-access-build.0`
3. accept permissions validation reply over Kafka reply topic.

Issue sub-tasks:
- [x]  Create `@amplication/code-store-service` with Amplication
- [x]  Add storage functionality to `@amplication/code-store-service`
- [x]  Create authorization flow in `@amplication/code-store-service`  
- [x]  Add build authorization consumer to `@amplication/server` 

Issue physical design:

LIST API REQUEST:
```
HTTP GET storage/[app-id]/[commit-id]/list?path=packages%2Fserver?per_page=10&page=0
requiered: app-id, commit-id
optional: 
	path - default to root folder
	per_page - default to 20
	page - default to 0
headers:
	Authorized
```
LIST API RESPONSE:
```
content-type: application/json
body: [
  {
     "type": "folder""name": "src"
  },
  {
    "type": "file""name": "package.json"
  }
]
```

CONCATENATE API REQUEST:
```

HTTP GET storage/[app-id]/[commit-id]/concatenate?path=packages%2Fserver%2Fpackage.json
requiered: app-id, commit-id, path
headers:
	Authorized
```
CONCATENATE API RESPONSE:
```
content-type: text/plain
```

![Blank diagram - Page 3](https://user-images.githubusercontent.com/91742238/165378145-0bd60bde-1466-438d-80f8-acaa555303cc.png)

yuval-hazaz:
@mshidlov what is the origin of the word CONCATENATE for the API? is that something common for this type of API? I am not sure if that is a typo or by design?

yuval-hazaz:
@mshidlov are we naming this service storage-api or code-store?

PR #2749, opened 2022-05-17
Feat/amplication storage gateway

tupe12334:
Partial fix for #2639

## PR Details

This pr create a new service called storage gateway
## PR Checklist


packages/amplication-client/.env
```diff
@@ -7,4 +7,5 @@ REACT_APP_FEATURE_FLAGS="{}"
 REACT_APP_AMPLITUDE_API_KEY=aVqxCAcXOdCx5DPI41s6EmwsSCB8Kjz6
 ֿREACT_APP_GITHUB_CLIENT_ID=2e4a744d73ea633fd615
 
-REACT_APP_DATA_SOURCE="/graphql"
\ No newline at end of file
+REACT_APP_DATA_SOURCE="/graphql"
+REACT_APP_STORAGE_GATEWAY_URL="http://localhost:3002"
```
mshidlov, line 11: it shouldn't end with with `/storage`?

mshidlov, line 11: it shouldn't end with with `/storage`?

packages/amplication-server/.env
```diff
@@ -31,6 +31,7 @@ CONTAINER_BUILDER_DEFAULT=docker
 
 # Kafka Topics
 GENERATE_PULL_REQUEST_TOPIC="git.pr.generate.message"
+CHECK_USER_ACCESS_TOPIC="auth.user.access"
```
mshidlov, line 34: Why this is not like the kafka library expects? kafka_topic_[name]

packages/amplication-server/src/core/build/build.controller.ts
```diff
@@ -48,4 +55,17 @@ export class BuildController {
...
+  @MessagePattern(
+    EnvironmentVariables.instance.get(CHECK_USER_ACCESS_TOPIC, true)
+  )
+  async checkUserAccess(
+    @Payload() message: KafkaMessage
+  ): Promise<{ value: ResultMessage<boolean> }> {
+    const validArgs = plainToInstance(CanUserAccessArgs, message.value);
+    const isUserCanAccess = await this.buildService.canUserAccess(validArgs);
+    return {
+      value: { error: null, status: StatusEnum.Success, value: isUserCanAccess }
```
mshidlov, line 68: Do we have a class for that message?

mshidlov, line 68: Do we have a class for that message?

packages/amplication-client/.env
mshidlov: and should you have client changes in this PR? it is not suppose to be here?

mshidlov: and should you have client changes in this PR? it is not suppose to be here?

yuval-hazaz:
@morhag90 @tupe12334 I made some refactoring on the codeViewBar component. 
Please have a look at it

yuval-hazaz:
you can have a look at all these commits:
image

packages/amplication-client/src/Application/code-view/CodeViewExplorer.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const handleSelectBuild = (build: Build) => {
+    setSelectedBuild(build);
+  };
+
+  useEffect(() => {
+    onFileSelected(null);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [selectedBuild]);
+
+  if (!data) {
```
barshimi, line 54: double return on react function

packages/amplication-client/src/Application/code-view/CodeViewExplorer.tsx
```diff
@@ -0,0 +1,96 @@
...
+    onCompleted: async (data) => {
+      setSelectedBuild(data.builds[0]);
+    },
+  });
+
+  const handleSelectBuild = (build: Build) => {
+    setSelectedBuild(build);
+  };
+
+  useEffect(() => {
```
barshimi, line 49: why do you need to set it to null on any change ?
you can change file selected on any select build instead

packages/amplication-client/src/Application/code-view/CodeViewExplorer.tsx
```diff
@@ -0,0 +1,96 @@
...
+import "./CodeViewBar.scss";
+
+const CLASS_NAME = "code-view-bar";
+
+type Props = {
+  app: App;
+  onFileSelected: (selectedFile: FileDetails | null) => void;
+};
+
+export class FileMeta {
```
barshimi, line 17: why class ?

packages/amplication-client/src/Application/code-view/CodeViewExplorer.tsx
```diff
@@ -0,0 +1,96 @@
...
+      orderBy: {
+        [CREATED_AT_FIELD]: SortOrder.Desc,
+      },
+    },
+    onCompleted: async (data) => {
+      setSelectedBuild(data.builds[0]);
+    },
+  });
+
+  const handleSelectBuild = (build: Build) => {
```
barshimi, line 45: this function is redundant

packages/amplication-client/src/Application/code-view/CodeViewExplorerTree.tsx
```diff
@@ -0,0 +1,120 @@
...
+import "./CodeViewBar.scss";
+
+const CLASS_NAME = "code-view-bar";
+
+type Props = {
+  selectedBuild: Build;
+  onFileSelected: (selectedFile: FileDetails) => void;
+};
+
+const NO_FILES_MESSAGE = "There are no available files to show for this build";
```
barshimi, line 21: this is a general comment. I think we should address texts in a different way and not at the end point component. it will give us the ability to handle languages as well as dynamic text change

packages/amplication-client/src/Application/code-view/CodeViewExplorerTree.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const INITIAL_ROOT_NODE: FileMeta = {
+  type: NodeTypeEnum.Folder,
+  name: "root",
+  path: "/",
+  children: [],
+  expanded: false,
+};
+
+const CodeViewExplorerTree = ({ selectedBuild, onFileSelected }: Props) => {
```
barshimi, line 31: React component should be assign her type: const CodeViewExplorerTree: React.FC<Props>

packages/amplication-client/src/Application/code-view/CodeViewExplorerTree.tsx
```diff
@@ -0,0 +1,120 @@
...
+  useEffect(() => {
+    //reset the state when the build changes
+    const newRootFile = { ...INITIAL_ROOT_NODE };
+    setRootFile(newRootFile);
+    setSelectedFolder(newRootFile);
+  }, [selectedBuild]);
+
+  const handleNodeClick = useCallback(
+    async (file: FileMeta) => {
+      switch (file.type) {
```
barshimi, line 62: I would convert switch case to hashMap object. it is efficient in many ways: 
{
  [NodeTypeEnum.File]: () => {
onFileSelected({
            buildId: selectedBuild.id,
            filePath: file.path,
            isFile: true,
            fileName: file.name,
          });
}
}

packages/amplication-client/src/Application/code-view/CodeViewExplorerTree.tsx
```diff
@@ -0,0 +1,120 @@
...
+        <div>
+          {isError && error}
+          {rootFile.children?.length ? (
+            <TreeView
+              expanded={expandedFolders}
+              defaultExpandIcon={<Icon icon="arrow_left" />}
+              defaultCollapseIcon={<Icon icon="arrow_left" />}
+            >
+              {rootFile.children?.map((child) => {
+                return (
```
barshimi, line 102: no need for return
rootFile.children?.map((child) => (
<FileExplorerNode
                    file={child}
                    key={child.path + JSON.stringify(child.children)}
                    onSelect={handleNodeClick}
                  />
)

packages/amplication-client/src/Application/code-view/CodeViewPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export type FileDetails = {
+  buildId: string;
+  filePath: string;
+  isFile: boolean;
+  fileName: string;
+};
+
+function CodeViewPage({ match }: Props) {
+  const applicationId = match.params.application;
```
barshimi, line 32: we can use useParams hook instead to pass it through the component

packages/amplication-client/src/Application/code-view/CodeViewPage.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [fileDetails, setFileDetails] = useState<FileDetails | null>(null);
+  useNavigationTabs(applicationId, NAVIGATION_KEY, match.url, "Code View");
```
barshimi, line 34: I think this mechanism should be decoupled from end point component and should be at the top level of the routing or app. it is not for this PR, but I think it will prevent performance issue in future

packages/amplication-client/src/Application/code-view/FileExplorerNode.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export type FileExplorerNodeProps = {
+  file: FileMeta;
+  onSelect: (data: FileMeta) => void;
+};
+
+export const FileExplorerNode = ({ file, onSelect }: FileExplorerNodeProps) => {
+  const handleNodeClick = useCallback(
+    (id: string, file: FileMeta) => {
+      onSelect && onSelect(file);
```
barshimi, line 14: onselect will always be on component, it a require prop

packages/amplication-client/src/Application/code-view/FileExplorerNode.tsx
```diff
@@ -0,0 +1,52 @@
...
+    },
+    [onSelect]
+  );
+
+  const iconName = useMemo(
```
barshimi, line 19: I would convert it to pure function outside the component.

packages/amplication-client/src/Application/code-view/FileExplorerNode.tsx
```diff
@@ -0,0 +1,52 @@
...
+    () => (file.type === NodeTypeEnum.Folder ? "folder" : "file"),
+    [file.type]
+  );
+
+  const farIcon = useMemo(
```
barshimi, line 24: instead:
if (file.type !== NodeTypeEnum.Folder) return null

return file.expanded ? <Icon icon="chevron_down" /> : <Icon icon="chevron_right" />

packages/amplication-client/package.json
```diff
@@ -17,9 +19,7 @@
     "events": "3.3.0",
     "formik": "2.2.9",
     "graphql": "15.0.0",
-    "lodash.debounce": "4.0.8",
-    "lodash.keyby": "4.6.0",
-    "lodash.omit": "4.5.0",
+    "lodash": "4.17.21",
```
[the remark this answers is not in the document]
yuval-hazaz: this change is not part of this PR. we actually added a separate PR to use a single import in all packages instead of separate imports. 
I think you may be right that on the client, we better use specific imports. 
If you think it is important, I suggest that you open a new issue to be fixed separately

packages/amplication-client/src/Application/ApplicationLayout.tsx
```diff
@@ -227,6 +236,10 @@ function ApplicationLayout({ match }: Props) {
                 path="/:application/github"
                 component={SyncWithGithubPage}
               />
+              <RouteWithAnalytics
+                path="/:application/code-view"
+                component={CodeViewPage}
```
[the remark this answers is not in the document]
yuval-hazaz: Please open a new issue to be handled separately

packages/amplication-client/src/Application/code-view/CodeViewBar.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { App } from "../../models";
+import { FileDetails } from "./CodeViewPage";
+import AppGitStatusPanel from "../git/AppGitStatusPanel";
+import CodeViewExplorer from "./CodeViewExplorer";
+import "./CodeViewBar.scss";
+
+const CLASS_NAME = "code-view-bar";
```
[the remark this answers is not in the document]
yuval-hazaz: Can you please share an example of how you would do that?
you may open a new issue to fix existing implementation or we can start doing it on new components.

packages/amplication-client/src/Application/code-view/CodeViewEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+  fileName: string;
+};
+
+const UNSUPPORTED_EXTENSIONS = ["png", "ico"];
+const UNSUPPORTED_EXTENSION_MESSAGE = "Preview is not available";
+
+const CodeViewEditor = ({ appId, buildId, filePath, fileName }: Props) => {
+  const [content, setContent] = useState<string>("");
+
+  const path = filePath;
```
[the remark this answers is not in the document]
yuval-hazaz: that is a good point - but why do we need to use path and cant use filePath?

packages/amplication-client/src/Application/code-view/CodeViewEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const fileExtension = useMemo(() => {
```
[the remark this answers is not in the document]
yuval-hazaz: ok

packages/amplication-client/src/Application/code-view/CodeViewExplorerTree.tsx
yuval-hazaz: I agree, but it will require a major change. There is already an open issue about localization of text.

yuval-hazaz:
@barshimi @morhag90 
I added some specific comments on the review comments from Shimi where I had additional input.
for all the rest - you can go ahead and implement what Shimi suggested

packages/amplication-client/src/Application/code-view/CodeViewBar.tsx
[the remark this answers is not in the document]
barshimi: const classConverter = (str) => {
  if (!str.length) return ''

  return str.split('').reduce((className, char, idx) => {
    const matchChar = char.match(/[A-Z]/g)
    className += matchChar ? `${idx > 0 ? '-' : ''}${char.toLowerCase()}` : char

    return className
  }, '')
}

console.log(classConverter('HelloWorld'))

packages/amplication-client/src/Application/code-view/StorageBaseAxios.ts
```diff
@@ -0,0 +1,75 @@
...
+  success: true;
+};
+
+export type StorageResponseFile = {
+  name: string;
+  type: NodeTypeEnum;
+  path: string;
+};
+
+export class StorageBaseAxios {
```
barshimi, line 19: I think we can upgrade this wrapper interceptors for retry, cancel, on load, error etc...

packages/amplication-client/src/Application/git/AppGitStatusPanel.tsx
```diff
@@ -0,0 +1,71 @@
+import React from "react";
+import { Button, EnumButtonStyle } from "../../Components/Button";
+import { Icon, Label, Tooltip } from "@amplication/design-system";
+import { Link } from "react-router-dom";
+import * as models from "../../models";
+import { isEmpty } from "lodash";
```
barshimi, line 6: it's better to import the specific lodash function then the library. e.g. import isEmpty from 'lodash.isempty'

packages/amplication-client/src/Components/BuildSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+const CLASS_NAME = "build-selector";
+
+type Props = {
+  builds: Build[];
+  selectedBuild: Build | null;
+  app: App;
+  onSelectBuild: (commit: Build) => void;
+};
+
+const BuildSelector = ({
```
barshimi, line 23: this component should be addressing all the selectors on our app. so it can be more generic and could render any selector. not for this PR

packages/amplication-client/src/Roles/RoleForm.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useMemo } from "react";
 import { Formik } from "formik";
 import { Form } from "../Components/Form";
-import omit from "lodash.omit";
+import { omit } from "lodash";
```
barshimi, line 4: the upper choice is better to my opinion

packages/amplication-client/src/util/formikAutoSave.tsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useCallback } from "react";
 import { useFormikContext } from "formik";
-import debounce from "lodash.debounce";
```
barshimi, line 3: again. could you explain why you prefer the library and not the specific function ?

PR #2799, opened 2022-05-24
added job cleanup after completed

germanilia:
Issue Number: N/A

## PR Checklist

### DIFF
--- a/helm/charts/services/amplication-server/templates/job.yaml
+++ b/helm/charts/services/amplication-server/templates/job.yaml
@@ -3,6 +3,7 @@ kind: Job
 metadata:
   name: "{{ .Values.name }}-migrate"
 spec:
+  ttlSecondsAfterFinished: 5
   template:
     metadata:
       name: "{{ .Values.name }}-migrate"


.github/workflows/cd-aws.yml
```diff
@@ -134,6 +134,13 @@ jobs:
         run: |
           curl -sSL -o /usr/local/bin/argocd https://github.com/argoproj/argo-cd/releases/latest/download/argocd-linux-amd64
           chmod +x /usr/local/bin/argocd
+      - name: delete job-migrate
```
mshidlov, line 137: I think that it is too aggressive, and it shouldn't be part of GH deployment, it should be managed from Argo

helm/charts/services/amplication-server/templates/job.yaml
```diff
@@ -3,6 +3,7 @@ kind: Job
 metadata:
   name: "{{ .Values.name }}-migrate"
 spec:
+  ttlSecondsAfterFinished: 5
```
mshidlov, line 6: This configuration is not solving that issue?

germanilia: it should

.github/workflows/cd-aws.yml
germanilia: I'll remove it

PR #2800, opened 2022-05-25
fix(amplication-server): validate numbers min/max to ensure min is not greater than max, fixes #515

### LINKED ISSUE #515: Client: numaric max/min values are not coordinated
the entitys numaric max/min values are not coordinated
therfore the user can set a min val greater then the max

MichaelSolati:
Issue Number: #515

## PR Details

Validate that `minimumValue` is not greater than, or equal to, `maximumValue`.

## PR Checklist
- [ x] Tests for the changes have been added
- [ x] `npm test` doesn't throw any error


yuval-hazaz:
well done @MichaelSolati 👏

yuval-hazaz (changes requested):
looks good to me
@MichaelSolati I think you should add a test to make it perfect

package-lock.json
```diff
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "amplication",
```
tupe12334, line 7: Wich node and npm versions are you working with?

tupe12334:
@MichaelSolati Pay attention to the branch name we usually use the issues number as this format 
<type>/<issus number>-<title>
fix/515-schema-validation

MichaelSolati:
> @MichaelSolati Pay attention to the branch name we usually use the issues number as this format /-<title> fix/515-schema-validation

Noted for next time

package-lock.json
tupe12334: @MichaelSolati ? I don't see why you added this property.

yuval-hazaz:
@MichaelSolati @tupe12334 what is the status of this PR?

package-lock.json
tupe12334: @MichaelSolati Can you please remove this change? or let me know if I can

arielweinberger:
Since there hasn't been a response in over a month, I am going to close this.

PR #2818, opened 2022-05-28
Refactor/2797 rename app

tupe12334:
fix: #2797 

## PR Details
This pr adding a new SQL migration that renames all the relevant parts from `App` to `Resource`

PLEASE!!! check it twice! its a very important migration


packages/amplication-prisma-db/prisma/migrations/20220524135837_from_app_to_resource/migration.sql
```diff
@@ -0,0 +1,59 @@
...
+ALTER TABLE "App"
+RENAME TO "Resource";
+ALTER TABLE "Resource" RENAME CONSTRAINT  "App_pkey" TO "Resource_pkey";
+ALTER TABLE "Resource" RENAME CONSTRAINT  "App_workspaceId_fkey" TO "Resource_workspaceId_fkey";
+ALTER TABLE "Resource" RENAME CONSTRAINT  "App_projectId_fkey" TO "Resource_projectId_fkey";
+ALTER INDEX "App.workspaceId_name_unique" RENAME TO "Resource.workspaceId_name_unique";
+
+-- GitRepository work
+ALTER TABLE "GitRepository"
+RENAME "appId" TO "resource";
```
yuval-hazaz, line 11: is that correct? it was appId before and now just resource (without id)

yuval-hazaz (commented):
good work @tupe12334.
please see my comment

packages/amplication-prisma-db/prisma/migrations/20220524135837_from_app_to_resource/migration.sql
tupe12334: You are right! thanks

Issue #2885, opened 2022-06-08
Implement Novu as notification infrastructure


abrl91:
### Feature description

Add Novu notification to our dashboard.

- Build service with methods for in-app events that we want (for example sandbox is ready)

Link to the POC: https://github.com/amplication/amplication/pull/2828 where we notify the user when the code is ready

### Are you willing to submit PR?

Yes!

abrl91:
reopened in new tasks

PR #2893, opened 2022-06-09
Feat/2885 implement novu

abrl91:
Issue Number: #2885 

## PR Details

Implement Novu notification in our dashboard

https://user-images.githubusercontent.com/39680385/173054164-bab651f7-8424-4fee-b7c1-0203bfc8d4a6.mov

## PR Checklist


abrl91:
@mshidlov please take a look at the ci-aws tests that failed

packages/amplication-server/src/core/notifications/modules/inApp/inAppNotification.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(
+    @Inject(WINSTON_MODULE_PROVIDER) private readonly logger: Logger
+  ) {}
+
+  async pushNotification(
+    notificationName: string,
+    userId: string,
+    payload: { [key: string]: any }
+  ): Promise<void> {
+    const novu = new Novu(process.env.NOVU_API_KEY);
```
scopsy, line 18: This also can be initiated one-time aswell. But I don't believe it will create an overhead for you having it here :)

packages/amplication-server/src/core/notifications/modules/inApp/inAppNotification.service.ts
```diff
@@ -0,0 +1,38 @@
+import { Injectable, Inject, BadRequestException } from '@nestjs/common';
+import { IInAppNotification } from '../../contracts/inAppNotification.interface';
+import { Novu } from '@novu/node';
+import { WINSTON_MODULE_PROVIDER } from 'nest-winston';
+import { Logger } from 'winston';
+
+@Injectable()
+export class InAppNotificationService implements IInAppNotification {
```
scopsy, line 8: This can also be called a NotificationService since Novu handles the abstraction between multiple channels in case you would want send other channels with our workflow editor. But will def work like this aswell :)

abrl91: @scopsy yes. I thought so, but I wasn't sure. On the one hand, I created a template for email notification and saw that I would need to call the same method with the same parameters. But, on the other hand, I thought that we might need more business logic for SMS and email notifications that the in-app notifications shouldn't care of 🤷🏽‍♀️

And thank you for your review. Much appreciate that 🙂

mshidlov:
Can you create an abstraction of notification service around the physical implementation? This will give you the option to use mocking in tests. This is recommended regardless for separation of concerns and implementation modularity.

packages/amplication-client/src/Components/Notifecation.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const onNotificationClick = (notification: IMessage) => {
+    navigator(notification.cta.data.url as string);
+  };
+
+  return data ? (
+    <NovuProvider
+      subscriberId={data.me.id}
+      applicationIdentifier={
+        process.env.REACT_APP_NOVU_APP_ID_FROM_ADMIN_PANEL!
```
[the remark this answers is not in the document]
barshimi: you can either set a variable with default value or env value or set alternative inside component like: 
applicationIdentifier={process.env.REACT_APP_NOVU_APP_ID_FROM_ADMIN_PANEL || ""}

packages/amplication-client/src/Components/Notifecation.tsx
```diff
@@ -0,0 +1,51 @@
...
+import * as models from "../models";
+
+type TData = {
+  me: models.User;
+};
+
+const THEME_DARK = "dark";
+const THEME_LIGHT = "light";
+
+export function Notifications() {
```
barshimi, line 19: I'm not sure if you're the address, but I would expect an empty stage for this component while there is no data. At the moment it seems like you're return null

barshimi: also, I expect that last comment will be present as notification note at the side of my screen and then will be part of the rest of the notification pul. is this part of the UX ?

packages/amplication-client/src/Components/Notifecation.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const themeContext = useContext(ThemeContext);
+  const isDarkMode = themeContext.theme === THEME_DARK;
+  const { data } = useQuery<TData>(AUTH_QUERY);
+
+  const onNotificationClick = (notification: IMessage) => {
```
barshimi, line 24: wrap it with useCallback hook

packages/amplication-server/.env
```diff
@@ -47,3 +47,8 @@ KAFKA_BROKERS="["localhost:9092"]"
 KAFKA_CLIENT_ID='server-queue-client'
 KAFKA_GROUP_ID = 'main-server-group'
 
+BASE_BUILDS_FOLDER=
+
+# Novu:
+NOVU_API_KEY=
```
barshimi, line 53: is this key public ?

abrl91: we have a development value and production value for it. I am not sure what you mean by 'public', but without this key there is no communication between the server and the client

packages/amplication-client/src/Components/Notifecation.tsx
abrl91: Actually, I copy&paste the react component from Novu's documentation and added the `onNotificationClick`.
https://docs.novu.co/docs/notification-center/react-components

What do you mean by notification note at the side of the screen? Like a toast notification? I don't think that Novu supports such a thing. If we want it, we need to add it on our own.

packages/amplication-server/.env
scopsy: This value is a secret key used to communicate with the notifications API. Better to keep it in your secret management store for the backend.

packages/amplication-client/src/Components/Notifecation.tsx
scopsy: The notification center itself has an empty state for when no notifications exist. (Attaching an image)
CleanShot 2022-06-12 at 11 46 39

Regarding the toastr, there is no toastr functionality available yet, but we can add a hook for you. So when a new notification received, you could create a floating note for it with your design-system. Let me know if that is something needed :)

abrl91:
new task #2903  
new PR #2940

Issue #2903, opened 2022-06-13
Create an in-app notifications service `@amplication/notifications-service`


mshidlov:
We should create a notifications service, `@amplication/notifications-service`. this service will consume kafka topic for in-app notifications and will trigger the in-app notifications.

the service will consume messages from topic: `notification.internal.in-app-push.event.0`
The messages schema on that topic will be:
```
{
  user: string,
  payload: {
    title?: string,
    message: string
  },
  template?: string
}
```
Field | Description | 
--- | --- | 
userId | contains the workspace user-id| 
payload.title | contains the notification title, this field is optional, and in case of not providing the template name, a default empty title will be used.| 
payload.message | contains the notification content.| 
template  | the name of the notification template, this field is optional, and in case of not providing the template name, a default template will be used. | 

![Blank diagram (5)](https://user-images.githubusercontent.com/91742238/173291946-aede5142-c7bf-42dc-95e2-581c83f7b115.png)

- [ ] create service using Amplication
- [ ] Consume kafka topic 
- [ ] implement notifications login
- [ ] create a service docker file
- [ ] include service docker-compose
- [ ] add new service helm chart

abrl91:
@mshidlov What about the client-side implementation? On a different issue?

PR #2940, opened 2022-06-16
Feat/2903 amplication notifications service

abrl91:
Issue Number: #2903 

## PR Details
- Added a new service - amplication/notification-service
- This service was built from amplication
- This service is a microservice listening to a kafka queue (consumer)
- Use Novu as a notification infrastructure

## PR Checklist


yuval-hazaz:
for next time, I think it is better to push a few separate PRs - one for the service, another for the chart and so on

PR #2951, opened 2022-06-20
Implement Amplication Notifications Service

abrl91:
Issue Number: #2903 

## PR Details
- Added a new service - amplication/notification-service
- This service was built from amplication
- This service is a microservice listening to a kafka queue (consumer)
- Use Novu as a notification infrastructure

## PR Checklist


packages/amplication-notifications-service/src/contracts/kafkaMessagePayload.ts
```diff
@@ -0,0 +1,15 @@
+export class KafkaMessagePayload {
```
arielweinberger, line 1: What is the value of making this a class rather than an interface?

packages/amplication-notifications-service/src/contracts/notificationPattern.ts
```diff
@@ -0,0 +1,15 @@
+export class NotificationPattern {
```
arielweinberger, line 1: Likewise here, what is the value of making this a class rather than an interface?

packages/amplication-notifications-service/src/core/notifications/notifications.controller.ts
```diff
@@ -0,0 +1,31 @@
...
+  async onNotificationReceived(
+    @Payload() message: { value: KafkaMessagePayload }
+  ) {
+    const { value } = message;
+    const notification: NotificationPattern = {
+      userId: value.userId,
+      notificationName: value.template,
+      payload: value.payload,
+    };
+    return this.notificationService.pushNotification(notification);
```
arielweinberger, line 29: We might want to implement a Dead Letter Queue here in case notifications fail. At the very least we should try-catch and log the error.

An alert of some sort would also be great, but maybe can be a part of a wider alerting strategy using Logz.io/CloudWatch Alerts/Discord or whatever other tools we have.

packages/amplication-notifications-service/src/health/base/health.controller.base.ts
```diff
@@ -0,0 +1,19 @@
+import { Get, HttpStatus, Res } from '@nestjs/common';
+import { Response } from 'express';
+import { HealthService } from '../health.service';
+
+export class HealthControllerBase {
```
arielweinberger, line 5: I believe we can safely remove these base classes. I don't see any value for this in a health check controller other than adding more files to the app ;)

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  const SECRET_KEY = 'SECRET_KEY';
+  const SECRET_VALUE = 'SECRET_VALUE';
+  const configService = mock<ConfigService>();
+  const secretsManagerServiceBase = new SecretsManagerServiceBase(
+    configService
+  );
+  beforeEach(() => {
+    configService.get.mockClear();
+  });
+  it('should return value from env', async () => {
```
arielweinberger, line 15: I'd rephrase this to:
`should get value from secret manager`

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+  it('should return null for unknown keys', async () => {
+    //ARRANGE
+    configService.get.mockReturnValue(undefined);
+    //ACT
+    const result = await secretsManagerServiceBase.getSecret(SECRET_KEY);
+    //ASSERT
+    expect(result).toBeNull();
+  });
+  it('should throw error if dont get key', () => {
```
arielweinberger, line 31: There's a typo here, it should be:
`should throw an error if key could not be found`

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    //@ts-ignore
+    return expect(secretsManagerServiceBase.getSecret()).rejects.toThrow();
+  });
+  it('should throw an exeption if getting null key', () => {
```
arielweinberger, line 35: I'd try to be consistent. Here we use `exception`, above we use `error`.

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface ISecretsManager {
+  getSecret: (key: string) => Promise<any | null>;
+}
+
+export class SecretsManagerServiceBase implements ISecretsManager {
+  constructor(protected readonly configService: ConfigService) {}
+  async getSecret<T>(key: string): Promise<T | null> {
+    if (!key) {
+      throw new Error("Didn't got the key");
```
arielweinberger, line 11: This also is incorrect grammatically. On top of that, there is not enough actionable information in this error message for us to even understand what went wrong. May I suggest:

```
throw new Error(`Key "${key}" could not be found`);
```

arielweinberger (changes requested):
Ideally, when using Kafka and topics, we want to have a common schema available to both consumer and producer.

Duplicating schemas between consumers and producers will result in a lot of code and tech debt that needs to be maintained.

We should think about a strategy for dealing with this, probably utilizing our monorepo. Anyway, it's out of scope for this one.

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
abrl91: this is an auto-generated file by Amplicaton

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
abrl91: this is an auto-generated file by Amplicaton, but I will fix it (we should fix it also in the generated app)

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.spec.ts
abrl91: this is an auto-generated file by Amplicaton, but I will fix it (we should fix it also in the generated app)

packages/amplication-notifications-service/src/providers/secrets/base/secretsManager.service.base.ts
abrl91: this is an auto-generated file by Amplicaton, but I will fix it (we should fix it also in the generated app)
There is also a grammar issue with "didn't got"😣

abrl91:
@arielweinberger 
I should mention the files/folders that were generated by Amplicaation (and I didn't change them except for removing things that I don't need like Prisma, DB....): 
- SecretManager
- health check
- types.ts

packages/amplication-notifications-service/src/contracts/kafkaMessagePayload.ts
abrl91: @mshidlov would you mind sharing your input about it? I remember we talked about it when I built the `git-pull-service.`

packages/amplication-notifications-service/src/core/notifications/notifications.controller.ts
abrl91: @arielweinberger I wrapped it with try-catch and logged the error but I am not sure how to implement a Dead Letter Queue. I think that if this is something that we want we also need to implement it in the other services that consume Kafka queue

arielweinberger: I think we should be okay to push this one for now and deal with the DLQ later, just to unblock you. Next time we'll make sure it's accounted for in the design. Well done!

packages/amplication-notifications-service/src/contracts/interfaces/messagePattern.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface IMessagePattern {
+  userId: string;
+  payload: { [key: string]: string };
+  template: string;
```
abrl91, line 4: @arielweinberger I remember we talked about some kind of mapping between template and notification name but I am not sure why we need it. For example:
image

we can use 'code-view-is-ready' as the template name as well as for the notification name

arielweinberger: Yes indeed. From a separation of concerns point of view, the publisher should not be concerned about a template ID.

We should use a human-readable name, such as "code-review-is-ready" (I would prefer something like `code_review.ready` so we can cover multiple states in the future.

Then it's up to the Notification Service to handle template names.

Today we use novu, tomorrow we might go back to SendGrid. Specific template IDs/names should be abstracted.

arielweinberger:
Another point of feedback I have is prefixing interfaces with `I`.  For example, `IMessagePattern` am not sure how commonly we do this across our codebase, but at the end of the day we generate code that many will use and built on top of.

I've been using TypeScript for a few years now and never seen this convention. I believe it is common in C# and Java. My suggestion is to remove it. Even in the NestJS official docs they do not suggest using the `I` prefix.

Here are some useful resources:
- [StackOVerflow: "Confused about the interface and class coding guidelines for TypeScript"](https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript/41967120#41967120)
- https://github.com/microsoft/TypeScript-Handbook/issues/121

Thoughts? @EugeneTseitlin @yuval-hazaz @abrl91 @mshidlov @barshimi @morhag90 @tupe12334

EugeneTseitlin:
> Another point of feedback I have is prefixing interfaces with `I`. For example, `IMessagePattern`. I am not sure how commonly we do this across our codebase, but at the end of the day we generate code that many will use and built on top of.
> 
> I've been using TypeScript for a few years now and never seen this convention. I believe it is common in C# and Java. My suggestion is to remove it. Even in the NestJS official docs they do not suggest using the `I` prefix.
> 
> Here are some useful resources:
> 
> * [StackOVerflow: "Confused about the interface and class coding guidelines for TypeScript"](https://stackoverflow.com/questions/31876947/confused-about-the-interface-and-class-coding-guidelines-for-typescript/41967120#41967120)
> * [Prohibition against prefixing interfaces with "I" microsoft/TypeScript-Handbook#121](https://github.com/microsoft/TypeScript-Handbook/issues/121)
> * [RFC: Remove I-prefix from TypeScript interfaces and types microsoft/fluentui#10266](https://github.com/microsoft/fluentui/issues/10266)
> 
> More specifically:
> 
> > "Consistency with the community. TypeScript coding conventions suggest avoiding. Most OSS TypeScript projects seem to follow this guidance and avoid prefixing."
> 
> EDIT: To unblock @abrl91 let's take this conversation to Discord. I don't think this is something critical for this PR but more of a higher level topic.

As a developer who started his career from C# I can tell you for sure that this `I` notation comes from there. Java actually has a bit different notation - `Impl` postfix at the end of class that implements an interface. So we usually have `class Repository: IRepository` in C# and `class RepositoryImpl extends Repository` in Java.
In my opinion both of them come from the situation when abstraction layer is added by default as a point of extension that may be used some day in future, but currently there is only one implementation of interface and developer doesn't want to invent a specific name for implementation :)
It's considered as anti-pattern for same reason as Hungarian notation - developer working with IDE can quickly understand what is the type of declaration, so there is no need for such a name convention.
For me it's kind of native, I worked a lot in such environments and it doesn't bother me, but I see the reasoning behind it.
I suggest to put it as an item in formatting and eslint rules issue that we have already. 
Also, I wouldn't give it a high priority, we have a lot of more important things to do.

packages/amplication-notifications-service/src/core/notifications/notifications.service.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(private configService: ConfigService) {
+    this.novuApiKey = this.configService.get<string>(NOVU_API_KEY_ENV) || '';
+  }
+
+  async pushNotification(
+    notificationData: NotificationMessagePattern
+  ): Promise<void> {
+    const { userId, payload, template } = notificationData;
+
+    const novu = new Novu(this.novuApiKey);
```
arielweinberger, line 22: We can create the novu client in the constructor and store it as a class property. This way we don't need to create a new instance for every notification.

yuval-hazaz:
Closing this PR until we revert back to it. we can keep the branch

PR #2954, opened 2022-06-20
2894 generated code file path

barshimi:
Issue Number: N/A

## PR Checklist


packages/amplication-client/src/Application/appSettings/DirectoriesSettingsForm.tsx
```diff
@@ -0,0 +1,208 @@
...
+  query appSettings($id: String!) {
+    appSettings(where: { id: $id }) {
+      id
+      dbHost
+      dbName
+      dbUser
+      dbPassword
+      dbPort
+      authProvider
+      serverSettings {
```
yuval-hazaz, line 197: import the queries from the one of the other forms instead of declaring it again

packages/amplication-client/src/Application/appSettings/GenerationSettingsForm.scss
```diff
@@ -0,0 +1,39 @@
...
+    h3 {
+      margin: 0;
+      @include title-18();
+    }
+  }
+  &__description {
+    @include regular-14();
+  }
+  &__formWrapper_field {
+    width: 280px;
```
yuval-hazaz, line 17: use param

packages/amplication-client/src/Application/appSettings/GenerationSettingsForm.scss
```diff
@@ -0,0 +1,39 @@
...
+  h2 {
+    font-size: 14px;
+  }
+  &__toggle_wrapper {
+    display: flex;
+    flex-direction: column;
+    align-items: flex-start;
+  }
+  .toggle-field {
+    margin-right: 26px;
```
yuval-hazaz, line 28: use param

packages/amplication-client/src/Application/appSettings/GenerationSettingsForm.tsx
```diff
@@ -0,0 +1,195 @@
...
+`;
+
+const GET_APP_SETTINGS = gql`
+  query appSettings($id: String!) {
+    appSettings(where: { id: $id }) {
+      id
+      dbHost
+      dbName
+      dbUser
+      dbPassword
```
yuval-hazaz, line 181: import queries from another form

yuval-hazaz (changes requested):
Please see my comments about the client. 
Please rebase this PR to remove all the unnecessary files and I will continue the review on the DSG

packages/amplication-client/src/Application/appSettings/DirectoriesSettingsForm.tsx
```diff
@@ -0,0 +1,163 @@
...
+            adminUISettings: {
+              generateAdminUI,
+              adminUIPath: adminUIPath || ""
+            },
+            serverSettings: {
+              generateRestApi,
+              generateGraphQL,
+              serverPath: serverPath || ""
+            },
+          },
```
yuval-hazaz, line 92: I think we can use spread on ...data in a way that is easier to maintain, especially since we know that we are only updating some of the properties. 
Also, I think it is worth checking the server-side, and maybe keeping any existing values on the app settings object unchanged if the input is undefined.

packages/amplication-client/src/Application/appSettings/GenerationSettingsForm.tsx
```diff
@@ -0,0 +1,195 @@
...
+      } = data;
+
+      trackEvent({
+        eventName: "updateAppSettings",
+      });
+      updateAppSettings({
+        variables: {
+          data: {
+            dbHost,
+            dbName,
```
yuval-hazaz, line 75: same as above about partial update

packages/amplication-data-service-generator/src/create-data-service-impl.ts
```diff
@@ -52,7 +52,7 @@ export async function createDataServiceImpl(
 
   const modules = (
     await Promise.all([
-      readStaticModules(STATIC_DIRECTORY, BASE_DIRECTORY),
+      readStaticModules(STATIC_DIRECTORY, BASE_DIRECTORY) || [],
```
yuval-hazaz, line 55: should we use a calculated value?

packages/amplication-data-service-generator/src/server/create-server.ts
```diff
@@ -26,37 +40,56 @@ export async function createServerModules(
   userEntity: Entity,
   logger: winston.Logger
 ): Promise<Module[]> {
+  const directoryManager = dynamicPathCreator(
+    get(appInfo, "settings.serverSettings.serverPath", "")
```
yuval-hazaz, line 44: why not use the object directly with its known type?

packages/amplication-data-service-generator/src/server/create-server.ts
```diff
@@ -26,37 +40,56 @@ export async function createServerModules(
...
+  );
+
+  logger.info(`Server path: ${directoryManager.BASE}`);
```
yuval-hazaz, line 47: IIUC .BASE is not the server path

packages/amplication-data-service-generator/src/server/create-server.ts
```diff
@@ -13,11 +14,24 @@ import { createPrismaSchemaModule } from "./prisma/create-prisma-schema-module";
...
 import { createSeedModule } from "./seed/create-seed";
-import { BASE_DIRECTORY, SRC_DIRECTORY } from "./constants";
+import { BASE_DIRECTORY } from "./constants";
 import { createAuthModules } from "./auth/createAuth";
 
 const STATIC_DIRECTORY = path.resolve(__dirname, "static");
 
+const validatePath = (serverPath: string) => serverPath.trim() || null;
+
+const dynamicPathCreator = (serverPath: string) => {
```
yuval-hazaz, line 24: It could be much better if DynamicPathCreator would return a fixed type, and instead of sending the individual values for each function, we could pass the entire typed object and each function could use what ever it needs.

packages/amplication-data-service-generator/src/server/resource/create-resource.ts
```diff
@@ -44,54 +46,66 @@ async function createResourceModules(
...
+    (appInfo.settings.serverSettings.generateGraphQL &&
+      (await createResolverModules(
+        entityName,
+        entityType,
+        serviceModule.path,
+        entity,
+        dtos,
+        srcDirectory
+      ))) ||
+    [];
```
yuval-hazaz, line 81: I don't like the fact that we are working on two separate major changes in a single PR. This would make it much harder to track the changes in the future

packages/amplication-data-service-generator/src/util/set-endpoint-permission.ts
```diff
@@ -50,7 +50,7 @@ export function setEndpointPermissions(
       );
     }
 
-    removeDecoratorByName(classMethod, USE_ROLES_DECORATOR_NAME);
+    //removeDecoratorByName(classMethod, USE_ROLES_DECORATOR_NAME);
```
yuval-hazaz, line 53: please explain

packages/amplication-design-system/src/style/css-variables.scss
```diff
@@ -69,13 +69,13 @@
   --double-spacing: 24px;
   --triple-spacing: 36px;
   --form-field-spacing: 16px;
-  --form-field-long: 400px;
+  --form-field-long: 320px;
```
yuval-hazaz, line 72: Did you check the impact of this change?

packages/amplication-server/src/core/appSettings/appSettings.service.ts
```diff
@@ -4,9 +4,9 @@ import { FindOneArgs } from 'src/dto';
 import { BlockService } from '../block/block.service';
 import { EnumBlockType } from 'src/enums/EnumBlockType';
 import { DEFAULT_APP_SETTINGS, AppSettingsValues } from './constants';
-import { isEmpty } from 'lodash';
 import { User } from 'src/models';
 import { EnumAuthProviderType } from './dto/EnumAuthenticationProviderType';
+// import { compare } from 'bcrypt';
```
yuval-hazaz, line 9: please remove if not used

packages/amplication-server/src/core/appSettings/appSettings.service.ts
```diff
@@ -23,7 +23,10 @@ export class AppSettingsService {
       dbPassword,
       dbPort,
       dbUser,
-      authProvider
+      authProvider,
+      serverSettings,
+      adminUISettings,
+      generateRootFiles
```
yuval-hazaz, line 29: please remove this property

packages/amplication-server/src/core/appSettings/constants.ts
```diff
@@ -30,5 +30,15 @@ export const DEFAULT_APP_SETTINGS: AppSettingsValuesExtended = {
...
+  serverSettings: {
+    generateGraphQL: true,
+    generateRestApi: true,
+    serverPath: ''
+  },
+  adminUISettings: {
+    generateAdminUI: true,
+    adminUIPath: ''
+  },
+  generateRootFiles: true
```
yuval-hazaz, line 43: please remove this property

packages/amplication-server/src/core/appSettings/dto/AdminUISettings.ts
```diff
@@ -0,0 +1,14 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+@ObjectType()
```
yuval-hazaz, line 3: I understand that the original spec included different groups for the server and the admin - but I am not sure if that is the correct structure for these parameters. 
maybe we should stick with the new structure we are using in the UI?
what do you think?

packages/amplication-data-service-generator/src/admin/create-admin.ts
```diff
@@ -26,6 +27,19 @@ import { createDotEnvModule } from "./create-dotenv";
 const STATIC_MODULES_PATH = path.join(__dirname, "static");
 const API_PATHNAME = "/api";
 
+const validatePath = (adminPath: string) => adminPath.trim() || null;
+const dynamicPathCreator = (adminPath: string) => {
```
yuval-hazaz, line 31: see my other comment about types - and maybe even one type to hold all the parameters both for the server and admin

yuval-hazaz (changes requested):
please see my comments. 

in general - I think we should not work on multiple packages in a single PR. 
in the future, it would make it much easier for maintenance and review to work on the single PR for the design-system ,then another one for the client, once for the server, and then the final one for the DSG. 

also, and even more critical - I dont think we should work on multiple tasks in a single PR. this PR includes the changes both for the modular generation, and the code path.

yuval-hazaz:
@barshimi please see my comments
@mshidlov @arielweinberger please see my comment above about the separation of PRs. 
I would love to hear your thoughts, and once agreed please share with the entire team .

arielweinberger:
It would be hard to review without reading the spec for this issue. Is it linked to some ticket on GitHub? @yuval-hazaz

packages/amplication-server/src/core/appSettings/dto/AdminUISettings.ts
arielweinberger: Unrelated to Yuval's comment and probably a wider change. In NestJS' best practices, DTO file names follow this format:
```
admin-ui-settings.dto.ts
```

And the actual DTO class will be called:
```
AdminUISettingsDto
```

I think it's important that we stick to best practices, especially if we generate code for the community to follow.

You can read more here:
https://docs.nestjs.com/techniques/validation

packages/amplication-data-service-generator/src/tests/create-data-service-without-restApi.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import { AppInfo } from "../types";
+import { EnumAuthProviderType } from "../models";
+
+const MODULE_EXTENSIONS_TO_SNAPSHOT = [
+  ".ts",
+  ".tsx",
+  ".prisma",
+  ".env",
+  ".yml",
+];
```
yuval-hazaz, line 13: I think this should be shared

packages/amplication-data-service-generator/src/tests/create-data-service-without-restApi.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+    serverSettings: {
+      generateGraphQL: true,
+      generateRestApi: false,
+      serverPath: "",
+    },
+    adminUISettings: {
+      generateAdminUI: true,
+      adminUIPath: "",
+    },
+  },
```
yuval-hazaz, line 37: maybe this can also be shared, and only override what you need ?

yuval-hazaz (changes requested):
see comments

Issue #3062, opened 2022-07-10
🐛 Bug Report: Type column misnamed in resource entity


tupe12334:
### What happened?

The type column that represents the type of the resource is misnamed, its needs to be called ```resourceType``` and not ```type``` the convention came from the blockType in the block entity

morhag90:
@tupe12334  Can we close this issue?

PR #3064, opened 2022-07-10
refactor(server & client): rename resourceType

tupe12334:
closes: #3062 

## PR Details

This pr refactor the name of the type column

## PR Checklist

### DIFF
--- a/packages/amplication-prisma-db/prisma/migrations/20220707120437_from_app_to_resource/migration.sql
+++ b/packages/amplication-prisma-db/prisma/migrations/20220707120437_from_app_to_resource/migration.sql
@@ -63,6 +63,6 @@ ALTER INDEX "EntityPermissionRole.entityVersionId_action_appRoleId_unique" RENAM
 CREATE TYPE "EnumResourceType" AS ENUM ('Service');
 
 -- Add value to current resources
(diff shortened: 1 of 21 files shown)


abrl91 (approved):

packages/amplication-prisma-db/prisma/migrations/20220707120437_from_app_to_resource/migration.sql
```diff
@@ -63,6 +63,6 @@ ALTER INDEX "EntityPermissionRole.entityVersionId_action_appRoleId_unique" RENAM
 CREATE TYPE "EnumResourceType" AS ENUM ('Service');
 
 -- Add value to current resources
-ALTER TABLE "Resource" ADD COLUMN     "type" "EnumResourceType" NULL;
-UPDATE "Resource" SET type = 'Service';
-ALTER TABLE "Resource" ALTER COLUMN "type" SET NOT NULL;
```
yuval-hazaz, line 68: @tupe12334 we should not alter migrations manually unless there is a real custom changes we need to implement and there is no other way.
We should never alter migrations after they are committed - by doing it you are breaking the environment of anyone who already pulled the change. 

Is there a reason why you could not create a new migration? 

@arielweinberger this should be included in our best practices and be circled around to everyone on the team

Issue #3077, opened 2022-07-11
🚀 Disconnect Resource from Workspace


abrl91:
### Feature description

We added a layer between Workspace and Resource: Project 
The Project holds the connection to Workspace, therefore we can disconnect the connection of Resource to Project.

- [x] remove workspace and workspaceId from the Resource model
- [x] remove resource[] from Workspace model
- [x] remove to auto-creation of a project while creating a new resource
- [x] remove the auto-deletion of a project while deleting a resource
- [x] remove the auto-update of a project while updating a resource
- [x] change resource resolver+service where needed
- [x] change workspace resolver+service where needed
- [x] update permission service
- [x] `/amplication-server/src/core/resource/dto/ResourceCreateInput.ts` change to project connect instead of workspace connect
- [x] remove the `?` from projects on the Resource model

PR #3085, opened 2022-07-11
Feat/3077 disconnect resource from workspace

abrl91:
Issue Number: #3077 

## PR Details

disconnect Resource from Workspace on Prisma schema and fix the places that are related to this change

## PR Checklist


packages/amplication-server/src/core/resource/resource.resolver.ts
```diff
@@ -197,8 +194,8 @@ export class ResourceResolver {
     return await this.resourceService.gitRepository(resource.id);
   }
 
-  @ResolveField(() => Workspace)
-  async workspace(@Parent() resource: Resource): Promise<Workspace> {
-    return this.resourceService.workspace(resource.id);
+  @ResolveField(() => Project)
+  async workspace(@Parent() resource: Resource): Promise<Project> {
```
yuval-hazaz, line 198:
```suggestion
  async project(@Parent() resource: Resource): Promise<Project> {
```

packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -29,13 +29,15 @@ import { AuthorizableResourceParameter } from 'src/enums/AuthorizableResourcePar
 import { AuthorizeContext } from 'src/decorators/authorizeContext.decorator';
 import { GitOrganization } from 'src/models/GitOrganization';
 import { Subscription } from '../subscription/dto/Subscription';
+import { ProjectService } from '../project/project.service';
 @Resolver(() => Workspace)
```
yuval-hazaz, line 33:
```suggestion

@Resolver(() => Workspace)
```

packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -29,13 +29,15 @@ import { AuthorizableResourceParameter } from 'src/enums/AuthorizableResourcePar
...
 @UseFilters(GqlResolverExceptionsFilter)
 @UseGuards(GqlAuthGuard)
 export class WorkspaceResolver {
   constructor(
     private readonly workspaceService: WorkspaceService,
-    private readonly resourceService: ResourceService
+    private readonly resourceService: ResourceService,
```
yuval-hazaz, line 39:
```suggestion
```

yuval-hazaz (changes requested):

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -428,8 +338,8 @@ export class ResourceService {
       where: {
         id: resourceId,
         deletedAt: null,
-        workspace: {
-          users: {
+        project: {
+          resources: {
             some: {
               id: user.id
```
yuval-hazaz, line 344: is that correct? Why are we using user.id to filter resources?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -461,8 +371,8 @@ export class ResourceService {
       where: {
         id: resourceId,
         deletedAt: null,
-        workspace: {
-          users: {
+        project: {
+          resources: {
             some: {
```
yuval-hazaz, line 376: is that correct? Why are we using user.id to filter resources?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -578,8 +488,8 @@ export class ResourceService {
...
```
yuval-hazaz, line 493: is that correct? Why are we using user.id to filter resources?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -659,9 +569,9 @@ export class ResourceService {
...
   }
 
-  async workspace(resourceId: string): Promise<Workspace> {
-    return await this.prisma.resource
-      .findUnique({ where: { id: resourceId } })
+  async project(projectId: string): Promise<Project> {
+    return this.prisma.project
+      .findUnique({ where: { id: projectId } })
       .workspace();
   }
```
yuval-hazaz, line 576: This is not the expected behavior...

Previously it was returning a workspace by a resource ID - I would expect it to return a project by resource ID... instead if return a workspace by project ID

packages/amplication-server/src/core/resource/sampleResource.ts
```diff
@@ -16,7 +16,8 @@ import { ResourceCreateInput } from './dto';
 export const SAMPLE_SERVICE_DATA: ResourceCreateInput = {
   description: 'Sample Service for task management',
   name: 'My sample service',
-  resourceType: EnumResourceType.Service
+  resourceType: EnumResourceType.Service,
+  project: { connect: { id: 'exampleProjectId' } }
```
yuval-hazaz, line 20: this will not work... 
I am not sure if we are using this const at all. it we are please test and fix - if we are not, please remove it

packages/amplication-server/src/models/Resource.ts
```diff
@@ -29,9 +29,9 @@ export class Resource {
   })
   updatedAt!: Date;
 
-  workspace?: Workspace;
+  project?: Project;
 
-  workspaceId?: string;
+  projectId?: string;
```
yuval-hazaz, line 34: why did you create these properties without @Field decorator?

yuval-hazaz (changes requested):

packages/amplication-server/src/models/Resource.ts
abrl91: I forgot to add. Previously it didn't use fields because it took the values from the context, now the user is providing it. Right?

packages/amplication-server/src/core/resource/sampleResource.ts
abrl91: @yuval-hazaz we are using only in `resource.service.spec.ts`

packages/amplication-server/src/core/resource/resource.service.ts
abrl91: @yuval-hazaz I fixed it.
This method is used in `resource.resolver.ts`

packages/amplication-server/src/models/Resource.ts
```diff
@@ -29,9 +29,10 @@ export class Resource {
   })
   updatedAt!: Date;
 
-  workspace?: Workspace;
+  project?: Project;
```
yuval-hazaz, line 32: please add @Field

yuval-hazaz (approved):
@abrl91 I approve the PR so we can move forward. Please see my comment and add the missing decorator

packages/amplication-server/src/models/Resource.ts
abrl91: done

abrl91:
@yuval-hazaz just be aware that we can't merge this branch yet

yuval-hazaz:
> @yuval-hazaz just be aware that we can't merge this branch yet

what is the blocker?

abrl91:
> > @yuval-hazaz just be aware that we can't merge this branch yet
> 
> what is the blocker?

We still don't have the UI that supports the new relation and behavior. 
There are some issues right now:
- clicking on add new resource button is not working 
- when you are on the resource list page (list of your apps) you get a forbidden resource error

I guess this happens because it searches for a projectId to create a new resource and view resources.
We added a check for user's permissions: view/create a resource inside a project with the provided projectId.

Correct me if I'm wrong.

yuval-hazaz:
> There are some issues right now:

I recommend merging this PR into a base feature branch that will allow another developer to complete the missing parts on the client. 

Don't keep this PR open. Let's merge it and move on to the next task

abrl91:
> > There are some issues right now:
> 
> I recommend merging this PR into a base feature branch that will allow another developer to complete the missing parts on the client.
> 
> Don't keep this PR open. Let's merge it and move on to the next task

Merge it into 0.15.0?

PR #3100, opened 2022-07-13
FIX: Deploying none existing image tag

shaharblanksela:
Add a new job before the deploying process to check if the branch name exists in the repo.

### DIFF
--- a/.github/workflows/scripts/deploy-valid-image.sh
+++ b/.github/workflows/scripts/deploy-valid-image.sh
@@ -20,6 +20,5 @@ do
     else
         echo "tag already exist in repository - $REPO_NAME"
         echo "::set-output name=image_exist::true"
-
     fi
 done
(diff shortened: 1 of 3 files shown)


.github/workflows/scripts/deploy-valid-image.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "cleaning up $dir"
+    SERVICE_NAME="$(basename $dir)"
+    REPO_NAME="$SERVICE_NAME"
+    echo "SERVICE_NAME: $SERVICE_NAME"
+    echo "REPO_NAME: $REPO_NAME"
+    set +e
+    cmd="$(aws ecr describe-images --repository-name=$REPO_NAME --image-ids=imageTag=$IMAGE_TAG_ANCHOR)"
+    if [ -z "$cmd" ]
+    then
+        echo "This branch name doesn't exist in repository - $REPO_NAME"
```
arielweinberger, line 18: `in repository` -> `in the repository`

And I think the file name can be `deploy-image.sh`. We do error handling and validations inside anyway, so I wouldn't implicitly mention that in the file name. Let's assume that all of our scripts assume happy path.

arielweinberger (changes requested):

Issue #3057, opened 2022-07-10
new project sreen


abrl91:
### Feature description

create the new project sidebar: create a project, view project, and view projects

related issues: #2782 and #3077

- [x] #3111
- [x] #3112
- [x] #3117
- [x] #3118
- [x] #3122

### Are you willing to submit PR?

Yes I am willing to submit a PR!

muhsinkamil:
Hey @abrl91, Anything left to pick up on this ? ( Sorry, little late, I guess ? )

abrl91:
> Hey @abrl91, Anything left to pick up on this ? ( Sorry, little late, I guess ? )

No, everything is done here. Sorry

abrl91:
All issues have been resolved

PR #3114, opened 2022-07-14
Feat/3111 project list

abrl91:
Issue Number: #3111 

## You see 11 files because this branch includes the changes from #3057 where I started to build the whole parent component with its children. So I checked out from 3057 to another branch, deleted non-relevant files, and left only with projectList and projectListItem components

image

image

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/appRoutes.tsx
+++ b/packages/amplication-client/src/appRoutes.tsx
@@ -270,7 +270,7 @@ export const OldRoutes: RouteDef[] = [
   },
   {
     path: "/project",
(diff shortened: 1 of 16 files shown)


packages/amplication-client/src/appRoutes.tsx
```diff
@@ -268,6 +268,14 @@ export const OldRoutes: RouteDef[] = [
     moduleName: "WorkspaceLayout",
     exactPath: false,
   },
+  {
+    path: "/project",
```
abrl91, line 272: This is just for me. To see what I am doing ;)

PR #3131, opened 2022-07-17
Change Kafka broker IDs to match the new MSK Cluster

shaharblanksela:
The msk-cluster changed, therefor the brokers changed as well and need to be changed.
For staging, it's a fix to correct OAuth.


arielweinberger:
Please add a description. Why should the broker IDs change and what is the outcome?

shaharblanksela:
> 

The production deployed different msk - cluster there for the brokers changed as well.

arielweinberger:
Please include this information in the description and title of the issue.

arielweinberger:
Thanks, I've updated the title to `Change broker IDs to match the new MSK Cluster`

Issue #3166, opened 2022-07-19
design system: circle badge


barshimi:
### Feature description

build circle badge according to design system.
https://www.figma.com/file/9vQeA4a87RBsyCViJTh13W/Design-System?node-id=8%3A443

PR #3179, opened 2022-07-19
Implement `CirlceBadge` component

barshimi:
Issue Number: #3166

## PR Details
Adjust design system circle badge by design
Add new implementation of circle badge - ResourceCircleBadge

## PR Checklist


arielweinberger:
Seems like a lot of commits and files changed. Is this intentional?

arielweinberger (approved):

packages/amplication-client/src/Components/ResourceCircleBadge.tsx
```diff
@@ -0,0 +1,34 @@
+import { CircleBadge, Icon } from "@amplication/design-system";
+import React from "react";
+
+interface Props {
+  type: "settings" | "service" | "api" | "queue";
+  size?: "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge";
```
yuval-hazaz, line 6: I think we should export the same type used in the circleBadge

packages/amplication-design-system/src/components/CircleBadge/CircleBadge.scss
```diff
@@ -18,3 +15,62 @@
   font-family: var(--primary-font-family);
   background-color: var(--secondary-turquoise);
 }
+
+.xxsmall {
+  width: 16px;
```
yuval-hazaz, line 20: please use params for all the values in the begining of this file

yuval-hazaz (approved):
Approving, with minor fixes. please fix and merge

packages/amplication-design-system/src/components/CircleBadge/CircleBadge.scss
barshimi: they are not existed. I can create them' but it's specific to circle

yuval-hazaz: because they are specific, I asked to create SCSS param at the beginning of the file 
see this for example:
packages/amplication-client/src/Application/ApplicationHome.scss

packages/amplication-client/src/Components/ResourceCircleBadge.tsx
yuval-hazaz: exactly, this is why we should not declare it twice. instead just export the inner type and use it here

barshimi:
duplicate PR

Issue #2578, opened 2022-04-14
🐛 Bug Report: one-to-one relation is generated incorrectly


yangguofu2022:
### What happened?

When the key of the associated entity of the user table is not required, but the generated database table structure is not displayed as null.

### What you expected to happen

Consistent type

### How to reproduce

<img width="1093" alt="image" src="https://user-images.githubusercontent.com/101538735/163352079-64e9d285-fc8c-432f-8d95-0901c43f20c3.png">
<img width="759" alt="image" src="https://user-images.githubusercontent.com/101538735/163352174-8c122ffe-6c56-41c2-b56a-aff80a6a731a.png">
<img width="1676" alt="image" src="https://user-images.githubusercontent.com/101538735/163352334-e26ddfa3-fe82-4837-9864-4c71f53e72a7.png">
<img width="747" alt="image" src="https://user-images.githubusercontent.com/101538735/163352394-7afdf6e3-b55c-47fc-855c-fa4473c1679f.png">

**### file:https://ufile.io/c7c7j0av**

### Amplication version

0.12.3

yangguofu2022:
In a many-to-one relationship, this problem can arise

yuval-hazaz:
Thanks @yangguofu2022 

@abrl91, would you please look into this and add your comments/ suggestions ?

abrl91:
@yuval-hazaz @yangguofu2022 
sure, I will take a look at that in the evening

abrl91:
@yangguofu2022 
Hi, I am still looking at it.

For now, you can add a question mark on the settings filed on `prisma.schema` file

![Screen Shot 2022-04-19 at 22 02 56](https://user-images.githubusercontent.com/39680385/164076948-d1f14871-6ed7-454f-b04c-17112f44fea2.png)

remove the docker container and run again
```
npm run docker:db && npm run prisma:generate && npm run db:init && npm run start:debug
```

abrl91:
@yuval-hazaz 
From Prisma documentation:
https://www.prisma.io/docs/concepts/components/prisma-schema/relations/one-to-one-relations#required-and-optional-1-1-relation-fields

our generated code:
![Screen Shot 2022-04-19 at 22 31 25](https://user-images.githubusercontent.com/39680385/164082163-07b91cbc-5e46-4d91-96fa-e09a0541747a.png)

"_In a one-to-one relation, the side of the relation without a relation scalar (the field representing the foreign key in the database) must be optional_"

[...]

PR #3180, opened 2022-07-19
Fix Prisma one to one relation

abrl91:
Issue Number: #2578 

## PR Details
When setting 1:1 relation and running the generated app, prisma schema was generated in a wrong way.

According to Prisma: _"In a one-to-one relation, the side of the relation without a relation scalar (the field representing the foreign key in the database) must be optional"_

The code was generated without the "?" meaning: required.

Another issue: the side of the relation **with** the foreign key must be `@unique`
image

The profile relation field of the User model references the id field of the Profile model.  In this case, you need to mark the field with the @unique attribute, to guarantee that there is only a single Profile connected to each User.

Screen Shot 2022-07-20 at 0 17 07

## PR Checklist


yuval-hazaz:
@abrl91 csn you please explain this :” Another issue: the side of the relation with the foreign key must be @unique”?

abrl91:
> @abrl91 csn you please explain this :” Another issue: the side of the relation with the foreign key must be @unique”?

done

arielweinberger (approved):

Issue #3222, opened 2022-07-25
⭐️ Adding `.amplicationignore`


tupe12334:
There is a problem today with making a new Amplication pr without overwriting the existing work of users.
We want to add .ampignore file in the root folder, and block overwrite in those files

## Implementation
- We will use a `.amplicationignore` file to define the files to ignore.
- We will use the roto `.amplication/*` folder to save the ignored files on codegen - in this case `.amplication/ignored/[commit-id]`

yuval-hazaz:
This fix is a temporary and undocumented fix to allow advanced users more manual control of the code sync with GitHub. 
The .ampignore file should work like .gitignore, and .dockerignore files where each line describes a separate rule using GLOB pattern matching. 
We should not support at this stage the option for a negative rule (to exclude files from an ignored folder, for example)


The implementation of the .ampignore file has a side effect that will prevent the developers from knowing if the ignored file was updated or not in the commit. 
To work around it, I suggest that we will push all the ignored files to a new folder under .ampignored/[commit-id]/ 
This will allow the developer to manually compare the ignored files and take any changes from them if they want to. 

@arielweinberger @mshidlov @tupe12334 @abrl91 please add your thoughts and comments

arielweinberger:
Sounds good. Two questions:
1. Not sure about naming. `.ampignore` is shorter and neater, but `.amplicationignore` is more accurate.

2. Did we check if there is an open-source package that handles glob patterns gitignore-style to handle this for us?

tupe12334:
After discussion with @yuval-hazaz We decided that the expected behavior is that the file will be generated even if its on .amplicationIgnore if it does not exist but will not be on the pr if it already exists.
So the .amplicationignore file is ignored for overwrite

arielweinberger:
With regards to the ignored folders with commit ID reference, let's take advantage of the plan in #2997 

Let's create an `.amplication` folder, and under that, use an `ignored` folder. Using a common `.amplication` folder for all of our "Amplication managed" code stuff will be scalable and won't pollute the project filesystem.

PR #3233, opened 2022-07-26
feat(git-service): add logic

tupe12334:
closes: #3222 

## PR Details

This pr check for .amplicationingnore file and not overwrite files that are there

## PR Checklist


packages/amplication-git-service/src/utils/constants.ts
```diff
@@ -7,3 +7,5 @@ export const GIT_REPOSITORY_EXIST =
 export const INVALID_GIT_REPOSITORY_ID = 'Git Repository does not exist';
 export const UNSUPPORTED_GIT_ORGANIZATION_TYPE =
   'Creation of repositories in a personal account is not supported';
+
```
arielweinberger, line 10: Let's remove this line break, it's inconsistent with the rest of the file.

packages/amplication-git-service/src/utils/isInAmplicationIgnore.ts
```diff
@@ -0,0 +1,18 @@
+import minimatch from 'minimatch';
+
+export function isInAmplicationIgnore(
+  amplicationIgnoreStatements: string[],
+  filePath: string
+): boolean {
+  if (amplicationIgnoreStatements.length === 0) {
```
arielweinberger, line 7: If the amplication ignore statements can be of length 0, then we should not call the `isInAmplicationIgnore()` function to begin with. Let's move this check to `github.service.ts`.

packages/amplication-git-service/src/utils/getAmplicationIgnoreStatements.ts
```diff
@@ -0,0 +1,27 @@
...
+    const amplicationignore = await gitClient.getFile(
+      userName,
+      repoName,
+      AMPLICATION_IGNORE_FILE_NAME,
+      baseBranchName,
+      installationId
+    );
+    amplicationIgnoreStatements = amplicationignore.content.split('\n') || [];
+  } catch (error) {
+    console.log("Repository don't have an .amplicationignore file");
```
arielweinberger, line 23: `Repository don't have an .amplicationignore file` -> `Repository does not have a .amplicationignore file`

packages/amplication-git-service/src/utils/isInAmplicationIgnore.ts
```diff
@@ -0,0 +1,18 @@
+import minimatch from 'minimatch';
+
+export function isInAmplicationIgnore(
+  amplicationIgnoreStatements: string[],
```
arielweinberger, line 4: In glob language, individual lines are called `Expressions` and not `Statements`. Let's rename this to be in line with it please.

https://mincong.io/2019/04/16/glob-expression-understanding/#:~:text=In%20computer%20programming%2C%20glob%20patterns,function%2C%20glob()%20later%20on.

arielweinberger (changes requested):

packages/amplication-data-service-generator/src/version.ts
```diff
@@ -1 +1 @@
-export const version = "0.14.5";
+export const version = "0.14.6";
```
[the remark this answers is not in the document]
tupe12334: I don't think so, i told you that @yuval-hazaz still want that

packages/amplication-git-service/src/utils/isInAmplicationIgnore.ts
tupe12334: I agree and disagree...
I think you are right that the check needs to be the `github.service.ts` but I think the function still needs to be responsible of herself, what do you think?

packages/amplication-git-service/src/utils/isInAmplicationIgnore.ts
tupe12334: You are right, I'm used to the AST statements vs expressions...

packages/amplication-git-service/src/providers/github.service.ts
```diff
@@ -185,8 +197,20 @@ export class GithubService implements IGitClient {
...
+      amplicationIgnoreExpressions.length > 0;
+
     const files = Object.fromEntries(
       modules.map(module => {
+        if (
+          isAmplicationIgnoreHaveExpressions &&
+          isInAmplicationIgnore(amplicationIgnoreExpressions, module.path)
+        ) {
+          return [
+            join('.amplication', 'ignored', amplicationBuildId, module.path),
```
yuval-hazaz, line 210: please use consts

packages/amplication-git-service/src/providers/github.service.ts
```diff
@@ -185,8 +197,20 @@ export class GithubService implements IGitClient {
 
     const authFolder = 'server/src/auth';
 
+    const isAmplicationIgnoreHaveExpressions =
+      amplicationIgnoreExpressions.length > 0;
+
     const files = Object.fromEntries(
       modules.map(module => {
+        if (
+          isAmplicationIgnoreHaveExpressions &&
```
yuval-hazaz, line 206: this is also validated in the function, do we need it twice?

yuval-hazaz (changes requested):

packages/amplication-git-service/src/Dto/entities/GithubFile.ts
```diff
@@ -12,7 +9,7 @@ export class GithubFile {
   path: string | null;
 
   @Field(() => String)
-  content: string;
+  content!: string;
```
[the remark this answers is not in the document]
arielweinberger: I was asking myself the same, and forgot to include it in the review. Thanks!

packages/amplication-git-service/src/utils/isInAmplicationIgnore.ts
arielweinberger: The `isInAmplicationIgnore` function accepts the "statements", which means the invoker already knows the statements. That means the invoker can make sure the statements have `length > 0`, so we remove the condition from this function and keep it simpler.

This function should already assume there is an `.amplicationignore` file and that it's not empty.

packages/amplication-git-service/src/Dto/entities/GithubFile.ts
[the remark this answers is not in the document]
tupe12334: Yes, we know that we get the content of the file as a string if the file exists if not, it will throw.
Its let us write `amplicationignore.content.split('\n') || [];` instead of `amplicationignore?.content?.split('\n') || [];`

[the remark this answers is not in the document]
arielweinberger: Not all files have a string value though. Some files don't follow the `utf-8` format. For example binary executable and images that we might want to export which will be buffers.

I suggest `amplicationignore?.content?.split('\n') || [];` for future proofness. At the end of the day, this class is called `GithubFile`, not `GithubTextFile`.

yuval-hazaz (approved):

arielweinberger:
Love it

arielweinberger (approved):

Issue #3235, opened 2022-07-26
Update getResources GraphQL query in `amplication/client`


abrl91:
Update the GraphQL queries in the client: `getResources`

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3248 merged to #3205

PR #3240, opened 2022-07-27
Update `getResources` query

abrl91:
Issue Number: #3235 

## PR Details
Updated the `getResources` query:
- Search resources by `projectId`
- separate the query to a different file
- create new custom hook name: `useResources`
- added the returned value form this hook to `AppContext`

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/context/appContext.tsx
+++ b/packages/amplication-client/src/context/appContext.tsx
@@ -9,7 +9,7 @@ export interface AppContextInterface {
   setNewProject: (data: models.ProjectCreateInput) => void;
   onNewProjectCompleted: (data: models.Project) => void;
   resources: models.Resource[];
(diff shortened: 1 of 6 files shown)


arielweinberger (approved):

packages/amplication-client/src/context/appContext.tsx
```diff
@@ -8,6 +8,10 @@ export interface AppContextInterface {
   projectsList: models.Project[];
   setNewProject: (data: models.ProjectCreateInput) => void;
   onNewProjectCompleted: (data: models.Project) => void;
+  resources: models.Resource[];
+  handleSearchChange: (value: any) => void
```
barshimi, line 12: why any and not string, also instead of value => searchResults: string

barshimi (approved):

Issue #3215, opened 2022-07-25
Update sign-up route to new create service wizard 


morhag90:
### Feature description

A new page was added : create service wizard, this page replace the existing create app page. 
Now, we need to change the route, so  after sign-up to it will route to create service wizard. 

### Use case

After user sign-up, route to create service wizard page.

souravjain540:
is this issue in progress? @morhag90

morhag90:
@souravjain540  yes :)

PR #3243, opened 2022-07-27
update route sign-up path to go to service wizard

morhag90:
Issue Number: #3215

## PR Details

After the change in the app routing and build new service wizard, we needed to update the route path, so it will get to the new create service wizard page with the rights values (workspace and project ids). 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
+++ b/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
@@ -41,7 +41,6 @@ const CreateServiceWizard: React.FC<Props> = ({ moduleClass, match }) => {
 
   const { trackEvent } = useTracking();
   const history = useHistory();
-  
   const [createResourceWithEntities, { loading, data, error }] = useMutation<
     TData
   >(CREATE_RESOURCE_WITH_ENTITIES, {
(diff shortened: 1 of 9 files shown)


packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
```diff
@@ -41,7 +41,7 @@ const CreateServiceWizard: React.FC<Props> = ({ moduleClass, match }) => {
 
   const { trackEvent } = useTracking();
   const history = useHistory();
-
+
```
barshimi, line 44: remove space

barshimi (approved):

packages/amplication-client/src/Workspaces/hooks/useWorkspaceSelector.ts
```diff
@@ -54,14 +54,17 @@ const useWorkspaceSelector = (authenticated: boolean) => {
...
-      history.push(`/${data.currentWorkspace.id}`);
+      history.push({
+        pathname: `/${data.currentWorkspace.id}`,
+        search: location.search,
+      });
 
     data &&
       data.currentWorkspace.id === workspace &&
       setCurrentWorkspace(data.currentWorkspace);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
arielweinberger, line 67: What's the reason we need this? I'd rather avoid these small hacks because they tend to pile up. Can we fix it?

Issue #3245, opened 2022-07-28
Add props type to routesUtil and use it where needed


abrl91:
### Feature description

Create a new AppRouteProps that contains the default props that every component needs.

### Use case

In the new architecture, we generically managed the routes and components.
Every component gets certain props and sometimes other props.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3248 merged to #3132

PR #3248, opened 2022-07-28
Add AppMatchRoute and AppRouteProps type

abrl91:
Issue Number: #3245 

## PR Details
Add AppMatchRoute and AppRouteProps type and use where needed

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Project/ProjectPage.tsx
+++ b/packages/amplication-client/src/Project/ProjectPage.tsx
@@ -2,9 +2,15 @@ import React from "react";
 import ResourceList from "../Workspaces/ResourceList";
 import "./ProjectPage.scss";
 import ProjectSideBar from "./ProjectSideBar";
-import { AppRouteProps, AppMatchRoute } from "../routes/routesUtil";
+import { AppRouteProps } from "../routes/routesUtil";
+import { match } from "react-router-dom";
 
-type Props = AppRouteProps & AppMatchRoute;
+type Props = AppRouteProps & {
+  match: match<{
+    workspace: string;
+    project: string;
+  }>;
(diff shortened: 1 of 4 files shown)


packages/amplication-client/src/Project/ProjectPage.tsx
```diff
@@ -1,38 +1,27 @@
...
-    project: string;
-  }>;
-  moduleClass: string;
-  noduleName: string;
-  // eslint-disable-next-line no-undef
-  InnerRoutes: JSX.Element | undefined;
-};
-
-const CLASS_NAME = "project-page";
+type Props = AppRouteProps & AppMatchRoute;
```
yuval-hazaz, line 7: what are you using it for?

packages/amplication-client/src/routes/routesUtil.tsx
```diff
@@ -1,8 +1,29 @@
...
+  RESOURCE = "resource",
+  COMMIT_ID = "commitId",
+  ENTITY_ID = "entityId",
+  FIELD_ID = "fieldId",
+  BUILD_ID = "buildId",
+}
+
+export type AppMatchRoute = {
+  match: match<{ [property in MatchRoute]: string }>;
+};
```
yuval-hazaz, line 18: by doind this you assume you have all the parameters in all the pages and that is not correct. 
especially for the use which use it, they will get a type with all the properties even when some of them are empty.
I don't think this change make sense. also- I don't see any reason why you would include this change in this PR

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #3246, opened 2022-07-28
Update new routes components after route refactor


morhag90:
### Feature description

After routes refactor, we need to update all routes props and ensure everything works as expected.

- [x] #3253
- [x] #3255

### Use case

Every route in the system will open the right page

muhsinkamil:
@morhag90 @abrl91 Can I help with this ? Or you already working on it ?

morhag90:
Hi @muhsinkamil we already working on this issue. 
Thank you :)

PR #3263, opened 2022-07-31
Update route paths and `ResourceHome` component

morhag90:
Issue Number:  #3246 

## PR Details

After routes refactor we needed to update the new routes paths and components 

## PR Checklist


packages/amplication-client/src/Resource/EntitiesTile.tsx
```diff
@@ -23,6 +23,7 @@ const EVENT_DATA: TrackEvent = {
 
 function EntitiesTile({ resourceId }: Props) {
   const history = useHistory();
+  const {workspace,project} = useParams();
```
yuval-hazaz, line 26: did you run formatter? (It seems like you are missing spaces..)

yuval-hazaz, line 26: is that how we are fetching the parameters now? 
aren't we using the context? 
If this stays, I think we should use a type

packages/amplication-client/src/Resource/ResourceHome.tsx
```diff
@@ -18,17 +16,22 @@ import OverviewTile from "./OverviewTile";
...
+
+type Props = AppRouteProps & {
+  match: match<{
+    workspace: string;
+    project: string;
+    resource: string; 
+  }>;
 };
 
 const CLASS_NAME = "resource-home";
```
yuval-hazaz, line 31: should we use moduleClass?

yuval-hazaz (changes requested):

packages/amplication-client/src/Resource/EntitiesTile.tsx
morhag90: i will change it to use context @yuval-hazaz

morhag90: @yuval-hazaz  yes, the CI pass .. i will check again just to be sure

packages/amplication-client/src/Resource/ResourceHome.tsx
morhag90: @yuval-hazaz  do we now make all the refactors needed or only focus on fix the route issues??

packages/amplication-client/src/Resource/EntitiesTile.tsx
yuval-hazaz: @mshidlov, something looks strange in the CI... I would expect the CI to fail on formatting error

packages/amplication-client/src/Resource/ResourceHome.tsx
yuval-hazaz: you are correct

packages/amplication-client/src/routes/appRoutes.tsx
```diff
@@ -75,7 +76,7 @@ export const Routes: RouteDef[] = [
     exactPath: true,
   },
   {
-    path: "/github-auth-app/callback",
+    path: "/github-auth-app/callback", //todo
```
abrl91, line 79: Remove the "todo" :)

packages/amplication-client/src/routes/resourceRoutes.tsx
```diff
@@ -4,17 +4,17 @@ import resourceSettingsRoutes from "./resourceSettingsRoutes";
 
 const resourceRoutes = [
   {
-    path: "/:workspace/:project/:resource/entities",
-    Component: lazy(() => import("../Resource/EntitiesTile")),
+    path: "/:workspace([A-Za-z0-9-]{20,})/:project([A-Za-z0-9-]{20,})/:resource([A-Za-z0-9-]{20,})/entities",
+    Component: lazy(() => import("../Entity/EntityList")),
```
abrl91, line 8: This looks strange. According to the path, it seems like you want this page
image

But you load the `EntityList` component

packages/amplication-client/src/Resource/EntitiesTile.tsx
morhag90: @yuval-hazaz  Done

packages/amplication-client/src/routes/resourceRoutes.tsx
morhag90: @abrl91  this is the path page we open, and the component name is EntityList, looks ok for me, am i missing something else?

packages/amplication-client/src/routes/appRoutes.tsx
morhag90: @abrl91  Because this branch is not my final work on routes i prefer to leave this comment to my new branch :)

yuval-hazaz (approved):

abrl91 (approved):

Issue #3264, opened 2022-07-31
separate project configuration resource


yuval-hazaz:
### Feature description

separate project configuration resource in UseResources hook

PR #3265, opened 2022-07-31
Project Configuration Resource

yuval-hazaz:
Issue Number: #3264 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/hooks/useResources.ts
+++ b/packages/amplication-client/src/Workspaces/hooks/useResources.ts
@@ -9,6 +9,10 @@ type TData = {
 
 const useResources = (currentProject: models.Project | undefined) => {
   const [resources, setResources] = useState<models.Resource[]>([]);
+  const [
+    projectConfigurationResource,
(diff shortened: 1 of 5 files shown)


packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -28,7 +32,15 @@ const useResources = (currentProject: models.Project | undefined) => {
 
   useEffect(() => {
     if (loadingResources || !resourcesData) return;
-    setResources(resourcesData.resources);
+    const projectConfigurationResource = resourcesData.resources.find(
```
barshimi, line 35: you can combine those two loops to one reduce

yuval-hazaz: The list of projects will often include a few items and not more then a few dozens in most extreme cases. 
Also, the find "loop" will stop after finding the first item. So, for readability, I think it is better to keep the current code instead of quizzing another micro improvement

Issue #3209, opened 2022-07-24
⭐️ Unified Logging - `@amplication/nest-logger-module`


arielweinberger:
# Unified Logging
## Introduction
We need a standard logging solution that is easy to consume in our back-end microservices, as a TypeScript package called `@amplication/nest-logger-module`. This should be an abstraction, and not require work with underlying logger solutions.

This solution must support structured logging, which enables us to perform robust and effective queries on Kibana.

## Requirements
- The logger is easily consumable in our monorepo as `@amplication/nest-logger-module`
- Must support the following log levels: `DEBUG` (3), `INFO` (2), `WARN` (1), `ERROR` (0). We can make `INFO` the default.
- Log level should be configurable when importing the module. Default value should be `INFO`.
- Logs are sent in as [Structured Logs](https://www.sumologic.com/glossary/structured-logging/#:~:text=Structured%20logging%20is%20the%20practice,data%20sets%20rather%20than%20text.)
- Local development logs are pretty-printed so they are easy for developers to work with. This should not be done for production logs.
- A detailed `README.md` with an example of how to consume the library.
- This library abstracts whatever technology it uses behind the scenes, such as `nest-winston` etc. Consumers should not have to import Winston-specific modules or types.
- Logs format supports the ELK stack (Elastic, Logstash, Kibana) 

### Logger Library
As originally started at #2616 we will utilize `nest-winston` and create a wrapper around it.

### Structued Logs
The structured JSON should support the following fields:
- timestamp.
- level (`DEBUG`, `INFO`, `WARN`, `ERROR`).
- message (string, required): Human readable log message.
- context (object, optional, defaults `{}`): Useful context related to the process in which the log was produced.
- service (string, required): Identifies the service that produced the log.

## API - NestJS Module
This should be a NestJS module that is easy to consume into microservices.
```ts
// x.module.ts
import { AmplicationLoggerModule } from "@amplication/nest-logger-module";

@Module({
  imports: [
    AmplicationLoggerModule.forRoot({ service: 'amplication-server' });
  ]
})

// x.service.ts
import { AMPLICATION_LOGGER_PROVIDER, Logger } from '@amplication/nest-logger-module';

export class SomeService {
  constructor(@Inject(AMPLICATION_LOGGER_PROVIDER) private readonly logger: Logger) { }
}
```

## Tasks

- [ ] New amplication nest-logger-module
- [ ] Update the existing package to the new logger package

## Next Steps
[...]

PR #3267, opened 2022-07-31
Create `@amplication/nest-logger-module` shared package

tupe12334:
closes: #3209 

## PR Details

This pr build and merge the new package called `@amplication/nest-logger-module` that aim to manage and standard all our logges

## PR Checklist


arielweinberger:
Title does not conform with our guidelines.

tupe12334:
> Title does not conform with our guidelines.

Got it, fix that

packages/amplication-nest-logger-module/src/logger.module.ts
```diff
@@ -0,0 +1,17 @@
+import { DynamicModule, Module } from "@nestjs/common";
+import { WinstonModule } from "nest-winston";
+import { defaultMeta, winstonConfigFactory } from "./config.factory";
+
+export type ModuleOptions = defaultMeta;
+
+@Module({})
+export class AmplicationLoggerModule {
+  static register(options: ModuleOptions): DynamicModule {
```
arielweinberger, line 9:
```suggestion
import { LoggerMetadata, winstonConfigFactory } from "./config.factory";

export interface AmplicationLoggerModuleOptions {
  metadata: LoggerMetadata;
}

@Module({})
export class AmplicationLoggerModule {
  static register(options: AmplicationLoggerModuleOptions): DynamicModule {
```
Let's rename some stuff here to be more specific. `ModuleOptions` is not a good name for autocompletion and will not make developers' life easier. It is very likely to clash with other modules.

As I mentioned above, this is a use case for an `interface` and far from a `type` use case.

Also, let's use `metadata` as an options property and not as the root of the AmplicationLoggerModule. Otherwise, we limit ourselves to the Winston API and cannot deviate from it.

packages/amplication-nest-logger-module/src/logger.module.ts
```diff
@@ -0,0 +1,17 @@
...
+    return WinstonModule.forRootAsync({
+      useFactory: () =>
+        winstonConfigFactory(process.env.NODE_ENV === "production", {
+          ...options,
+        }),
```
arielweinberger, line 14:
```suggestion
        winstonConfigFactory(process.env.NODE_ENV === "production", options.metadata),
```

arielweinberger (changes requested):
Overall really nicely done, added some comments about semantics etc. It is important that we don't use the Winston `defaultMeta` call signature as the configuration of our `AmplicationLoggerModule` because that is really limiting.

Instead, let's use a `.metadata` property for that. This allows for better extensibility in the future if needed.

Finally, we should stop abusing the usage of `type`. It is really a last resort in TypeScript, and in my few good years doing TypeScript I have rarely encountered a good use case for it (unless developing a type inference engine).

You can [read more here](https://stackoverflow.com/questions/57936452/has-type-been-deprecated-in-favor-of-interface). I quote:
> "Because an ideal property of software is being open to extension, you should always use an interface over a type alias if possible."

arielweinberger:
I've changed the base branch to `0.14.10`, as I might be able to release `0.14.9` earlier.

packages/amplication-nest-logger-module/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "target": "es2016" /* Set the JavaScript language version for emitted JavaScript and include compatible library declarations. */,
```
arielweinberger, line 3: I meant to remove _all_ comments from `tsconfig.json` :) If we need to we can check out the docs. I don't think it looks very professional.

arielweinberger (changes requested):

PR #3287, opened 2022-08-02
Added project resolve field to a resource

abrl91:
Issue Number: #3288

## PR Details
When querying a project/projects, the resources field returned as null, I added the project resolver, removed the field to resources

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/project/project.resolver.ts
+++ b/packages/amplication-server/src/core/project/project.resolver.ts
@@ -20,12 +20,16 @@ import { GqlAuthGuard } from 'src/guards/gql-auth.guard';
 import { AuthorizeContext } from 'src/decorators/authorizeContext.decorator';
 import { AuthorizableResourceParameter } from 'src/enums/AuthorizableResourceParameter';
 import { UserEntity } from 'src/decorators/user.decorator';
+import { ResourceService } from '../resource/resource.service';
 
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -49,4 +56,9 @@ export class ProjectResolver {
   ): Promise<Project> {
     return this.projectService.createProject(args, user.id);
   }
+
+  @ResolveField(() => Resource)
```
yuval-hazaz, line 60: I think this should be [Resource]

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -38,4 +38,8 @@ export class ProjectService {
 
     return project;
   }
+
+  async resources(projectId: string): Promise<Resource[]> {
+    return this.prisma.resource.findMany({ where: { projectId: projectId } });
+  }
```
yuval-hazaz, line 44: Please use resourceService

yuval-hazaz: one reason why for example, you didn't filter our deleted resources

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #3300, opened 2022-08-02
🐛 Bug Report: Search is not working in the new project layout


yuval-hazaz:
### What happened?

Search is not working in the new project layout

### What you expected to happen

Search to work

### How to reproduce

click on the search icon in the header

### Amplication version

0.15.0

PR #3301, opened 2022-08-02
Fix search functionality in the new project layout

yuval-hazaz:
Issue Number: #3300 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/CommandPalette/CommandPalette.tsx
+++ b/packages/amplication-client/src/CommandPalette/CommandPalette.tsx
@@ -1,14 +1,15 @@
-import React, { useState, useMemo } from "react";
+import React, { useState, useMemo, useContext } from "react";
 // @ts-ignore
 import ReactCommandPalette from "react-command-palette";
 // @ts-ignore
 import { useQuery, gql } from "@apollo/client";
 import { History } from "history";
 
-import { useHistory, useRouteMatch } from "react-router-dom";
+import { useHistory } from "react-router-dom";
 import { CircleBadge, Icon } from "@amplication/design-system";
 
 import * as models from "../models";
+import { AppContext } from "../context/appContext";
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -101,21 +102,27 @@ type Props = {
 };
 
 const CommandPalette = ({ trigger }: Props) => {
-  const match = useRouteMatch<{ resourceId: string }>("/:resourceId/");
+  const { currentWorkspace, currentProject, currentResource } = useContext(
+    AppContext
+  );
 
-  const { resourceId } = match?.params || {};
+  const projectBaseUrl = useMemo(
```
barshimi, line 109: we need to add a generic function to create this url and check for undefined entities.

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
```diff
@@ -101,21 +102,27 @@ type Props = {
...
+  );
 
   const history = useHistory();
   const [query, setQuery] = useState("");
   const handleChange = (inputValue: string, userQuery: string) => {
     setQuery(userQuery);
   };
   const { data } = useQuery<TData>(SEARCH, {
-    variables: { query },
+    variables: { query, projectId: currentProject?.id },
```
barshimi, line 120: do we need to add skip ?

yuval-hazaz: no, it brings all data when it first opens

packages/amplication-client/src/CommandPalette/CommandPalette.tsx
yuval-hazaz: I agree, I event think it can be part of the context

Issue #3261, opened 2022-07-31
🐛 Bug Report: Inconsistency between the autogenerated project and creating a new project


abrl91:
### What happened?

When a user signs up, a new workspace with an autogenerated project is created with an auto-generated service.

![image](https://user-images.githubusercontent.com/39680385/182019807-cb3b3778-611b-4df3-9034-6a2bd2a3af44.png)

When a user creates a project manually, he gets a project with `Project Configuration`

![image](https://user-images.githubusercontent.com/39680385/182019971-63ade013-310f-4e92-ac19-033238bdda34.png)

### What you expected to happen

I expect that on every project creation, automatically or manually, `Project Configuration` will be created automatically

### How to reproduce

Signup and then create a new project. You will see that the autogenerated project that you get doesn't have `Project Configuration` and the one that you create manually has the `Project Configuration`.

### Amplication version

0.15.0

souravjain540:
@abrl91 labels?

PR #3351, opened 2022-08-04
Same signup workflow

tupe12334:
closes: #3261 

## PR Details

This PR adds the creation of a project and extracts the user creation logic to a separate method

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/auth/auth.service.ts
+++ b/packages/amplication-server/src/core/auth/auth.service.ts
@@ -118,7 +118,7 @@ export class AuthService {
     }
   }
 
-  async bootstrapUser(
+  private async bootstrapUser(
     account: Account,
     workspaceName: string
   ): Promise<AuthUser> {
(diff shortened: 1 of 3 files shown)


packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -103,41 +101,40 @@ export class AuthService {
...
+    account: Account,
+    workspaceName: string
+  ): Promise<AuthUser> {
+    const workspace = await this.createWorkspace(workspaceName, account);
+
+    const [user] = workspace.users;
+
+    await this.accountService.setCurrentUser(account.id, user.id);
+
+    await this.projectService.createProject(
```
arielweinberger, line 127:
```suggestion
    const workspace = await this.createWorkspace(workspaceName, account);
    const [user] = workspace.users;
    
    await this.accountService.setCurrentUser(account.id, user.id);
    await this.projectService.createProject(
```
We use double line spacing with no meaning here.

I've adjusted it to separate between variable declarations and "await" calls. Then it is meaningful. Otherwise people will assume we just insert double linebreaks after every expression/statement.

arielweinberger (changes requested):

arielweinberger:
Let's come up with a better title for this. Strangers looking at it have no idea what it means.

arielweinberger (approved):

Issue #3370, opened 2022-08-06
🐛 Bug Report:  current resource undefined 


morhag90:
### What happened?

When we didn't choose a resource from the resources list (in the header, default value = Resource list) 
Then, click on one of the resources, and the menu list is stuck on loading because the current resource is undefined. 

 

### What you expected to happen

After choose a resource i will route to resource home page with menu list. 

### How to reproduce

Go to project page => in the header Resource list, leave the default value (resource list) =>
Click on one of the resources on the page, and you will get the resource home page without the menu list on the left side (infinity loader). 

### Amplication version

0.15.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #3371, opened 2022-08-06
 Fix current resource undefined when clicking on Resource and in the Header resources list didn't choose resource

morhag90:
Issue Number: #3370 

## PR Details

Fix the current resource context when clicking on resource from the main list (resource list component)

## PR Checklist


yuval-hazaz:
@morhag90 I reviewed the PR but I have some questions, so let me know when you have time, and we will review it together

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
```diff
@@ -34,36 +40,41 @@ type TData = {
...
-      });
-    },
-  });
+  const [createResourceWithEntities, { loading, error }] = useMutation<TData>(
+    CREATE_RESOURCE_WITH_ENTITIES,
+    {
+      onCompleted: (data) => {
+        onNewResourceCompleted(data.createResourceWithEntities);
+      },
+      update(cache, { data }) {
```
barshimi, line 57: why not use refetch instead of absolute update

packages/amplication-client/src/Workspaces/ResourceList.tsx
```diff
@@ -62,6 +63,16 @@ function ResourceList() {
     },
   });
 
+  const handleSelect = useCallback(
```
barshimi, line 66: you can add the track event inside useResource hook and add setResource directly to resourceListItem instead of passing it two levels

packages/amplication-client/src/Workspaces/ResourceListItem.tsx
```diff
@@ -20,13 +20,14 @@ import { AppContext } from "../context/appContext";
 type Props = {
   resource: models.Resource;
   onDelete?: (resource: models.Resource) => void;
+  onSelectResource?: (resource: models.Resource) => void;
```
barshimi, line 23: redundant

packages/amplication-client/src/Workspaces/ResourceListItem.tsx
```diff
@@ -52,10 +53,11 @@ function ResourceListItem({ resource, onDelete }: Props) {
   }, [onDelete, resource]);
 
   const handleClick = useCallback(() => {
+    onSelectResource && onSelectResource(resource);
     trackEvent({
```
barshimi, line 57: you are invoking track event twice under the same event

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -47,6 +48,21 @@ const useResources = (
...
 
+  const resourceRedirect = useCallback(
+    (resourceId: string) =>
+      history.push({
+        pathname: `/${currentWorkspace?.id}/${currentProject?.id}/${resourceId}`,
+      }),
+    [currentWorkspace, history, currentProject]
+  );
+
+  const onNewResourceCompleted = useCallback(
```
barshimi, line 59: I don't understand why you need to expose this outside of the hook. It's a straight forward event that happen after creating new resource

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
morhag90: removed, we don't need it at this point

packages/amplication-client/src/Workspaces/hooks/useResources.ts
morhag90: Done

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -44,9 +52,44 @@ const useResources = (
...
+    onCompleted: (data) => {
+      onNewResourceCompleted(data.createResourceWithEntities);
+    },
+  });
+
+  const createResource = (
+    data: models.ResourceCreateWithEntitiesInput,
+    eventName: string
+  ) => {
+    console.log(data);
```
barshimi, line 86: remove console.log

morhag90: Done

PR #3387, opened 2022-08-07
Remove old tests and mocks in the server

tupe12334:
As a good scout boy.
On my honor, I will do my best
To do my duty to God and my country and to obey the Scout Law;
To help other people at all times;
To keep myself physically strong, mentally awake and morally straight.
@arielweinberger This is for you ✌️

## PR Details

This PR removes old test and mock that was forgotten in the separation of the git-service


morhag90 (commented):
@tupe12334 We talked about it, remove only the classes used by the git-service

arielweinberger (approved):

Issue #3400, opened 2022-08-09
Pending changes panel - client GraphQL queries


abrl91:
Pending changes panel - fix the GraphQL queries to work with commit and padding changes on the project level (instead of the resource level)

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3409

PR #3409, opened 2022-08-09
Implement pending changes panel

abrl91:
Issue Number: #3400 

## PR Details
removed `packages/amplication-client/src/Resource/ResourceLayout.tsx` and `packages/amplication-client/src/Resource/ResourceLayout.scss`

- pending changes
- pending changes status
- commits

### build still doesn't work in the client
It works in the server. This is the query:

```
query {
  builds (where: {resource: {id: "cl6m9gr1f23704r4p5njdg78u0"}}) {
    id
    status
  }
}
```

And this is the result: 

```
{
  "data": {
    "builds": [
      {
        "id": "cl6mazzp574694r4p58q158wsv",
        "status": "Failed"
      },
      {
        "id": "cl6mb6abv78357r4p5erzvh50t",
        "status": "Failed"
      },
      {
        "id": "cl6osclp031812qwp5f28oc7t4",
        "status": "Failed"
      }
    ]
  }
}
```

## PR Checklist


packages/amplication-client/src/VersionControl/DiscardChanges.tsx
```diff
@@ -52,7 +52,7 @@ const DiscardChanges = ({ resourceId, onComplete, onCancel }: Props) => {
       {
         query: GET_PENDING_CHANGES,
         variables: {
-          resourceId: resourceId,
+          projectId: projectId,
```
arielweinberger, line 55:
```suggestion
          projectId,
```

packages/amplication-client/src/VersionControl/DiscardChanges.tsx
```diff
@@ -61,10 +61,10 @@ const DiscardChanges = ({ resourceId, onComplete, onCancel }: Props) => {
   const handleConfirm = useCallback(() => {
     discardChanges({
       variables: {
-        resourceId: resourceId,
+        projectId: projectId,
```
arielweinberger, line 64:
```suggestion
        projectId,
```

packages/amplication-client/src/VersionControl/PendingChangesPage.tsx
```diff
@@ -46,7 +46,7 @@ const PendingChangesPage = ({ match }: Props) => {
   );
   const { data, error } = useQuery<TData>(GET_PENDING_CHANGES, {
     variables: {
-      resourceId: resource,
+      projectId: project,
```
arielweinberger, line 49: Let's rename the parameter from `project` to `projectId` to be consistent. If it's not an entire Project object, then it should be named `projectId`.

arielweinberger (changes requested):

packages/amplication-client/src/VersionControl/PendingChangesPage.tsx
abrl91: I think it is called like that because it comes from `match.params` and in the params is called "project" and not "projected"

abrl91: You are right, It won't take more than a few minutes, but if I change it in `appRoutes` I will have to change it in all the places that we use the `match.params` and this is out of my task scope.

arielweinberger (approved):

PR #3415, opened 2022-08-11
Git repository overwrite and sql migration

### LINKED ISSUE #3386: Project and Service scoped Git repository
### Feature description

Currently, the Git repository is connected directly to the service; as we move forward in our plans to make a project hierarchy, we want to have a project scoped git connection.
A project scoped git repository will be the automatic default for all the services inside the project.
The user can overwrite the project repository and set a custom service repository as an option.
I attached a screenshot from our Figma to show the UI for this situation.

The main task compose from two smaller ones, the UI and the backend
- [ ] Custom git repositories for services (UI)
- [ ] Custom git repositories for services (Backend)

Service - GitHub sync b

### Use case

_No response_

_No response_

tupe12334:
closes #3386 

## PR Details

This pr changes the data models from a Git Repository that saves the connection to Resource to Resource that saves the connection to Git Repository

## PR Checklist


packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -68,7 +68,7 @@ export class GitResolver {
   )
   async deleteGitRepository(
     @Args() args: DeleteGitRepositoryArgs
-  ): Promise<Resource> {
+  ): Promise<boolean> {
```
yuval-hazaz, line 71: why did you decide to change the return type?

packages/amplication-server/src/core/git/git.resolver.ts
```diff
@@ -33,7 +33,7 @@ export class GitResolver {
   async createGitRepository(
     @Args() args: CreateGitRepositoryArgs
   ): Promise<Resource> {
-    return this.gitService.createGitRepository(args.data);
+    return this.gitService.createRemoteGitRepository(args.data);
```
yuval-hazaz, line 36: why did you decide to change the name?

packages/amplication-server/src/core/resource/dto/ResourceWhereInput.ts
```diff
@@ -35,4 +35,7 @@ export class ResourceWhereInput {
     description: undefined
   })
   project?: WhereUniqueInput | null;
+
+  @Field(() => String, { nullable: true })
+  projectId?: string | null;
```
yuval-hazaz, line 40: why did you add `projectId` while we already have the property `project`

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -109,10 +110,23 @@ export class ResourceService {
       throw new InvalidColorError(color);
     }
 
+    if (args.data.resourceType === EnumResourceType.ProjectConfiguration) {
+      throw new AmplicationError(
+        'Try to create project configuration from createResource function'
+      );
```
yuval-hazaz, line 116: Resource of type ProjectConnfiguration cannot be created manually

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -109,10 +110,23 @@ export class ResourceService {
...
+    const projectId = args.data.project.connect.id;
+
+    const projectConfiguration = await this.projectConfiguration(projectId);
+
     const resource = await this.prisma.resource.create({
       data: {
         ...DEFAULT_SERVICE_DATA,
         ...args.data,
+        gitRepository: {
+          connect: { id: projectConfiguration.gitRepositoryId }
```
yuval-hazaz, line 128: This will throw an exception in case projectConfiguration is null (which should not happen logically, but still)

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -337,20 +351,6 @@ export class ResourceService {
       throw new Error(INVALID_DELETE_PROJECT_CONFIGURATION);
     }
 
-    const gitRepo = await this.prisma.gitRepository.findUnique({
```
yuval-hazaz, line 340: you can still delete the gitRepo record in case `override` is true - knowing that the resource we delete is not the  projectConfiguration

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -609,17 +609,37 @@ export class ResourceService {
...
+      })
+    ).gitRepository.gitOrganization;
   }
 
   async project(resourceId: string): Promise<Project> {
-    return this.prisma.project.findUnique({ where: { id: resourceId } });
+    return this.prisma.resource
+      .findUnique({ where: { id: resourceId } })
+      .project();
+  }
```
yuval-hazaz, line 635: we agreed that we should remove this function and the resolver should use the project service instead of the resource service

yuval-hazaz (changes requested):
please see my comments

packages/amplication-server/src/core/resource/dto/ResourceWhereInput.ts
tupe12334: To match more closely to the Prisma orm

packages/amplication-server/src/core/git/git.resolver.ts
tupe12334: To match our convention of calling remote for the non DB git repositories

packages/amplication-server/src/core/git/git.resolver.ts
tupe12334: From now on, which Resource to return?
Git repository have several resources.
So I checked the usage of this endpoint and tested the I don't change existing behavior.

Issue #3410, opened 2022-08-09
Pending changes panel - Show as part of the new Project layout at the project and resource level


abrl91:
Pending changes panel - Show as part of the new Project layout at the project and resource level (requires UI and API changes) => build the pending changes UI

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3416

PR #3416, opened 2022-08-11
Implement pending changes panel - UI

abrl91:
Issue Number: #3410 

## PR Details
- pending changes panel - UI

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/VersionControl/PendingChange.scss
+++ b/packages/amplication-client/src/VersionControl/PendingChange.scss
@@ -1,5 +1,8 @@
 @import "../style/index.scss";
 
+$before-spacing-left: 10px;
+$before-spacing: -6px;
+
 .pending-change {
   margin-left: 0;
   margin-right: 0;
@@ -15,8 +18,8 @@
     flex: 1;
   }
   &__action {
-    height: var(--large-spacing);
-    width: var(--large-spacing);
+    height: var(--medium-spacing);
+    width: var(--medium-spacing);
     font-size: var(--text-field-label-font-size);
     min-width: var(--double-spacing);
     min-height: var(--double-spacing);
@@ -33,10 +36,10 @@
       content: "";
       position: absolute;
       display: block;
-      left: 10px;
+      left: $before-spacing-left;
       border-right: 1px solid var(--black30);
-      top: -6px;
-      bottom: -6px;
+      top: $before-spacing;
+      bottom: $before-spacing;
     }
 
     &.update {
--- a/packages/amplication-client/src/VersionControl/PendingChange.scss
+++ b/packages/amplication-client/src/VersionControl/PendingChange.scss
@@ -14,15 +14,19 @@ $before-spacing: -6px;
     margin-bottom: var(--double-spacing);
   }
 
+  &__link {
+    font-size: var(--title2-font-size);
+  }
+
   &__spacer {
     flex: 1;
   }
   &__action {
-    height: var(--medium-spacing);
-    width: var(--medium-spacing);
+    height: var(--large-spacing);
+    width: var(--large-spacing);
     font-size: var(--text-field-label-font-size);
-    min-width: var(--double-spacing);
-    min-height: var(--double-spacing);
+    min-width: var(--medium-spacing);
+    min-height: var(--medium-spacing);
     border-radius: 2px;
     display: flex;
     align-items: center;
--- a/packages/amplication-client/src/VersionControl/PendingChange.scss
+++ b/packages/amplication-client/src/VersionControl/PendingChange.scss
@@ -4,14 +4,20 @@
   margin-left: 0;
   margin-right: 0;
   margin-bottom: var(--default-spacing);
+  position: relative;
   @include flexFullRow;
+
+  &:last-child {
+    margin-bottom: var(--double-spacing);
+  }
 
   &__spacer {
     flex: 1;
   }
   &__action {
-    height: var(--double-spacing);
-    width: var(--double-spacing);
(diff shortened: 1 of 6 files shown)


packages/amplication-client/src/VersionControl/PendingChange.scss
```diff
@@ -23,6 +29,16 @@
...
     margin-right: var(--default-spacing);
 
+    &::before {
+      content: "";
+      position: absolute;
+      display: block;
+      left: 10px;
+      border-right: 1px solid var(--black30);
+      top: -6px;
+      bottom: -6px;
```
yuval-hazaz, line 39: if you can use existing variables please do so. 
if you need exactly these values, please create SCSS parameters at the top of the file

packages/amplication-client/src/VersionControl/PendingChange.tsx
```diff
@@ -56,7 +51,11 @@ const PendingChange = ({
       );
     }
     if (linkToOrigin) {
-      return <Link to={url}>{change.origin.displayName}</Link>;
+      return (
+        <Link to={url} style={{ fontSize: "12px" }}>
```
yuval-hazaz, line 55: use CSS instead of inline style

packages/amplication-client/src/VersionControl/PendingChanges.scss
```diff
@@ -5,6 +5,15 @@ $counter-height: 18px;
 
 .pending-changes {
   padding: var(--default-spacing);
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+
+  &__divider {
+    border: 1px solid var(--black5);
```
yuval-hazaz, line 13: use $border-black5

packages/amplication-client/src/VersionControl/PendingChanges.scss
```diff
@@ -5,6 +5,15 @@ $counter-height: 18px;
...
+    width: 99%;
```
yuval-hazaz, line 14: dont use 99% - you probably can get the results you want in a different way

packages/amplication-client/src/VersionControl/PendingChanges.scss
```diff
@@ -71,14 +78,32 @@ $counter-height: 18px;
     }
   }
 
+  &__changes-wrapper {
+    display: flex;
+    flex-direction: column;
+    border: 1px solid var(--black10);
```
yuval-hazaz, line 84: use $border-black10

packages/amplication-design-system/src/style/css-variables.scss
```diff
@@ -54,20 +54,30 @@
...
+  --icon-size-medium: 32px;
+  --icon-size-large: 44px;
+  --icon-size-xlarge: 54px;
+  --icon-size-xxlarge: 64px;
+
   --icon-spacing: 4px;
   --default-spacing: 12px;
   --default-spacing-small: 8px;
-  --large-spacing: 16px;
+  --large-spacing: 14px;
```
yuval-hazaz, line 80: did you check where else we use this var and what is the impact?

packages/amplication-design-system/src/style/css-variables.scss
```diff
@@ -54,20 +54,30 @@
 
   --form-horizontal-width: 700px;
   --form-horizontal-label-width: 300px;
-  --title1-font-size: 15px;
+  --large-icon-font-size: 18px;
+  --title1-font-size: 16px;
```
yuval-hazaz, line 58: did you check where else we use this var and what is the impact?

yuval-hazaz (changes requested):
great work

packages/amplication-design-system/src/style/css-variables.scss
abrl91: I didn't check but I asked @alexbass86 about it

packages/amplication-design-system/src/style/css-variables.scss
abrl91: I just added a var and used it. What can it affect?

yuval-hazaz: I thought1 this is not a new one var

packages/amplication-design-system/src/style/css-variables.scss
yuval-hazaz: 68 places.. you may be fixing one place to be exactly 14 as you want, but changing many other places that Alex may want to keep it 16... anyway, since we are working with multiple of 4, I don't think we should use 14 

image

packages/amplication-design-system/src/style/css-variables.scss
abrl91: `--title1-font-size` => always been there
`--large-icon-font-size` => the variable I added

packages/amplication-client/src/VersionControl/PendingChanges.tsx
```diff
@@ -114,20 +104,66 @@ const PendingChanges = ({ projectId }: Props) => {
...
-                  key={change.originId}
-                  change={change}
-                  resourceId={projectId}
-                  linkToOrigin
+          <div className={`${CLASS_NAME}__changes-wrapper`}>
+            {loading ? (
+              <span>Loading...</span>
+            ) : (
+              <div className={`${CLASS_NAME}__changes`}>
+                {pendingChangesByResource.map((group) => (
```
barshimi, line 112: do we need to group it every render ? does it change ?

barshimi (approved):

PR #3482, opened 2022-08-17
Adding bottom margin to tabs

### LINKED ISSUE #3479: 🐛 Bug Report: Inner navigation - Bad padding between tabs
### What happened?

We have bad padding between the tabs inside the roles section at service resource.

Bad
image

How it spouse to look
image

We need to ensure each tab has 8px padding from the next one and from the "add new button."

### How to reproduce

Go to the Roles section.

### Amplication version

0.15.0

### Environment

_No response_

tupe12334:
closes #3479 

## PR Details

This pr add bottom margin to the last tab

## PR Checklist


packages/amplication-client/src/Layout/InnerTabLink.scss
```diff
@@ -19,7 +19,7 @@
   }
 
   &:last-child {
-    margin-bottom: 0;
+    margin-bottom: 8px;
```
yuval-hazaz, line 22: please use var

yuval-hazaz (changes requested):

packages/amplication-client/src/Layout/InnerTabLink.scss
EugeneTseitlin: margin bottom should be 0 for last child. Margin should be applied to a wrapper of list item and not to inner parts.

abrl91 (approved):
I approved, but you still need to fix the CI before merging :)

Issue #1868, opened 2021-11-01
🐛 Bug Report: Exiting "Create new workspace" shows error


MatanForU:
### 📜 Description

when clicking on the X button in order to exit "create new workspace" popup, the input behaving like its an error and not exit smooth.

<img width="479" alt="Screen Shot 2021-11-01 at 9 47 48" src="https://user-images.githubusercontent.com/8940907/139639901-c8b70822-af0f-4978-a268-1f63781a350e.png">
### 👟 Reproduction steps

click new workspace > than click the X button > before it quits it shows an error in the input and quits.

### 👍 Expected behavior

should be exit without any error behavior.

### 👎 Actual Behavior

should be exit the pop-up without any behavior that doesn't require the action that requested.

### 💻 Device

Desktop

### 💻 Operating system

MacOS

### 🌍 Browser

Chrome

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/amplication/amplication/blob/master/code_of_conduct.md)

yuval-hazaz:
@souravjain540 please make sure this is still open, and if it is you may implement the same fix you did in the other issue

Shubhdeep12:
@souravjain540 can I work on this issue

souravjain540:
Sure @Shubhdeep12!

assigned, join our discord for any help :)

Shubhdeep12:
Sure,
thanks.

PR #3509, opened 2022-08-18
fix: error on closing the new workspace Dialog without any name

Shubhdeep12:
Issue Number: #1868

## PR Details

**Required Info.**
- For creating the new workspace, Formik library is used.
- And, a prop validateOnMount was used in order to validate the text input to disable the Create Workspace button.

**What is causing Issue**
- As mentioned in issue, when user click the cross button to close the Dialog, user see the error message, which is not a good UX.
- Its happening because when text field gets blurred, it validates the value, which is wrong in case of empty.
- And, validateOnMount also forcefully cause validation on blur along with its validation on mount.

**Solution**
- Since, after removing the validation on Blur doesn't affect the UX, so in this PR:
1. added validationOnBlur=false
2. removed the validationOnMount
3. And, added a condition to check length of values in button, to make it disable until required condition is not met.

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/NewWorkspace.tsx
+++ b/packages/amplication-client/src/Workspaces/NewWorkspace.tsx
@@ -22,12 +22,14 @@ const INITIAL_VALUES: CreateWorkspaceType = {
   name: "",
 };
 
+const NAME_MIN_LENGTH = 2;
+
 const FORM_SCHEMA = {
   required: ["name"],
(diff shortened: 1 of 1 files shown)


yuval-hazaz (approved):

yuval-hazaz:
@Shubhdeep12  thank you very much for your first contribution with Amplication 🚀

Issue #3586, opened 2022-08-25
🐛 Bug Report: Text over flow in commit selector


tupe12334:
### What happened?

The commit selector in the code, view page is overflow

<img width="427" alt="Screen Shot 2022-08-22 at 17 46 54" src="https://user-images.githubusercontent.com/61761153/186614798-0344ccb0-9c43-45d0-a99d-6060d0511989.png">
<img width="419" alt="Screen Shot 2022-08-25 at 10 56 02" src="https://user-images.githubusercontent.com/61761153/186614802-9008390c-4972-4343-b729-c70992bdb63f.png">

### What you expected to happen

not relevant

### How to reproduce

not relevant

### Amplication version

0.15.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3594

PR #3594, opened 2022-08-25
Fix UI issues on code view selectors

abrl91:
Issue Number: #3586 

## PR Details

Fix UI issues on code view selectors:
- create a new component, `CommitData`, consisting of the user badge, commit message and commit timestamp. This component is also used on the commits page.
- Truncate String with Ellipsis with css

image

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Components/CommitSelectorItem.scss
+++ b/packages/amplication-client/src/Components/CommitSelectorItem.scss
@@ -1,4 +1,5 @@
 @import "../style/index.scss";
+$selector-title-max-width: 180px;
 
 .commit-selector-item {
   display: flex;
@@ -7,7 +8,7 @@
   &__title {
     @include regular-12;
     margin: 0 var(--user-badge-spacing);
-    width: 180px;
+    max-width: $selector-title-max-width;
     white-space: nowrap;
     text-align: left;
     overflow: hidden;
--- a/packages/amplication-client/src/Components/CommitSelectorItem.scss
+++ b/packages/amplication-client/src/Components/CommitSelectorItem.scss
@@ -6,6 +6,6 @@
   margin-right: auto;
   &__title {
     @include regular-12;
-    margin: 0 var(--icon-spacing);
+    margin: 0 var(--user-badge-spacing);
   }
 }
--- a/packages/amplication-client/src/Components/CommitSelectorItem.scss
+++ b/packages/amplication-client/src/Components/CommitSelectorItem.scss
@@ -1,17 +1,11 @@
 @import "../style/index.scss";
 
 .commit-selector-item {
-  @include flexFullRowWithSpacing;
-  flex: 1;
+  display: flex;
+  align-items: center;
+  margin-right: auto;
   &__title {
-    display: flex;
-    flex: 1;
-    white-space: nowrap;
-    text-align: left;
-    overflow: hidden;
-    text-overflow: ellipsis;
-    font-weight: 400;
     @include regular-12;
-    justify-content: space-between;
+    margin: 0 var(--icon-spacing);
   }
 }
--- a/packages/amplication-client/src/Components/CommitSelectorItem.scss
+++ b/packages/amplication-client/src/Components/CommitSelectorItem.scss
@@ -7,5 +7,10 @@
   &__title {
     @include regular-12;
     margin: 0 var(--user-badge-spacing);
+    width: 180px;
+    white-space: nowrap;
+    text-align: left;
+    overflow: hidden;
+    text-overflow: ellipsis;
   }
 }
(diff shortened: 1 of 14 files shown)


packages/amplication-client/src/util/truncatedWithEllipsis.ts
```diff
@@ -0,0 +1,8 @@
+export const truncateWithEllipsis = (
+  input: string | undefined,
+  limit: number,
+  defaultValue: string
+) => {
+  if (!input) return defaultValue;
+  return input.length > limit ? `${input.substring(0, limit)}...` : input;
+};
```
yuval-hazaz, line 8: You should be able to do it with just CSS.

packages/amplication-design-system/src/components/UserAvatar/UserAvatar.scss
```diff
@@ -16,7 +16,7 @@
     text-align: center;
     width: var(--small-menu-item-width);
     color: var(--static-white);
-    font-size: var(--title2-font-size);
+    font-size: var(--text-field-label-font-size);
```
yuval-hazaz, line 19: You made a change to a design system. Did you test the impact on all other places where we use this component? Is this aligned with the design-system requirements?

packages/amplication-client/src/style/mixin.scss
```diff
@@ -468,3 +467,11 @@ $active-indicator-width: 4px;
   line-height: 15px;
   color: #f5f6fa;
 }
+
+@mixin regular-8 {
+  @include font-base();
+  font-weight: 400;
+  font-size: 8px;
+  line-height: 12px;
+  color: #f5f6fa;
```
yuval-hazaz, line 476: I dont think we should include colors in the mixin for fonts

packages/amplication-client/src/Components/ResourceSelectorItem.tsx
```diff
@@ -9,12 +11,23 @@ type Props = {
...
 export const ResourceSelectorItem = ({ resource }: Props) => {
+  const createdAtHour = resource
+    ? formatTimeToNow(new Date(resource?.createdAt))
+    : null;
+
+  const truncateServiceName = truncateWithEllipsis(
+    resource?.name.trim(),
+    15,
+    "Service Name"
+  );
```
yuval-hazaz, line 22: you should use useMemo for things like this otherwise the function will be called with every render

packages/amplication-client/src/Components/CommitSelectorItem.tsx
```diff
@@ -17,12 +19,17 @@ export const CommitSelectorItem = ({ commit }: Props) => {
     <label className={`commit-selector__hour`}>{createdAtHour}</label>
   );
 
+  const truncateCommitMessage = truncateWithEllipsis(
+    commit?.message.trim(),
+    15,
+    "No commit message"
+  );
+
```
yuval-hazaz, line 27: you should use useMemo for things like this otherwise the function will be called with every render

yuval-hazaz (changes requested):

packages/amplication-design-system/src/components/UserAvatar/UserAvatar.scss
abrl91: Yes, it is aligned with the design, and yes I tested the impact.

packages/amplication-client/src/style/mixin.scss
abrl91: Take a look at this whole file. It has a lot of mixins with a color property
image

packages/amplication-client/src/util/truncatedWithEllipsis.ts
abrl91: How? and should I do it? I fixed the places that used this function with `useMemo()`

packages/amplication-design-system/src/components/UserAvatar/UserAvatar.scss
abrl91: @alexbass86 we saw together that the font needs to be 10px, right?

packages/amplication-client/src/Components/CommitSelectorItem.scss
```diff
@@ -1,17 +1,16 @@
...
-  flex: 1;
+  display: flex;
+  align-items: center;
+  margin-right: auto;
   &__title {
-    display: flex;
-    flex: 1;
+    @include regular-12;
+    margin: 0 var(--user-badge-spacing);
+    width: 180px;
```
barshimi, line 10: do you need width or max-width ?

packages/amplication-client/src/style/mixin.scss
barshimi: this is the main color. the rest should run over at the local css file

abrl91: I don't understand, I only added the regular-8 mixin. The rest was there, and they have the color property.

barshimi: I'ts ok

barshimi (approved):

Issue #3601, opened 2022-08-27
🐛 Bug Report: Sync with GitHub selector - `chevron_down` icon position is not permanent


abrl91:
### What happened?

Sync with GitHub selector - `chevron_down` icon position is not permanent
<img width="695" alt="Screen Shot 2022-08-27 at 21 46 03" src="https://user-images.githubusercontent.com/39680385/187044242-72d82153-1994-4354-a5f9-a9216d600435.png">
<img width="692" alt="Screen Shot 2022-08-27 at 21 47 05" src="https://user-images.githubusercontent.com/39680385/187044243-d7fd68a2-c4ea-4dcb-a4ad-3588f1069518.png">

### What you expected to happen

I expect the icon to be located at the same place no matter how long the text is

### How to reproduce

- go to sync with GitHub page
- select two organizations with different names and make sure one is longer than another

### Amplication version

0.15.0

orlandoortegajr:
Hi @abrl91 👋, I found a solution to this issue and submitted a PR for it. Let me know what you think 😃

PR #3605, opened 2022-08-28
fix(client): chevron_down icon position not permanent

orlandoortegajr:
Issue Number: #3601 

## PR Details

The pull request solves the issue of the parent container not properly fitting the width occupied by the child items. Previously this had a 248 px width, however this was causing issues when organization names vary in length, some even exceeding this set width. My proposed solution to this problem is that instead of having a constant width, we make the display: inline-flex (or any other kind of inline would work, I chose this one because adding a display property overrides that of the child, which previously was inline-flex, so to keep the same behavior inline-flex is probably the best option). This way, the parents width matches that of which is occupied by the child so that the chevron_down icon is always a certain amount pixels away from the edge of the select menu border. See screenshots below (keep in mind it is the zoom level that varies within them not the size of the text):

firstsc
secondsc
thirdsc

## PR Checklist

**Jest specified tests don't need to be run as no files changed outside of css**


souravjain540:
@alexbass86 @yuval-hazaz can you guys please give a review on this?

alexbass86:
> [orlandoortegajr](/orlandoortegajr)

Great job! But one of the main reasons we wanted to keep the width 248px is to make sure we have consistency with the look of the whole component no matter the text length, and the problem is that currently, it is not working as planned.
The idea is when we have a long text name, we hide it behind three dots.
 
image

Also, the chevron icon should stay in the same position, no matter the text length. Here is the style attribute for this dropdown field.

height: 28px;
width: 248px;
left: 0px;
top: 19px;
border-radius: 4px;
padding: 5px 8px 5px 8px;

alexbass86 (changes requested):
I added comments on the main issue comment section

orlandoortegajr:
> > [orlandoortegajr](/orlandoortegajr)
> 
> Great job! But one of the main reasons we wanted to keep the width 248px is to make sure we have consistency with the look of the whole component no matter the text length, and the problem is that currently, it is not working as planned. The idea is when we have a long text name, we hide it behind three dots.
> 
> image
> 
> Also, the chevron icon should stay in the same position, no matter the text length. Here is the style attribute for this dropdown field.
> 
> height: 28px; width: 248px; left: 0px; top: 19px; border-radius: 4px; padding: 5px 8px 5px 8px;

I see makes sense, so when the text reaches a certain length we should be truncating it, I'll work on this, should have a solution up soon!

yuval-hazaz:
hi @orlandoortegajr please let us know if you need any help with this

orlandoortegajr:
> hi @orlandoortegajr please let us know if you need any help with this

Hi @yuval-hazaz, unfortunately I have been extremely busy for the past few weeks so I haven't had the time to look further into the issue. I'll unassign for now so that someone else can work on it. When I do get a bit more time I'll check again if it is still up, sorry for the incovenience, best of luck 😄!

yuval-hazaz:
@orlandoortegajr , thank you very much for your contribution! We appreciate it. feel free to reach out anytime

orlandoortegajr:
Hello! @alexbass86 @yuval-hazaz Found some time to continue working on this issue and found the solution, now it should stay in the same location and add an ellipsis at the end. See the following screenshots:

amplificationsc1
amplicationsc2
amplicationsc3

Let me know what y'all think! I tried a few manual tests using the widest letter (M) and thinnest letter (l) to make sure that it truly stays within the limits. I also unassigned myself from the issue so I guess you can assign it back to me if no one has fixed it yet.

yuval-hazaz (approved):

yuval-hazaz:
Thanks! looks good

souravjain540:
@orlandoortegajr  need your email for swag.

orlandoortegajr:
> @orlandoortegajr  need your email for swag.

Hey! It's orlandoortegajr@gmail.com, didn't know you guys gave out swag 😮

souravjain540:
> [orlandoortegajr@gmail.com](mailto:orlandoortegajr@gmail.com)

well now you know ;)

Issue #3619, opened 2022-08-29
🐛 Bug Report: Analytics doesn't  work after routes refactor 


morhag90:
### What happened?

After the refactor of the routes, the analytics doesn't work as before on all the relevant paths. 

### What you expected to happen

Analytics will work as before 

### How to reproduce

open the network => filter by hubspot
Go to resource page: in the network, you will not see an event called: resourceCardClick . 

### Amplication version

0.15.1

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #3620, opened 2022-08-29
Analytics stop work after routes refactor in most of the pages 

morhag90:
Issue Number:  #3619 

## PR Details

Fix the new routes, so the relevant  paths will have analytics 

### DIFF
--- a/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.scss
+++ b/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.scss
@@ -41,7 +41,7 @@ $max-width: 700px;
 
     .circle-badge i {
       color: var(--black5);
-      font-size: 14px;
+      font-size: 20px;
     }
   }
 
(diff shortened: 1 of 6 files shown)


packages/amplication-client/src/routes/routesUtil.tsx
```diff
@@ -90,3 +93,13 @@ export const routesGenerator: (
     </Switch>
   );
 };
+
+
+const pageTracking = (path: string, url: string, params: any)=> {
+  console.log({path});
+  analytics.page(path.replaceAll("/", "-"), {
```
yuval-hazaz, line 100:
```suggestion
  analytics.page(path.replaceAll("/", "-"), {
```

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.scss
```diff
@@ -41,7 +41,7 @@ $max-width: 700px;
 
     .circle-badge i {
       color: var(--black5);
-      font-size: 14px;
+      font-size: 20px;
```
yuval-hazaz, line 44: Is that relevant?

yuval-hazaz (changes requested):

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.scss
morhag90: @yuval-hazaz  I fixes this in another pr , i will remove it

yuval-hazaz (approved):

Issue #3670, opened 2022-08-31
🐛 Bug Report: Gap between main section and Pending Changes


EugeneTseitlin:
### What happened?

![image](https://user-images.githubusercontent.com/6080188/187687150-fa77fd3f-0074-4396-9157-6297c232b007.png)
Gap between main section and Pending Changes is bigger than required
It's 32px instead of 16px.

### What you expected to happen

Gap should be 16px.

### How to reproduce

0.15.2

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #3673, opened 2022-08-31
3670 Make the gap between and pending changes accurate

EugeneTseitlin:
Closes: #3670

## PR Details
Added padding to main section, removed redundant styles. 

## PR Checklist


packages/amplication-client/src/Layout/PageContent.scss
```diff
@@ -2,16 +2,15 @@
...
-  align-items: stretch;
-  padding: 0 var(--large-spacing);
+  flex: 1;
+  gap: var(--large-spacing);
+  padding: var(--large-spacing);
   background: var(--page-content-background);
   box-sizing: border-box;
-  flex: 1;
   overflow: hidden;
-  height: 100%;
```
yuval-hazaz, line 14: @EugeneTseitlin why do we need to update the height property to fix the horizontal gap? If there is another reason to remove it, please describe in the details - I am afraid this may have bad impact on the scroll of pages

packages/amplication-client/src/Layout/PageContent.scss
```diff
@@ -33,20 +31,16 @@ $tabs-width: 300px;
     @include scrollbars(10px, var(--black60), transparent);
     flex: 1;
     overflow: auto;
-    height: 100%;
-    flex: 1;
-    padding: var(--large-spacing);
     position: relative;
     box-sizing: border-box;
-    height: 100%;
```
yuval-hazaz, line 41: same here

packages/amplication-client/src/Layout/PageContent.scss
```diff
@@ -33,20 +31,16 @@ $tabs-width: 300px;
...
 
     @media only screen and (max-width: $page-content-min-width) {
       min-width: calc(#{$page-content-min-width} - var(--menu-width));
     }
   }
+
   .log-container {
     flex: 1;
-    width: 50%;
```
yuval-hazaz, line 49: why did you remove this? is that related to the issue?

yuval-hazaz (changes requested):
@EugeneTseitlin there are several changes to properties that are not related to the issue described here. please add more details or revert the changes

arielweinberger:
We should aim to provide screenshots (preferably "before and after") in UI changes. It's not realistic to have to check out the branch, clean, bootstrap just to test a simple change. Please provide screenshots.

packages/amplication-client/src/Layout/PageContent.scss
EugeneTseitlin: The reason of the bug is redundant styles that conflicted with each other or just were applied not properly. As a solution I've decided to remove everything that is related to positioning (width, height, margin, padding) and make layout from scratch. 
Regarding the height, it should not be used together with flex, flex-grow and flex-shrink are responsible for stretching element relatively to parent container.
Please, see the result on screenshot.

morhag90 (approved):

packages/amplication-client/src/Layout/PageContent.scss
EugeneTseitlin: It's a clean up. Not related specifically to this issue.

Issue #3660, opened 2022-08-30
Remove deleted files from GitHub 


yuval-hazaz:
### Feature description

When files that were generated in the previous build are no longer included in a new build, they should be removed from GitHub. The PR that is created with any new or changed files should also include deleted files.

### Use case

A user is generating a build that does not include some files that were previously included in the build - for example, an entity is deleted, and the entire folder of that entity is no longer generated as part of the code. 

The deleted files should also be deleted from GitHub.

We should use the saved build files and compare the current build with the previous build to identity any files that no longer exist.

PR #3680, opened 2022-09-01
Delete files after removal

tupe12334:
closes: #3660 

## PR Details

This PR adds logic that deletes files in git after removing of data from Amplication

## PR Checklist

### DIFF
--- a/packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
+++ b/packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
@@ -4,12 +4,14 @@ import {
 } from '@amplication/nest-logger-module';
 import { Inject, Injectable } from '@nestjs/common';
 import assert from 'assert';
-import { compare, DiffSet } from 'dir-compare';
+import { compare } from 'dir-compare';
 import { sync } from 'fast-glob';
 import { existsSync, readFileSync } from 'fs';
-import { join, normalize } from 'path';
+import { normalize } from 'path';
 import { PrModule } from '../../constants';
+import { mapDiffSetToPrModule } from './diffset-mapper';
 import { BuildPathFactory } from './utils/BuildPathFactory';
+import { deleteFilesVisitor } from './visitors/delete-files';
 
 @Injectable()
 export class DiffService {
@@ -29,7 +31,7 @@ export class DiffService {
     );
     // return all the new files if an old build folder dont exist
     if (!previousAmplicationBuildId) {
-      return this.firstBuild(newBuildPath);
+      return this.allNewModules(newBuildPath);
     }
     const oldBuildPath = this.buildsPathFactory.get(
       resourceId,
@@ -56,56 +58,25 @@ export class DiffService {
       compareDate: false,
       compareSize: false,
       compareSymlink: false,
+      skipEmptyDirs: true,
     });
 
     this.logger.debug('Finish the dir-compare lib process');
-    if (!res?.diffSet) {
-      throw new Error('');
-    }
-    // const changedFiles = res?.diffSet?.filter((diff) => {
-    //   if (diff.state !== 'equal' && diff.type2 === 'file') {
-    //     //make sure that only new files enter and ignore old files
-    //     if (diff.state !== 'left') {
-    //       return true;
-    //     }
-    //   }
-    //   return false;
-    // });
-    const removedFiles = DiffService.removedFiles(res.diffSet);
-    // this.logger.info('The list of the changed files', { changedFiles });
-    // if (!changedFiles) {
-    //   throw new Error("Didn't have changed files");
-    // }
-    const removedModules = removedFiles.map(({ name1, relativePath }) => {
-      if (!name1) {
-        throw new Error("Didn't have name1");
-      }
-      const path = join(relativePath, name1);
 
-      return {
-        path: path,
-        code: null,
-      };
-    });
-    // const modules = changedFiles.map(async ({ name2, path2, relativePath }) => {
-    //   if (!name2 || !path2) {
-    //     throw new Error("Didn't got valid props");
-    //   }
-    //   const path = join(relativePath, name2);
-    //   const code = await readFileSync(join(path2, name2)).toString('utf8');
-    //   return {
-    //     path,
-    //     code,
-    //   };
-    // });
+    if (!res?.diffSet) {
+      throw new Error('Error in creating a diff set');
+    }
+
+    const modules = mapDiffSetToPrModule(res.diffSet, [deleteFilesVisitor]);
+
     const resultModule = await Promise.all([
-      /*...modules,*/ ...(await this.firstBuild(newBuildPath)),
-      ...removedModules,
+      ...(await this.allNewModules(newBuildPath)),
+      ...modules,
     ]);
     return resultModule;
   }
 
-  private async firstBuild(newBuildPath: string) {
+  private async allNewModules(newBuildPath: string) {
     const basePathLength = newBuildPath.length;
     const files = sync(`${newBuildPath}/**`, { dot: true }).map(
       async (fullPath) => {
@@ -122,18 +93,5 @@ export class DiffService {
 
   private static assertBuildExist(buildPath) {
     assert(existsSync(buildPath));
-  }
-
-  private static removedFiles(diff: DiffSet): DiffSet {
-    return diff.filter((diff) => {
-      if (
-        diff.state === 'left' &&
-        diff.type2 === 'missing' &&
-        diff.type1 !== 'directory'
-      ) {
-        return true;
-      }
-      return false;
-    });
   }
 }
--- a/packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
+++ b/packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
@@ -4,9 +4,11 @@ import {
 } from '@amplication/nest-logger-module';
 import { Inject, Injectable } from '@nestjs/common';
 import assert from 'assert';
+import { compare, DiffSet } from 'dir-compare';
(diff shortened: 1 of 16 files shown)


arielweinberger:
I lack some explanations about your approach to solving this. Please elaborate on the approach you chose, the visitor pattern etc.

packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
```diff
@@ -20,13 +24,14 @@ export class DiffService {
     resourceId: string,
     previousAmplicationBuildId: string | undefined,
     newAmplicationBuildId: string
-  ): Promise<{ path: string; code: string }[]> {
+  ): Promise<PrModule[]> {
     const newBuildPath = this.buildsPathFactory.get(
       resourceId,
       newAmplicationBuildId
     );
+    // return all the new files if an old build folder dont exist
```
arielweinberger, line 32:
```suggestion

```

packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
```diff
@@ -43,52 +48,35 @@ export class DiffService {
...
-    return this.firstBuild(newBuildPath);
-    // // return all the new files if an old build folder dont exist
-    // if (existsSync(oldBuildPath) === false) {
-    //   return this.firstBuild(newBuildPath);
-    // }
+    DiffService.assertBuildExist(oldBuildPath);
+    DiffService.assertBuildExist(newBuildPath);
 
-    // DiffService.assertBuildExist(newBuildPath);
+    this.logger.info({ oldBuildPath, newBuildPath });
```
arielweinberger, line 54: This logger message is meaningless. Please add some human-readable text so we can understand what this means.

For example,
```ts
this.logger.info('This thing does something', { oldBuildPath, newBuildPath });
```

packages/amplication-git-pull-request-service/src/core/diff/diff.service.ts
```diff
@@ -43,52 +48,35 @@ export class DiffService {
...
-    // return await Promise.all(modules);
+    const resultModule = await Promise.all([
+      ...(await this.allNewModules(newBuildPath)),
+      ...modules,
+    ]);
+    return resultModule;
   }
 
-  private async firstBuild(newBuildPath: string) {
+  private async allNewModules(newBuildPath: string) {
```
arielweinberger, line 79: `allNewModules` refers to a noun, but this is a method.

What are we doing here? Reflect that in the name of the method please.

Also, in the context of this method, the `new` in `newBuildPath` is meaningless. It is the only build path the method is aware of. I suggest renaming it to `buildPath`.

packages/amplication-git-pull-request-service/src/core/diff/diffset-mapper.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { DiffSet } from 'dir-compare';
+import { mapDiffSetToPrModule } from './diffset-mapper';
+import { DiffVisitorFn } from './visitors/types';
+
+describe('Testing the diffset module parser', () => {
```
arielweinberger, line 5: We renamed it, so it's not the `diffset module parser` anymore :)

packages/amplication-git-pull-request-service/src/core/diff/diffset-mapper.spec.ts
```diff
@@ -0,0 +1,12 @@
...
+  it('should return an empty array if no visitors are provided', () => {
+    const diffSet: DiffSet = [];
+    const visitors: DiffVisitorFn[] = [];
+    const result = mapDiffSetToPrModule(diffSet, visitors);
+    expect(result).toEqual([]);
+  });
```
arielweinberger, line 11: You are only testing the negative outcome of an empty array. What about the positive outcome? Please add a test for that.

packages/amplication-git-pull-request-service/src/core/diff/diffset-mapper.ts
```diff
@@ -0,0 +1,26 @@
+import { DiffSet } from 'dir-compare';
+import { PrModule } from '../../constants';
+import { DiffVisitorFn } from './visitors/types';
+
+export function mapDiffSetToPrModule(
```
arielweinberger, line 5:
```suggestion
/**
 * This method takes the diffSets and visitor functions and returns an array of modules.
 * We use the visitor pattern here (https://en.wikipedia.org/wiki/Visitor_pattern), which means
 * this function can be provided with any amount of "visitors". Each visitor function will be executed
 * and provided with each diffset.
 * 
 * Empty results (null/undefined) will not be included in the final result.
 */
export function mapDiffSetToPrModule(
```

I believe this function deserves an explanation.

packages/amplication-git-pull-request-service/src/core/diff/visitors/delete-files.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { Difference } from 'dir-compare';
+import { mock } from 'jest-mock-extended';
+import { deleteFilesVisitor } from './delete-files';
+import { missingFolderMock } from './mocks';
+
+describe('Testing the delete files visitor', () => {
+  it('return null when its a new file', () => {
+    // Arrange
+    // The name1 is not existing
```
arielweinberger, line 9:
```suggestion
```
Not sure why this is here.

packages/amplication-git-pull-request-service/src/core/diff/visitors/delete-files.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { Difference } from 'dir-compare';
+import { mock } from 'jest-mock-extended';
+import { deleteFilesVisitor } from './delete-files';
+import { missingFolderMock } from './mocks';
+
+describe('Testing the delete files visitor', () => {
+  it('return null when its a new file', () => {
```
arielweinberger, line 7:
```suggestion
  it('should return null if diff indicates an addition', () => {
```
To be consistent with the `should` you are doing below.

packages/amplication-git-pull-request-service/src/core/diff/visitors/delete-files.ts
```diff
@@ -0,0 +1,24 @@
...
+import { join } from 'path';
+import { PrModule } from '../../../constants';
+
+export function deleteFilesVisitor({
+  state,
+  name1,
+  type1,
+  type2,
+  relativePath,
+}: Difference): PrModule | null {
```
arielweinberger, line 11:
```suggestion
export function deleteFilesVisitor(diff: Difference): PrModule | null {
const { state, name1, type1, type2, relativePath } = diff;
```
I think it would be a lot cleaner to perform destructuring inside the function block.

packages/amplication-git-pull-request-service/src/core/diff/visitors/mocks.ts
```diff
@@ -0,0 +1,11 @@
+import { Difference } from 'dir-compare';
+import { mock } from 'jest-mock-extended';
+
+export const name1PlaceHolder = 'name1';
+
+export const missingFolderMock = mock<Difference>({
+  name1: name1PlaceHolder,
```
arielweinberger, line 7:
```suggestion
export const missingFolderMock = mock<Difference>({
  name1: 'name1',
```
Can be a lot simpler :)

packages/amplication-git-service/src/providers/github.service.ts
```diff
@@ -227,9 +228,15 @@ export class GithubService implements IGitClient {
...
       modules.map(module => {
+        // ignored file
         if (amplicationIgnoreManger.isIgnored(module.path)) {
           return [join(AMPLICATION_IGNORED_FOLDER, module.path), module.code];
         }
+        // Deleted file
+        if (module.code === null) {
+          return [module.path, module.code];
+        }
+        // Regex ignored file
```
arielweinberger, line 239: I think here the comments actually make sense, as they help the developer understand the flow. So this kind of comments we can keep.

arielweinberger (changes requested):

packages/amplication-git-pull-request-service/src/core/diff/visitors/mocks.ts
tupe12334: Yes, but @yuval-hazaz wants us to put consts in outside vars, @yuval-hazaz Please approve / disaprrove

packages/amplication-git-pull-request-service/src/core/diff/diffset-mapper.spec.ts
tupe12334: What do you mean by positive outcome?

arielweinberger: This file only has one test, for an empty array. We are missing tests for a scenario where the array has items in it.

packages/amplication-git-pull-request-service/src/core/diff/visitors/mocks.ts
arielweinberger: Let's not let this be a blocker. If we can move faster, I am okay with keeping it as is.

arielweinberger: I understand, but you are not making use of this exported variable at all ([check here](https://github.com/amplication/amplication/blob/80609f1df85a9cff5198500ad452f994ffe78038/packages/amplication-git-pull-request-service/src/core/diff/visitors/delete-files.spec.ts#L19))

So it has exactly the same meaning if you put it inside your `mock()` function, which contains a bunch of other consts anyway.

arielweinberger (approved):

Issue #2551, opened 2022-04-07
🐛 Bug Report: GitHub repos list is missing repositories


yuval-hazaz:
### What happened?

The Github repos list only shows 30 repos and not all the repos of the user. 
With the previous implementation of the GitHub OAuth app, we could not run a search on the server-side so we only returned the top 30 repos but we also sorted the list on the server-side by creation date so the user at least can find the repo they just created.

In the current implementation with GitHub App we should check whether we can run a search on the server-side (on GitHub API) - if so, we should implement a search. 
If not, we should consider adding pagination.

In any case, to make sure most users are not affected we must sort the list of repos by the creation or last update date. Since most if not all users will search for the repo they just created for their app. 

### What you expected to happen

We should sort the list of repos by the last update date

### How to reproduce

use an organization or an account with more then 30 repos and open the repo selection page

jainpawan21:
As Github API supports all these query parameters, So all requirements can be implemented
<img width="604" alt="Screenshot 2022-05-21 at 5 19 26 PM" src="https://user-images.githubusercontent.com/39362422/169650330-a354d47a-926c-4aaf-9d75-b52553bba534.png">

 Do we need a design update to work on this?

yuval-hazaz:
@jainpawan21, no need for a design to start working on it. 
fee free to implement the pagination and @alexbass86 will review and suggested fixes after you finish

Thanks for your continuous support!

jainpawan21:
Hi @yuval-hazaz ,
What options I should give users to change and what options I should hard code?
1. sort
2. direction
3. per_page
4. page

PR #3695, opened 2022-09-03
fix: add pagination for getOrganisationRepos

jainpawan21:
Issue Number: #2551

## PR Details

https://user-images.githubusercontent.com/39362422/188266026-a75bb107-e518-468d-b2cc-09e2253648e3.mov

## PR Checklist


alexbass86 (changes requested):
It's a great addition, but this isn't the proper design or UX approach for pagination. Please allow me to prepare something that is aligned with our current design.

jainpawan21:
@alexbass86 Any update on design?

alexbass86:
> @alexbass86 Any update on design?
@jainpawan21  Here is the design, once again, thank you for your contribution :)
https://www.figma.com/file/tIfcueoxU4DtYtCHmmLuE7/Community---Task-Assets?node-id=935%3A13277

Pagination

yuval-hazaz:
@alexbass86 I don't think I understand from the design how it should work.
Did you mean we need to use infinite scroll? 
If yes, what is the DD on the top with the page number?

alexbass86:
> @alexbass86 I don't think I understand from the design how it should work. Did you mean we need to use infinite scroll? If yes, what is the DD on the top with the page number?

My fault I forgot to remove the last comp in the design. It fixed now
we will have pages, not an infinity scroll

jainpawan21:
Hi @alexbass86, thanks for design
I see limit (items per page is missing) or I should keep it constant?
and I see pageNo. is dropdown and it will take values from [1 ..... 120]? Isn't this will be too long list for dropdown?

alexbass86:
> Hi @alexbass86, thanks for design I see limit (items per page) is missing or I should keep it constant? and I see pageNo. is dropdown and it will take values from [1 ..... 120]? Isn't this will be too long list for dropdown?

Yes, we will show only eight items per page
about the 120, I give you the edge case here with three numbers, but probably most users won't have more than two pages max :)

jainpawan21:
Updated design as per figma:- 

https://user-images.githubusercontent.com/39362422/189209757-485ee37b-8857-4773-abf9-625ab5bc511c.mov

cc:- @alexbass86 @yuval-hazaz

yuval-hazaz:
@jainpawan21 thanks!

@alexbass86 we can move forward with the current design, but I think there is something strange in the current behavior. I didn't see pagination with a single dropdown. Usually, at least "next" and "prev" buttons are provided. Since all users have no more than a few dozen of repos, the pagination is not for them - we can just use 50 or 100 items per page and solve the issue. 

This solution is for users with hundreds of repos - and for them, I think we just make it harder to use small pages ot 8 items and a drop-down to change pages. 

So I suggest using a larger page size (like 50) and moving forward. And then re-visit this page to consider any further needed improvements. 


@jainpawan21 please resolve the conflicts and use 50 for page size, and we will merge this PR

Thanks for your hard work 💜

alexbass86:
> @jainpawan21 thanks!
> 
> @alexbass86 we can move forward with the current design, but I think there is something strange in the current behavior. I didn't see pagination with a single dropdown. Usually, at least "next" and "prev" buttons are provided. Since all users have no more than a few dozen of repos, the pagination is not for them - we can just use 50 or 100 items per page and solve the issue.
> 
> This solution is for users with hundreds of repos - and for them, I think we just make it harder to use small pages ot 8 items and a drop-down to change pages.
> 
> So I suggest using a larger page size (like 50) and moving forward. And then re-visit this page to consider any further needed improvements.
> 
> @jainpawan21 please resolve the conflicts and use 50 for page size, and we will merge this PR
> 
> Thanks for your hard work 💜

agree

jainpawan21:
@yuval-hazaz Updated this PR with 50 as page size

yuval-hazaz (approved):
impressive work 👏

tupe12334:
Wow @jainpawan21, great work on this issue!

Issue #3692, opened 2022-09-03
⭐️ Feature : New resource type - message broker 


yuval-hazaz:
### Feature description

Add support for a new resource type "Message Broker" to allow communication between services

See the design on Figma
https://www.figma.com/file/GLTbBuQoYvwas6XsCMHnqd/%23none---Resources?node-id=1877%3A8665

PR #3697, opened 2022-09-03
Create `ServiceMessageBrokerConnection` Block 

yuval-hazaz:
Issue Number: #3692 

## PR Details

This PR adds the server-side functionality to manage the connection between a service and a message broker
The functionality is based on the Block infrastructure.

## PR Checklist


packages/amplication-server/src/core/serviceMessageBrokerConnection/dto/ServiceMessageBrokerConnectionUpdateInput.ts
```diff
@@ -0,0 +1,17 @@
...
+  @Field(() => String, {
+    nullable: false
+  })
+  messageBrokerId!: string;
+
+  @Field(() => Boolean, {
+    nullable: false
+  })
+  enabled!: boolean;
+}
```
tupe12334, line 17: Why not give the option for a partial update?

yuval-hazaz: Because these properties are saved in a JSON object in the data, and I don't think we are handling the merging of existing and new values correctly in this field. 
Since we control the API and the client, the current design will allow us to have less errors and move faster.

yuval-hazaz: @tupe12334 
This is a good comment, can you add something about it here?
https://www.notion.so/amplication/Creating-a-new-Block-Type-1d70ff31ae444aceba412fc0f45b28a0

Issue #3675, opened 2022-08-31
After signup, the user should land on the service creation wizard instead of the project page


yuval-hazaz:
### Feature description

When a user signs up (enter the platform for the first time) they should land on the service creation wizard instead of the project page. 
,

PR #3707, opened 2022-09-04
Route to create service wizard after signup with GitHub 

morhag90:
Issue Number: #3675 

## PR Details

After signup with GitHub route to create service wizard instead of the project page

## PR Checklist


yuval-hazaz (approved):
looks great! well done

PR #3718, opened 2022-09-05
fix(server): missing resource id

tupe12334:
This PR make sure that the resource id if getting back in the find one of the block service


packages/amplication-server/src/core/serviceMessageBrokerConnection/serviceMessageBrokerConnection.service.ts
```diff
@@ -77,6 +78,8 @@ export class ServiceMessageBrokerConnectionService extends BlockTypeService<
       where: args.where
     });
 
+    await validateOrReject(block);
```
yuval-hazaz, line 81: why do we need this?

packages/amplication-server/src/models/IBlock.ts
```diff
@@ -15,7 +16,8 @@ export abstract class IBlock {
   })
   createdAt!: Date;
 
-  resourceId?: string;
+  @IsString()
```
yuval-hazaz, line 19: This class is not used to get any input from users or API, so there is no reason to use validation decorators

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #3714, opened 2022-09-04
Display names and links for changed blocks in Pending Changes panel


yuval-hazaz:
### What happened?

The pending changes panel shows a list of changed entities and blocks.
The items in the list are clickable, and on click, the user should navigate to the page of the entity or block. 

<img width="300" alt="image" src="https://user-images.githubusercontent.com/43705455/188328097-e5a31e28-3454-429b-bcde-243e05569d04.png">

We are now adding new types of blocks like "Topic" and "ServiceMessageBrokerConnection" - when clicking on items of these types in the list, the URL used for navigation is not correct. Each block type should navigate to a different page. 

Also, when showing an entity in the list, we show the name of the entity - but not all types of block have a distinct display name, and we should make sure the user can understand what item is in context. 

| type  | text  | url  | 
|---|---|---|
| Entity  | {entity.displayName}  | `/[workspace-id]/[project-id]/[resource-id]/entities/[entity-id]`   |
| ServiceSettings  | "Service Settings"  |  `/[workspace-id]/[project-id]/[resource-id]/settings/update` | 
| ProjectConfigurationSettings  | "Project Settings"  |  `/[workspace-id]/[project-id]/[resource-id]/settings/update` | 
| Topic  |  {block.displayName}  |  `/[workspace-id]/[project-id]/[resource-id]/topics/[topic-id]`   |
| ServiceMessageBrokerConnection  | The name of the message broker that is connected (based on ServiceMessageBrokerConnection.messageBrokerId  |  `/[workspace-id]/[project-id]/[resource-id]/service-connections/[conected-message-broker-id]` | 
| ServiceConnectionTopic  | The name of the topic  |  TBC |

yuval-hazaz:
Also, we may want to add an icon next to each item based on the block type 

@alexbass86  please let us know what you think

PR #3721, opened 2022-09-05
order resource list => fix routes and changes names 

morhag90:
Issue Number: #3714

## PR Details
fix routes and changes names  , more info in issue: #3714
## PR Checklist


packages/amplication-client/src/VersionControl/PendingChange.tsx
```diff
@@ -86,21 +86,40 @@ const PendingChange = ({ change, linkToOrigin = false }: Props) => {
...
+          relativeUrl: `settings/update`,
+          icon: "",
+          displayName: "Project Settings",
+        },
+        [models.EnumBlockType.Topic]: {
+          relativeUrl: `settings/update`,
+          icon: "",
+          displayName: change.origin.displayName,
+        },
+      };
```
yuval-hazaz, line 120: We need to define this function external to the parent function, so it will not be declared for every execution of the parent

yuval-hazaz (changes requested):
@morhag90 one small change in the structure of the mapping function that I missed yesterday

PR #3726, opened 2022-09-06
feat(server): adding patterns in the service message broker connectio…

tupe12334:
This PR add patterns property to the service message broker connection block


packages/amplication-server/src/core/serviceMessageBrokerConnection/dto/messagePattern/EnumMessagePatternConnectionOptions.ts
```diff
@@ -0,0 +1,11 @@
+import { registerEnumType } from '@nestjs/graphql';
+
+export enum EnumMessagePatternConnectionOptions {
+  'None',
+  'Receive', //! I think its need to be listener and broadcaster
```
yuval-hazaz, line 5: we can remove this comment

yuval-hazaz (approved):

Issue #3595, opened 2022-08-26
Project configuration "name" should change the project name


yuval-hazaz:
### Feature description

The `name` field in the `settings` tab in the Project Configuration should change the Project name and not the resource name 

<img width="1306" alt="image" src="https://user-images.githubusercontent.com/43705455/186851396-7b3e140a-f33b-40c2-b338-8f2212a6e4c5.png">

### Use case

When the user wants to change the name of the project, they should go into the project configuration and change the name in the settings tab

jainpawan21:
Hi Yuval, I have figured out the root cause
Can you please assign this to me?

EugeneTseitlin:
@jainpawan21 Go for it.

PR #3747, opened 2022-09-08
fix: add updateProject method in project service

jainpawan21:
Issue Number: #3595 

## PR Details
Added updateProject in project service and resolver [Server]
Added separate file for resource and project update [client]
Updated react-animate-height version to latest to fix one type issue

## PR Checklist


jainpawan21:
> @jainpawan21 your work is epic! well done!
> 
> I imagined something else, let me know what you think: Instead of letting the client being charge of all this logic and acting differently for project configuration - it is better to give the responsibility to the server. so the client does not really care where and how the data is saved. when a client sends a request to update the name field of a resource - the server checks if the resource type is "Project Configuration" and updates the name of the project. and, instead of showing the name of the resource in the UI, we should just show hard-coded "Project Configuration"

I agree Yuval,
This approach is better than handling at client side

yuval-hazaz:
> I agree Yuval,
> This approach is better than handling at client side

Happy to hear that. 
Should I close this PR and wait for a new one?

jainpawan21:
@yuval-hazaz, I updated this PR with your suggestion.

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -52,7 +52,7 @@ const WorkspaceHeader: React.FC<{}> = () => {
       (isResourceRoute && currentResource) ||
       (isResourceRoute && currentProjectConfiguration)
     )
-      return currentResource?.name || currentProjectConfiguration?.name;
+      return currentResource?.resourceType ===  models.EnumResourceType.Service ?  currentResource?.name : 'Project Configuration'
```
yuval-hazaz, line 55: use the condition based on ProjectConfiguration since there are other resource types

```suggestion
      return currentResource?.resourceType ===  models.EnumResourceType.ProjectConfiguration ?  'Project Configuration' : currentResource?.name
```

yuval-hazaz, line 55: use a const instead of hard-coded value and use this const in other pages 

export const PROJECT_CONFIGURATION_RESOURCE_NAME = "Project Configuration";

yuval-hazaz, line 55: Please use the const to show the name also in `ResourceListItem`

yuval-hazaz, line 55: also, after updating the resource name from the client, we need to refetch the project list because with now the project name is not refreshed on the UI

yuval-hazaz (changes requested):
Well done! exactly what I imagined. 
see my comments

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
jainpawan21: Done

jainpawan21: Done

jainpawan21: Done

packages/amplication-client/src/Resource/ResourceForm.tsx
```diff
@@ -67,9 +73,15 @@ function ResourceForm({ match }: Props) {
...
           resourceId: resourceId,
         },
-      }).catch(console.error);
+      })
+        .then(() => {
+          if (currentResource?.resourceType === models.EnumResourceType.ProjectConfiguration) {
+            refetchProjects()
+          }
+        })
+        .catch(console.error);
```
yuval-hazaz, line 82: There is a built-in support in apollo client to reftech query - 
please have a look at this
https://github.com/amplication/amplication/blob/198ba57d95a08fb61118cebff509ac7cff2c6477/packages/amplication-client/src/VersionControl/Commit.tsx#L52

yuval-hazaz (changes requested):
one last comment

PR #3755, opened 2022-09-08
External Apis data from server to dsg

tupe12334:
This PR place the building blocks for the transfer of data of external apis from the server to the DSG


yuval-hazaz:
@tupe12334 please add an issue #

packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -148,3 +149,20 @@ export type ResourceGenerationConfig = {
   dataServiceGeneratorVersion: string;
   appInfo: AppInfo;
 };
+
+export enum EnumMessagePatternConnectionOptions {
+  "None" = "None",
+  "Receive" = "Receive",
+  "Send" = "Send",
+}
```
yuval-hazaz, line 157: The dsg types should be based on the server types and we should not recreated them from scratch

yuval-hazaz (changes requested):
@tupe12334 to my opinion the approach is incorrect.
I don't think we should "prepare" the data for the DSG - if we will start to do it, we will find ourselves very quickly getting a mess of names and types, and for every change in the models, we weill need to maintenance all the new types. 


I think we should just send the data to the DSG "AS IS".
if it is a list of "Service Topics" so just send ServiceTopcis[] as a property  - don't call it "external API" (which means nothing) and don't transform the data before sending to the DSG. 

If the DSG needs the names of the topics, so send it the list of Topics - for example, we may need the name of the topics for other data types and not just the "ServiceTopcis" in the future. 

What I suggest to do is to send the all data structure from the server to the DSG.
Create a new data type for the input of the DSG. instead of multiple parameters - use a single object with properties for the different data pieces:
```typeScript
class DSGResourceData{

   resourceType: ResourceType;
   resourecInfo: AppInfo; //make this object generic so it can include ProjectCOnfigurationSettings | MessageBrokerSeetings | ServiceSettings
   
   entities: Entity[];
   roles: Role[];
   plugins: Plugin[]
   serviceTopics: ServiceTopic[];
   topics: Topic[];

   otherResources: DSGResourceData[]; //a list of all other services, message brokers, and future resource types, with their data. (even Project Configuration)
}
```

packages/amplication-code-gen-types/src/code-gen-types.ts
tupe12334: Because of that we removed the old enum from the server

packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -188,8 +167,18 @@ export type ResourceGenerationConfig = {
   appInfo: AppInfo;
 };
 
-export type Plugin = Omit<
-  models.PluginInstallation,
+export enum EnumMessagePatternConnectionOptions {
```
yuval-hazaz, line 170: move back to server, code-gen-types should get this from models.ts

packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -188,8 +167,18 @@ export type ResourceGenerationConfig = {
...
+  "None" = "None",
+  "Receive" = "Receive",
+  "Send" = "Send",
+}
+
+export type Plugin = BlockOmittedFields<models.PluginInstallation>;
+
+export * from "./dsg-resource-data";
```
yuval-hazaz, line 178: move to index.ts

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -598,4 +588,72 @@ export class BuildService {
...
+    buildId: string,
+    buildVersion: string,
+    user: User,
+    rootGeneration = true
+  ): Promise<CodeGenTypes.DSGResourceData> {
+    const resources = await this.resourceService.resources({
+      where: { project: { resources: { some: { id: resourceId } } } }
+    });
+
+    const resource = resources.find(({ id }) => id === resourceId);
```
yuval-hazaz, line 603: run this first, then look for all other resource by projectId

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -55,7 +56,8 @@ export class ResourceService {
     private serviceSettingsService: ServiceSettingsService,
     private readonly projectConfigurationSettingsService: ProjectConfigurationSettingsService,
     @Inject(forwardRef(() => ProjectService))
-    private readonly projectService: ProjectService
+    private readonly projectService: ProjectService,
+    protected readonly blockService: BlockService
```
yuval-hazaz, line 60: undo the changes on this file

yuval-hazaz (changes requested):
well done!

Issue #3610, opened 2022-08-28
🐛 Bug Report: Commit card - Update components view


alexbass86:
### What happened?

We figured out that the current look and behavior of the commit card aren't working as we thought it would work, and currently, it's not clear enough where the action buttons are and what is the next step in the user flow. 

**Old** - we have three action buttons here (icon, changes, view log)
It's tough to understand that this is clickable, and it's different from other look-a-like components where the whole card is clickable.
![123](https://user-images.githubusercontent.com/96179897/187074512-c081f126-9240-40af-b7b3-709697c557f8.png)

**New** - the whole card is now clickable and works exactly like the resource card and other look-a-like cards. We will be redirected to the view log screen by clicking on the card.
![commit_card](https://user-images.githubusercontent.com/96179897/188451815-70539170-f6b9-4a36-b53c-86a41201ff88.png)

Figma:
https://www.figma.com/file/9vQeA4a87RBsyCViJTh13W/Design-System?node-id=1307%3A13971

### What you expected to happen

We need to be consistent with other components in the system.

### How to reproduce

Go to the commit screen.

### Amplication version

0.15.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

yuval-hazaz:
@alexbass86 can you please update the screenshot with the recent changes as we discussed?

alexbass86:
@arielweinberger let's try to push this change to 15.2

yuval-hazaz:
@alexbass86 I am missing a link to the resource in the new design. Did we remove this on purpose?

abrl91:
> @alexbass86 I am missing a link to the resource in the new design. Did we remove this on purpose?

the commit card is the link to the resource

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3760

PR #3760, opened 2022-09-09
Update commit card UI

abrl91:
Issue Number: #3610 

## PR Details
- update commit card style
- fix commit data width

https://user-images.githubusercontent.com/39680385/189370827-bad4c1bf-81ce-4e5b-8388-5e30d986ba7f.mov

## PR Checklist


packages/amplication-client/src/VersionControl/CommitData.scss
```diff
@@ -1,5 +1,5 @@
 @import "../style/index.scss";
-$metadata-width: 150px;
+$metadata-width: 180px;
```
yuval-hazaz, line 2: The CommitData component is also used in the CommitSelector. just making sure we tested the impact in both places

packages/amplication-client/src/VersionControl/PendingChangeContent.tsx
```diff
@@ -18,7 +18,7 @@ const PendingChangeContent = ({
 }: Props) => {
   const { currentWorkspace, currentProject } = useContext(AppContext);
 
-  const url = `/${currentWorkspace?.id}/${currentProject?.id}/${change.resource.id}/${relativeUrl}`;
+  const url = `/${currentWorkspace?.id}/${currentProject?.id}/${change.origin.resource?.id}/${relativeUrl}`;
```
yuval-hazaz, line 21: by looking at the query in packages/amplication-client/src/Workspaces/queries/projectQueries.ts 
it looks like we fetch "change.resource" and not "change.origin.resource".

I assume there is a reason for this fix, but for my best understanding, it will not work - at least not in all cases 

image

packages/amplication-client/src/VersionControl/CommitResourceListItem.tsx
```diff
@@ -31,54 +31,51 @@ const CommitResourceListItem = ({ build }: Props) => {
...
+            <div className={`${CLASS_NAME}__row`}>
+              <ResourceCircleBadge type={build.resource.resourceType} />
+
+              <span className={`${CLASS_NAME}__title`}>
+                {build.resource.name}
+              </span>
+
+              <span className="spacer" />
               <Link
-                to={`/${currentWorkspace?.id}/${currentProject?.id}/${build.resourceId}`}
```
yuval-hazaz, line 44: I think we should keep the link to the resource somehow. I see that it is not included in the new design. 
@alexbass86 please let me know what you think

yuval-hazaz (changes requested):

packages/amplication-client/src/VersionControl/PendingChangeContent.tsx
abrl91: This fix is for the bug I told you about on the commits page, the link for the changes. In which cases do you think it will not work?

packages/amplication-client/src/VersionControl/CommitResourceListItem.tsx
abrl91: The link for the resource is there. Instead of an icon that holds the link, the whole card is the link

abrl91: You can see it on the video I attached to the PR description

packages/amplication-client/src/VersionControl/CommitData.scss
abrl91: checked, it's OK there

packages/amplication-client/src/VersionControl/PendingChangeContent.tsx
yuval-hazaz: I just don't understand how you get the value in origin.resource if we are not fetching this data from the server 

Based on the query, it looks like the original code is the correct one. I also saw that in action, so I am confused...

yuval-hazaz: @morhag90 can you please have a look at this

packages/amplication-client/src/VersionControl/CommitResourceListItem.tsx
yuval-hazaz: The idea was that a click on the card would link to the log (build page) - the user is looking on a list of builds - the default action should be to navigate to the build, not to the service

abrl91:
@yuval-hazaz  @alexbass86 
The commit card's link redirects to the view log page (build id) and next to the resources name I put back the link icon that redirects to the resource page

yuval-hazaz:
Thanks @abrl91 
I will approve the code changes, but let's wait to hear what @morhag90 says about the pending changes panel

packages/amplication-client/src/VersionControl/PendingChangeContent.tsx
morhag90: @yuval-hazaz @abrl91 In our case, it will not work; as Yuval says, we don't get this value in the query, I checked this by printing the result, and it always returns undefined. 
@abrl91  In which case it breaks the URL??

abrl91: in Commits page => changes

abrl91: @morhag90 @yuval-hazaz undo my change. The purpose of this PR has been completed. I will open an issue about this bug.

abrl91: https://github.com/amplication/amplication/issues/3773

Issue #3582, opened 2022-08-25
🐛 Bug Report: Commit empty state is missing


yuval-hazaz:
### What happened?

Empty state for the commit list looks bad
![image](https://user-images.githubusercontent.com/43705455/186596148-88ffd9a6-ad7c-4010-b8b1-8cd056c727c1.png)

### What you expected to happen

show a proper empty list message with the empty state graphics 

### How to reproduce

create a new project and go to the commits page

### Amplication version

0.15.0

yuval-hazaz:
@alexbass86 please prepare a mock

alexbass86:
Please use this empty state.
https://www.figma.com/file/1BN7FXfK6trj6xsvHLSnk6/%233582%2C-%233583---view-empty-state?node-id=1704%3A3057

![Commit empty state](https://user-images.githubusercontent.com/96179897/188854183-393a7a5a-7211-428b-a8ef-d971667943e6.png)

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3761

Issue #3710, opened 2022-09-04
🐛 Bug Report: New project commits page totally broken after building


arielweinberger:
### What happened?

When building a new project (that has no resources), after hitting Build, the Commits page is totally broken. No build actually happens.

![Screenshot 2022-09-04 at 16 33 32](https://user-images.githubusercontent.com/4976416/188319106-b87c6570-d702-4802-ad72-d759b0f82255.png)

### What you expected to happen

Either something preventing me from building if it's not the right time (maybe I need a resource), or a better error message/use friendly message.

### How to reproduce

1. Create a new project
2. Hit "Commit & Build"
3. Navigate to the specific commit's page

### Amplication version

0.15.3

abrl91:
This issue was fixed by #3761 

@yuval-hazaz this issue happens when the user clicks on the build button right after creating a new service when the only pending change he has is `project configuration settings`.   The resource type of project configuration settings is filtered out from the build process, that is the reason that in the image below, we can see the commit, but there are no builds.

This is a little bit confusing as the page is called `commits page`, but it actually shows a list of builds of a given (clicked) commit.

The question is if this PR (#3761 ) resolved this issue

You can see that I have the last commit, but I don't have any services:
![image](https://user-images.githubusercontent.com/39680385/189410727-799c12f8-6c4e-4937-ad23-18b2db9c5ff3.png)

abrl91:
fixed by https://github.com/amplication/amplication/pull/3761

PR #3761, opened 2022-09-09
Implement commits page empty sate

abrl91:
Issue Number: #3582 

## PR Details
Implement commits page empty sate

image

## PR Checklist


packages/amplication-client/src/Project/ProjectPage.scss
```diff
@@ -4,4 +4,21 @@
...
+    left: 44%;
+    top: 28%;
+    p {
+      @include regular-12;
+      color: var(--black20);
+      margin: 0 auto;
+    }
+    .svg-theme-image {
+      width: 200px;
+      height: 200px;
```
yuval-hazaz, line 21: please use $param at the top of the page

packages/amplication-client/src/Project/ProjectPage.scss
```diff
@@ -4,4 +4,21 @@
   display: flex;
   height: 100%;
   flex: 1 auto;
+
+  .commit-page-empty-state {
+    display: flex;
+    flex-direction: column;
+    position: absolute;
+    left: 44%;
+    top: 28%;
```
yuval-hazaz, line 13: why are we using these excat values?

packages/amplication-client/src/Project/ProjectPage.scss
```diff
@@ -4,4 +4,21 @@
   display: flex;
   height: 100%;
   flex: 1 auto;
+
+  .commit-page-empty-state {
```
yuval-hazaz, line 8: why does this class is defined in project.scss?

packages/amplication-client/src/VersionControl/CommitsPage.tsx
```diff
@@ -26,6 +27,8 @@ const CommitsPage: React.FC<Props> = ({ match, moduleClass }) => {
   const { currentProject, currentWorkspace } = useContext(AppContext);
   const { commits, commitsError, commitsLoading } = useCommit();
   const commitIdx = getCommitIdx(commits, commitId);
+  const commitsBuilds =
+    commits.length > 0 && commitIdx > -1 && commits[commitIdx].builds?.length;
```
yuval-hazaz, line 31: I think we should use different states - one when there are no commits, and a second one when there are no, builds for a specific commit. 

it may be confusing for a user to click on "commit" and see nothing in the list even when there are commits to show. In this case we should show the list of commits and in the space of the builds we need to show a message "There are no builds to show"
We can continue with the current design and open an issue to complete the second state at a later stage.

packages/amplication-client/src/VersionControl/CommitsPage.tsx
```diff
@@ -40,21 +43,26 @@ const CommitsPage: React.FC<Props> = ({ match, moduleClass }) => {
...
-        )}
+      {commitsBuilds ? (
+        <CommitResourceList
+          builds={commits[commitIdx].builds}
+          commitId={commitId}
+        />
+      ) : (
+        <div className="commit-page-empty-state">
+          <SvgThemeImage image={EnumImages.CommitEmptyState} />
+          <p>There are not commits to show</p>
```
yuval-hazaz, line 63:
```suggestion
          <p>There are no commits to show</p>
```

packages/amplication-client/src/style/dark-css-variables.css
```diff
@@ -40,8 +40,8 @@
   --image-drop-excel: url("../assets/images/mono/drop-excel-dark.svg");
   --image-my-resources: url("../assets/images/mono/my-resources-dark.svg");
   --image-generating: url("../assets/images/mono/generating-dark.svg");
-  --image-create-service-wizard : url("../assets/images/mono/service-wizard-create.svg");
-
+  --image-create-service-wizard: url("../assets/images/mono/service-wizard-create.svg");
+  --image-commit-empty-state: url("../assets/images/mono/commit-empty-state.svg");
```
yuval-hazaz, line 44: please not that there are two flavor for each image - one for "dark" and another for "light". 
Altought we are not using the light theme any more, as long as we are keeping the infrastructure for themes, let's keep the new staff aligned. 

so, please rename the image to include the suffix "-dark" at the end, even if we don't have the light version

yuval-hazaz (changes requested):

packages/amplication-client/src/Project/ProjectPage.scss
abrl91: Because that's how we built it when. The page content is of the commits page rendered on the project page. When there are no commits, we render this state on the project page instead the page content.

packages/amplication-client/src/Project/ProjectPage.scss
abrl91: yes. I used position relative on the parent and absolute on the child and I need these exact values

packages/amplication-client/src/VersionControl/CommitsPage.tsx
abrl91: I commented about this on another issue

abrl91: https://github.com/amplication/amplication/issues/3710

packages/amplication-client/src/Project/ProjectPage.scss
yuval-hazaz: 44% may work, but it is not accurate.  especially on larger or smaller screens then the one you used. 
There are a couple of other ways to achieve it - 
If you want to use %, to be more accurate, you can use -  left : calc(50% - $image-width/2)
but a better and easier way to do it is to use flex with align center. 
This way is built proof for any change in the image size and does not require you to deal with %. 
also, it resolve both width and height

```css 
    display: flex;
    align-items: center;
    justify-content: center;
    height: 100%;
    width: 100%;
```

packages/amplication-client/src/Project/ProjectPage.scss
yuval-hazaz: But I see the code that adds the element with the class .commit-page-empty-state in CommitPage

image

packages/amplication-client/src/Project/ProjectPage.scss
abrl91: I will try it. Thanks

packages/amplication-client/src/VersionControl/CommitsPage.tsx
yuval-hazaz: I looked at the issue, and if I understand correctly you agree with what I wrote above - so are we fixing this or opening a new issue to be fixed later?

packages/amplication-client/src/VersionControl/CommitsPage.tsx
```diff
@@ -52,15 +55,16 @@ const CommitsPage: React.FC<Props> = ({ match, moduleClass }) => {
...
           builds={commits[commitIdx].builds}
           commitId={commitId}
         />
       ) : (
-        <div className="commit-page-empty-state">
+        <div className={`${moduleClass}__empty-state`}>
           <SvgThemeImage image={EnumImages.CommitEmptyState} />
-          <p>There are no commits to show</p>
+          {!commitsState && <p>There are no commits to show</p>}
+          {commitsState && !commitsBuildsState && <p>There are no builds to show</p>}
```
yuval-hazaz, line 67: I think it makes more sense to handle the empty state inside the CommitResourceList.
The component should be able to encapsulate the logic related to displaying the list, including when there are no items in the list. 

Also, it will make the code more readable and maintainable, especially if needed to use the same component elsewhere.

yuval-hazaz (approved):
Approved.
Please see my comment about CommitResourceList and decide whether you want to change that.

Issue #3768, opened 2022-09-10
🧩 Add a  plugin for postgreSQL


yuval-hazaz:
### Feature description

Currently, we generate code with PostgreSQL as the database. 
To allow support for using other types of databases like mySQL and MongoDB, we first need to extract the logic that generates the code needed to use Postgres to a plugin. 

Once we have the plugin structure in place, we will be able to create more plugins to support other DBs. 

As a results of this task, the output of the DSG should not change. We expect the exact same code to be generated - just by using a plugin instead of the default behavior of the DSG. 

We can keep the default behavior in place, so when no plugin is being used, PostgreSQL will still be used.

abrl91:
fixed by https://github.com/amplication/plugins/pull/1

abrl91:
fixed by https://github.com/amplication/plugins/pull/1

GreenMachine01:
Approved By:
Amit

PR #3780, opened 2022-09-11
refactor DSG extract `createPrismaSchema` and `createDotEnvModuleInternal` to params and s…

abrl91:
Issue Number: a part of #3768 

## PR Details

`createPrismaSchema` 
- added params and used them in the function body instead of constants
- call the function with the constants variables

`createDotEnvModuleInternal`
- changed param name form 'additionalVariables' to `envVariables`
- removed the content of the `.env` file in the server and extracted it to a constant variable (type `VariableDictionary`)
- use this constant when calling to the function (`createDotEnvModuleInternal`)

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/server/create-dotenv.ts
+++ b/packages/amplication-data-service-generator/src/server/create-dotenv.ts
@@ -29,15 +29,13 @@ export function createDotEnvModule(
  */
 export async function createDotEnvModuleInternal({
   baseDirectory,
-  additionalVariables,
+  envVariables,
 }: CreateServerDotEnvParams["before"]): Promise<Module[]> {
   const context = DsgContext.getInstance;
   const { appInfo } = context;
 
   const code = await readCode(templatePath);
-  const formattedAdditionalVariables = convertToKeyValueSting(
-    additionalVariables
-  );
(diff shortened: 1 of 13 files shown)


packages/amplication-data-service-generator/src/server/create-dotenv.ts
```diff
@@ -65,12 +63,17 @@ function convertToKeyValueSting(arr: VariableDictionary): string {
   if (!arr.length) return "";
   return arr
     .map((item) =>
-      Object.entries(item).map(([key, value]) => `${key}=${value}`)
+      Object.entries(item).map(([key, value]) => {
+        if (!key.trim() || !value.trim())
+          throw new Error("key or value are missing");
```
yuval-hazaz, line 68: Do not throw errors we can use empty strings - it makes more sense to get the code with a missing parameter instead of failing the entire process - also - it will be very hard for the user to understand the reason for the failure.

yuval-hazaz (approved):
well done @abrl91

yuval-hazaz:
@abrl91 I added a single comment, feel free to fix and merge 

@morhag90 please add a plugin event for packages/amplication-data-service-generator/src/server/prisma/create-prisma-schema.ts

PR #3790, opened 2022-09-12
Refactor DSG - generate the docker-compose files instead of sending them as static files

abrl91:
A part of #3768 

## PR Details
Generate the docker-compose files instead of sending them as static files
- create docker-compose folder with 2 files: `docker-comose.yml`, `docker-compose.db.yml`
- create `createDockerComposeFile` function and call it from 'create-server' file

```
export async function createDockerComposeFile(
  baseDirectory: string,
  fileName: string,
  updateProperties: { [key: string]: any } = {}
): Promise<Module> {
  const filePath = path.resolve(__dirname, "docker-compose", fileName);
  const content = await fs.readFile(filePath, "utf-8");
  const parsed = YAML.parse(content);
  const updated = Object.assign(parsed, updateProperties);

  return {
    path: path.join(baseDirectory, fileName),
    code: YAML.stringify(updated, { nullStr: "~" }),
  };
}
```

`updateProperties` => an object that represents the `YAML` part that we want to change

## PR Checklist


abrl91:
@morhag90 I need an event for `createDockerComposeFile`

yuval-hazaz (approved):

Issue #3657, opened 2022-08-30
🐛 Bug Report: when trying to exit 'create new workspace' it shows a typo error 


MatanForU:
### What happened?

When clicking the X it shows a typo error and the icon is not aligned to the center
<img width="605" alt="Screen Shot 2022-08-30 at 17 34 36" src="https://user-images.githubusercontent.com/8940907/187466006-c92b2dd0-2e00-4fd7-8771-184f0309f780.png">

### What you expected to happen

it shouldn't show it

### How to reproduce

.

### Amplication version

0.15.2

### Plan 
Go to "NewWorkspace" which is located at "amplication/packages/amplication-client/src/Workspaces/NewWorkspace.tsx"
Here is the method that should be used in order to inject server side error to the form:
https://formik.org/docs/api/formik#seterrors-fields--field-string-string---void

Plan was written by @EugeneTseitlin

souravjain540:
@alexbass86 please accept/reject this

alexbass86:
> ### What happened?
> When clicking the X it shows a typo error and the icon is not aligned to the center <img alt="Screen Shot 2022-08-30 at 17 34 36" width="605" src="https://user-images.githubusercontent.com/8940907/187466006-c92b2dd0-2e00-4fd7-8771-184f0309f780.png">
> 
> ### What you expected to happen
> it shouldn't show it
> 
> ### How to reproduce
> .
> 
> ### Amplication version
> 0.15.2
> 
> ### Environment
> _No response_
> 
> ### Are you willing to submit PR?
> _No response_

You are right, and it's a bug.

Himanxu1:
assign this to me. @yuval-hazaz

Himanxu1:
should I have to made a PR on release/0.15.3 ?
@souravjain540

souravjain540:
@GreenMachine01 this is fixed

PR #3795, opened 2022-09-13
Fix/3657 fix new workspace validation

AllMikeNoIke:
Issue Number: #3657

## PR Details
Validation error would appear when closing new workspace modal, even without user input. 
This was due to validating on mount and causing error to trigger when clicking outside of focused form field.
Initially resolved this issue by switching to validating upon change. However, this led to the submit button being active by default. Resolved this by adding check for to see if form is pristine (form field values are the same as initial values). If so, it will be disabled. See below for demonstration.

https://user-images.githubusercontent.com/20914059/189778969-a86d0b96-ed32-40ec-a8be-f7f394040d33.mov

## PR Checklist


yuval-hazaz:
Thanks!
I see you also needed to comment out the key in the client .env file 
It was suppose to be commented out anyway, so I will keep it this way and will merge this PR

yuval-hazaz (approved):

PR #3796, opened 2022-09-13
fix docker-compose copy files to `dist`

abrl91:
## PR Details

fix docker-compose copy files to `dist`

## PR Checklist


tupe12334 (approved):

packages/amplication-data-service-generator/scripts/copy-files.ts
```diff
@@ -22,6 +22,8 @@ const ADMIN_SRC_DIRECTORY_GLOB = normalize(ADMIN_SRC_DIRECTORY);
 const GLOB_SOURCES: string[] = [
   `${SRC_DIRECTORY_GLOB}/**/*.template.env`,
   `${SERVER_SRC_DIRECTORY_GLOB}/**/*.template.(ts|env|json)`,
```
yuval-hazaz, line 24:
```suggestion
  `${SERVER_SRC_DIRECTORY_GLOB}/**/*.template.(ts|env|json|yml)`,
```

yuval-hazaz (changes requested):
@abrl91 I think we better stick with the current standard and use the suffix template - see my code suggestion below

Issue #3655, opened 2022-08-30
🐛 Bug Report: Admin UI base directory setting visible even if Admin UI is disabled


arielweinberger:
### What happened?

I don't expect to see the Admin UI area in the Base Directory setting if it is disabled.
<img width="760" alt="Screenshot 2022-08-30 at 15 49 12" src="https://user-images.githubusercontent.com/4976416/187454453-06cb9ac8-db31-4609-a05d-95ab850ec714.png">

### Amplication version

0.15

jainpawan21:
it is disabled but visible in this condition:-
<img width="447" alt="Screenshot 2022-09-03 at 7 04 13 PM" src="https://user-images.githubusercontent.com/39362422/188272835-ced4b2f8-0acd-48d1-be08-1da6bb19af40.png">
 
and
 
visible and editable in this condition
<img width="447" alt="Screenshot 2022-09-03 at 7 02 45 PM" src="https://user-images.githubusercontent.com/39362422/188272766-5bc00ec2-6a47-4887-b4ea-39f8d77d2de6.png">

what should happen in both the conditions?

arielweinberger:
Hey @jainpawan21 I think you are mixing it with a different screen. What you demonstrated is by design - our Admin UI relies on a GraphQL API to function.

With regards to my original issue, my expectation is that if the Admin UI is switched off, I won't see the option to set its base directory at all.

arielweinberger:
@souravjain540 I've set the "good first issue" label to this one, I think it could be good for a seasoned community member to pick up.

rakshitgondwal:
Hey @arielweinberger !!
Can I try taking upon this issue?

PR #3798, opened 2022-09-13
fix(client): admin ui base directory form field appears...

AllMikeNoIke:
… even when admin ui settings are toggled off and type issue for resource id.

Issue Number: #3655 

## PR Details
Resource ID type checking threw an error, as the resource id should either be of type undefined or string from output of optional chaining (use this with caution, as this doesn't throw an error).
Updated the type for our resource id parameter in the useSettings hook, accordingly. If the resource id needs to be a string (even an empty one), we can just do  `const resourceId = currentResource?.id ?? ""` in our directories service settings form.

Demonstration of updated UI below.

https://user-images.githubusercontent.com/20914059/189873999-9e333708-3738-4269-84cb-0a529d0e559f.mov

## PR Checklist


yuval-hazaz (approved):

yuval-hazaz:
Thanks for your contribution @AllMikeNoIke

Issue #3739, opened 2022-09-07
⭐️ DSG: Generate Message Broker Code


yuval-hazaz:
### Feature description

1. Send all the needed Blocks with the relevant data to the DSG 
2. Update the generated code of the service to include the needed kafka module and other parts. based on this PR https://github.com/amplication/kafka-monorepo-demo/pull/22
3. (not in current scope) create a new "code" in the DSG to generated the common parts needed to run Kafka (e.g. docker file ) 

- `.env` file
- `package.json` file
- kafka module in the consumer service and the producer service
- import kafka module in the `app.module` of the consumer service and producer service

PR #3834, opened 2022-09-15
Feat/kafka moudle impl

tupe12334:
closes: #3739 

## PR Checklist


arielweinberger:
@tupe12334 missing details and a proper title, please fix this.

packages/amplication-data-service-generator/src/update-package-jsons.ts
```diff
@@ -24,9 +25,8 @@ export function updatePackageJSONs(
 }
 
 function updatePackageJSON(module: Module, update: Record<string, any>) {
-  const pkg = JSON.parse(module.code);
-
-  Object.assign(pkg, update);
+  let pkg = JSON.parse(module.code);
+  pkg = merge(pkg, update);
```
yuval-hazaz, line 29: Why did you decide to use merge instead of object.assign? 
Also, I think you should look at a solution @abrl91 implemented for YAML files that allow the user to select specific paths to be updated (using _.set)

https://github.com/amplication/amplication/blob/refactor/changes-for-db-plugins/packages/amplication-data-service-generator/src/util/prepare-yaml-file.ts

packages/amplication-data-service-generator/src/plugin-wrapper.ts
```diff
@@ -5,7 +5,7 @@ export type PluginWrapper = (
   func: (...args: any) => Module[] | Promise<Module[]>,
   event: EventNames,
   ...args: any
-) => any;
+) => Promise<Module[]>;
```
yuval-hazaz, line 8: please use a seperate PRs for changes like this for next time, we details about the reason behind the change for the record 

I understand why you did it, and it makes sense but I would love in cases like this to allow the original writer to review (in this case @barshimi )
It can be a mistake but it also may be on purpose

packages/amplication-data-service-generator/src/server/message-broker/create-service-message-broker-modules.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export async function createMessageBrokerInternal({
+  serviceTopicsWithName,
+}: CreateMessageBrokerParams["before"]): Promise<Module[]> {
+  const { serverDirectories } = DsgContext.getInstance;
+  const { messageBrokerDirectory } = serverDirectories;
+
+  const generateMessageBrokerClientOptionsModule = await createGenerateMessageBrokerClientOptionsFunction(
+    messageBrokerDirectory
+  );
```
yuval-hazaz, line 31: use context inside the event

packages/amplication-data-service-generator/src/server/message-broker/generate-message-broker-client-options/generate-message-broker-client-options.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  CreateMessageBrokerClientOptionsFactoryParams,
+  EventNames,
+  Module,
+} from "@amplication/code-gen-types";
+import pluginWrapper from "../../../plugin-wrapper";
+
+export async function createGenerateMessageBrokerClientOptionsFunction(
```
yuval-hazaz, line 8: why the duplicate naem "createGenerate" ?
```suggestion
export async function createMessageBrokerClientOptions(
```

packages/amplication-data-service-generator/src/server/message-broker/generate-message-broker-client-options/generate-message-broker-client-options.ts
```diff
@@ -0,0 +1,22 @@
...
+  eventParams: CreateMessageBrokerClientOptionsFactoryParams["before"]
+): Promise<Module[]> {
+  return pluginWrapper(
+    createGenerateMessageBrokerClientOptionsFunctionInternal,
```
yuval-hazaz, line 12:
```suggestion
    ()=>[],
```

packages/amplication-data-service-generator/src/server/message-broker/generate-message-broker-client-options/generate-message-broker-client-options.ts
```diff
@@ -0,0 +1,22 @@
...
+    EventNames.CreateMessageBrokerClientOptionsFactory,
+    eventParams
+  );
+}
+
+export async function createGenerateMessageBrokerClientOptionsFunctionInternal(
+  eventParams: CreateMessageBrokerClientOptionsFactoryParams["before"]
+): Promise<Module[]> {
+  return [];
+}
```
yuval-hazaz, line 22: Remove this function.

knowing that we may use a lot of skeleton events that will only be handled by plugins with no default behavior, we should create an easier and more readable code for these cases. 


for now I suggested using ()=>[] for an empty function, but maybe we should allow null for the second parameter of the PluginWrapper 

@barshimi what do you think?

packages/amplication-data-service-generator/src/server/message-broker/message-broker-module/create-message-broker-module.ts
```diff
@@ -0,0 +1,22 @@
...
+    EventNames.CreateMessageBrokerNestJSModule,
+    eventParams
+  );
+}
+
+export async function createMessageBrokerModuleInternal(
+  eventParams: CreateMessageBrokerNestJSModuleParams["before"]
+): Promise<Module[]> {
+  return [];
+}
```
yuval-hazaz, line 22: same as above - use ()=>[]

packages/amplication-data-service-generator/src/server/message-broker/message-broker-service/create-message-broker-service.ts
```diff
@@ -0,0 +1,17 @@
...
+      async () => [],
+      EventNames.CreateMessageBrokerService,
+      {}
+    )),
+    ...(await pluginWrapper(
+      async () => [],
+      EventNames.CreateMessageBrokerServiceBase,
+      {}
+    )),
+  ];
```
yuval-hazaz, line 16: I think whether we should use a default structure for the plugin writer in this case - since we request the plugin to create a base file - I think we should provide a starting point. 
what do you think?

packages/amplication-data-service-generator/src/server/message-broker/topics-enum/createTopicsEnum.ts
```diff
@@ -0,0 +1,60 @@
...
+      astEnum.createMember(
+        pascalCase(topic.topicName),
+        dotCase(topic.topicName)
+      );
+    });
+    return astEnum.ast;
+  });
+  astFile.program.body.push(...topics);
+
+  const path = join(serverDirectories.messageBrokerDirectory, "topics.ts"); //TODO get the folder form context
```
yuval-hazaz, line 58: use const for the file name

packages/amplication-data-service-generator/src/util/message-broker.ts
```diff
@@ -0,0 +1,37 @@
+import { DSGResourceData, ServiceTopics } from "@amplication/code-gen-types";
+import { EnumResourceType } from "../models";
+
+export function fromServiceTopicsToTopicsWithName(
```
yuval-hazaz, line 4:
```suggestion
export function resolveTopicNames(
```

packages/amplication-data-service-generator/src/util/message-broker.ts
```diff
@@ -0,0 +1,37 @@
...
+  serviceTopicsList: ServiceTopics[],
+  messageBrokers: DSGResourceData[]
```
yuval-hazaz, line 6: I think you can send the entire dsgResourceData object (or event the context object) to the function, and then extract the messageBrokers and topics instead of asking the calling function to do it for you

packages/amplication-data-service-generator/src/util/message-broker.ts
```diff
@@ -0,0 +1,37 @@
...
+): ServiceTopics[] {
+  messageBrokers.forEach((resource) => {
+    if (resource.resourceType === EnumResourceType.MessageBroker) {
+      throw new Error("resourceType is not EnumResourceType.MessageBroker");
+    }
+  });
```
yuval-hazaz, line 12: if you extract the data from the context in the function - you don't need this validation

yuval-hazaz (changes requested):
well done @tupe12334 
I added a few comments and questions.
Also, you must merge with the latest 0.17 after @abrl91 merges her branch.

packages/amplication-data-service-generator/src/update-package-jsons.ts
tupe12334: I think that @abrl91 also used merge, merge function works recursively.

packages/amplication-data-service-generator/src/util/message-broker.ts
tupe12334: I don't think like you, we need to send only the relevant information to the function

packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -167,12 +169,13 @@ export type ResourceGenerationConfig = {
   appInfo: AppInfo;
 };
 
-export type Plugin = BlockOmittedFields<models.PluginInstallation>;
+export type PluginInstallation = BlockOmittedFieldsWithoutId<
```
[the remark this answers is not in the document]
tupe12334: We need the id of the topic to the merge topics and serviceTopics

Issue #3826, opened 2022-09-15
DSG: Default plugins


yuval-hazaz:

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3843

GreenMachine01:
Approved By:
Amit
Yuval
Shimi

PR #3843, opened 2022-09-16
Implement default plugins registration

abrl91:
Issue Number: #3826 

## PR Details
Implement default plugins registration

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/register-plugin.ts
+++ b/packages/amplication-data-service-generator/src/register-plugin.ts
@@ -7,17 +7,7 @@ import {
 
 class EmptyClass {}
 
-export const POSTGRESQL_PLUGIN_ID = "@amplication/plugin-db-postgres";
-export const POSTGRESQL_NPM = "@amplication/plugin-db-postgres";
-
 const functionsObject = ["[object Function]", "[object AsyncFunction]"];
-const defaultPlugins: Plugin[] = [
-  {
-    pluginId: POSTGRESQL_PLUGIN_ID,
-    npm: POSTGRESQL_NPM,
-    enabled: true,
-  },
-];
 
 /**
  * generator function that import the plugin requested by user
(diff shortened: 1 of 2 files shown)


packages/amplication-data-service-generator/src/register-plugin.ts
```diff
@@ -41,9 +51,10 @@ async function* getPluginFuncGenerator(
  */
 const getAllPlugins = async (pluginList: Plugin[]): Promise<Events[]> => {
   if (!pluginList.length) return [];
+  const allPlugins = [...pluginList, ...defaultPlugins];
   const pluginFuncsArr: Events[] = [];
 
-  for await (const pluginFunc of getPluginFuncGenerator(pluginList)) {
+  for await (const pluginFunc of getPluginFuncGenerator(allPlugins)) {
```
yuval-hazaz, line 57: I prefer that this logic will not be part of the plugin engine, rather it is better to do it on the entry point of the DSG when preparing all other data for the generation process.  

maybe here: https://github.com/amplication/amplication/blob/a598f4d2861388ae5d43a3058ddd0f162da58023/packages/amplication-data-service-generator/src/create-data-service-impl.ts#L32

packages/amplication-data-service-generator/src/register-plugin.ts
```diff
@@ -41,9 +51,10 @@ async function* getPluginFuncGenerator(
  */
 const getAllPlugins = async (pluginList: Plugin[]): Promise<Events[]> => {
   if (!pluginList.length) return [];
+  const allPlugins = [...pluginList, ...defaultPlugins];
```
yuval-hazaz, line 54: it is better to check whether the default plugins are not listed already - to avoid duplicate execution of the same plugin

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #3769, opened 2022-09-10
🧩 Add a plugin for mySQL


yuval-hazaz:
### Feature description

Following the completion of #3768 , we want to add a new plugin to support code generation with mySQL instead of PostgreSQL. 

The output of the generated code should include a docker file for mySQL, and a Prisma generator for mySQL. 

Also, due to some limitation of mySQL, the plugin should validate the list of entity fields and throw an error when the user tries to generate code for "Multi Select Option Set"

arielweinberger:
test

GreenMachine01:
Approved By:
Amit

PR #3844, opened 2022-09-16
Throwing an error when MySQL plugin is installed and one of the entities' fields types is MultiSelectOptionSet

abrl91:
Issue Number: part of #3769 

## PR Details

Throwing an error when MySQL plugin is installed and one of the entities' fields types is MultiSelectOptionSet

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/create-data-service-impl.ts
+++ b/packages/amplication-data-service-generator/src/create-data-service-impl.ts
@@ -19,7 +19,7 @@ import { createServerModules } from "./server/create-server";
 import DsgContext from "./dsg-context";
 import pluralize from "pluralize";
 import { camelCase } from "camel-case";
-import registerPlugins from "./register-plugin";
+import registerPlugins, { MYSQL_PLUGIN_ID } from "./register-plugin";
 import { get } from "lodash";
 import { SERVER_BASE_DIRECTORY } from "./server/constants";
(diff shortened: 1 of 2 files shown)


abrl91:
@MoFoGo I assigned you as a reviewer because I need your input about the error message

packages/amplication-data-service-generator/src/create-data-service-impl.ts
```diff
@@ -128,7 +132,13 @@ function prepareEntityPluralName(entities: Entity[]): Entity[] {
...
 
-function resolveLookupFields(entities: Entity[]): Entity[] {
+function resolveLookupFields(
+  entities: Entity[],
+  installedPlugins: Plugin[]
+): Entity[] {
+  const isMySQLPluginInstalled = installedPlugins.find(
+    (plugin) => plugin.pluginId === MYSQL_PLUGIN_ID
+  );
   const entityIdToEntity: Record<string, Entity> = {};
```
yuval-hazaz, line 142: this validation should be part of the plugin and not part of DSG
The plugin should handle one of the main events (e.g. beforeCreateServer) and if needed the plugin will:
1. write the error message to the logger 
2. throw and error? or "skip" the entire process

yuval-hazaz (changes requested):

abrl91:
Will be handled on the plugin

Issue #3811, opened 2022-09-14
🐛 Bug Report: Open with GitHub button open the repository and not the last pull request


abrl91:
### What happened?

Open with GitHub button open the repository and not the last pull request.

### What you expected to happen

I expect the button to open the last pull request 

### How to reproduce

- create a project with a service
- connect to GitHub
- build
- click on the Open with GitHub button on the footer
- see that the link redirects to the GitHub repository

### Amplication version

0.16.1

### Are you willing to submit PR?

Yes I am willing to submit a PR!

abrl91:
Fixed by https://github.com/amplication/amplication/pull/3950

PR #3950, opened 2022-09-24
Fix open pull request on GitHub button

abrl91:
Issue Number: #3811 

## PR Details
- when the user is in the context of a service (resource), the open with GitHub button opens the last pull request (last build)
- in a context of  project, the open with GitHub button opens the repository URL that is connected to the **project**.

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/WorkspaceFooter.tsx
+++ b/packages/amplication-client/src/Workspaces/WorkspaceFooter.tsx
@@ -10,6 +10,7 @@ import "./WorkspaceFooter.scss";
 import * as models from "../models";
 import { useQuery } from "@apollo/client";
 import { Link } from "react-router-dom";
+import { PUSH_TO_GITHUB_STEP_NAME } from "../VersionControl/BuildSteps";
 
 type TDataCommit = {
   commits: models.Commit[];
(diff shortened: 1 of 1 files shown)


abrl91:
"in a context of project, the open with GitHub button opens the repository URL that is connected to the project"

@GreenMachine01 @yuval-hazaz 
This is the decision I have made, but we may think about it.
A project can have more than one resource. Each resource can be connected to a different git repository. 
When the user clicks on the build button, it creates a commit with x number of builds, a build to every resource on the project. 
We can decide that the button opens the last build no matter what, but then it can cause strange behavior. For example, the user is on one of his services pages, and clicks on the open with GitHub button, but the current service is not the one that was built last, and as a result, the wrong pull request will be opened

packages/amplication-client/src/Workspaces/WorkspaceFooter.tsx
```diff
@@ -86,13 +102,23 @@ const WorkspaceFooter: React.FC<{}> = () => {
...
+              </a>
+            ) : (
+              <a
+                className={`${CLASS_NAME}__gh-link`}
+                href={gitRepositoryUrl}
+                target="github"
+              >
+                Open With GitHub
+              </a>
+            )}
```
yuval-hazaz, line 121: use only one <a> and make sure that value returned from githubUrl is equal to githubUrl when needed

packages/amplication-client/src/Workspaces/WorkspaceFooter.tsx
```diff
@@ -75,6 +76,21 @@ const WorkspaceFooter: React.FC<{}> = () => {
     />
   );
 
+  const githubUrl = useMemo(() => {
+    if (!lastResourceBuild?.action?.steps?.length) {
+      return null;
```
yuval-hazaz, line 81:
```suggestion
      return gitRepositoryUrl;
```

packages/amplication-client/src/Workspaces/WorkspaceFooter.tsx
```diff
@@ -75,6 +76,21 @@ const WorkspaceFooter: React.FC<{}> = () => {
...
+    }
+    const stepGithub = lastResourceBuild?.action.steps.find(
+      (step) => step.name === PUSH_TO_GITHUB_STEP_NAME
+    );
+
+    const log = stepGithub?.logs?.find(
+      (log) => !isEmpty(log.meta) && !isEmpty(log.meta.githubUrl)
+    );
+
+    return log?.meta?.githubUrl || null;
```
yuval-hazaz, line 91:
```suggestion
    return log?.meta?.githubUrl || gitRepositoryUrl;
```

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #3949, opened 2022-09-23
🐛 Bug Report: Generated app server - `Invalid role(s): []` error


abrl91:
### What happened?

based on the a discord thread: https://discord.com/channels/757179260417867879/1022628873210429511

![image](https://user-images.githubusercontent.com/39680385/192026193-50ab7ae7-00ea-4504-8225-2cce8345e357.png)

### How to reproduce

- create a project with a service
- clone and run the app
- login with admin/admin
- make any other request

### Amplication version

1.0.0

shubhwip:
Looks similar to #3745

abrl91:
> 

Thank you. It's similar but in this case, it is not in a test

tupe12334:
This issue is resolved in #3957

sukiejosh:
I am still facing the issue 

![Screenshot 2022-09-30 182230](https://user-images.githubusercontent.com/44656210/193323888-6f68453d-eed8-4033-9258-b83a7c432678.png)

My Payload

![Screenshot 2022-09-30 182600](https://user-images.githubusercontent.com/44656210/193324263-56cf0756-ec53-454c-aa88-1e82e0bbab20.png)

tupe12334:
> I am still facing the issue
> 
> ![Screenshot 2022-09-30 182230](https://user-images.githubusercontent.com/44656210/193323888-6f68453d-eed8-4033-9258-b83a7c432678.png)
> 
> My Payload
> 
> ![Screenshot 2022-09-30 182600](https://user-images.githubusercontent.com/44656210/193324263-56cf0756-ec53-454c-aa88-1e82e0bbab20.png)

Please wait for the 1.0.2 release

sukiejosh:
> > I am still facing the issue
> > ![Screenshot 2022-09-30 182230](https://user-images.githubusercontent.com/44656210/193323888-6f68453d-eed8-4033-9258-b83a7c432678.png)
> > My Payload
> > ![Screenshot 2022-09-30 182600](https://user-images.githubusercontent.com/44656210/193324263-56cf0756-ec53-454c-aa88-1e82e0bbab20.png)
> 
> Please wait for the 1.0.2 release

Alright..thank you

PR #3957, opened 2022-09-25
Refactor user role json value

abrl91:
Issue Number: #3949 

## PR Details
Refactor user role json value

## PR Checklist


yuval-hazaz:
@tupe12334 this change by @abrl91 revert the changes I did on the Roles field, and instead of using an object with a property with an array, it seems likes it uses an object with an array (without the property).
I think this is also what you mentioned in our talk last week. 

But, I also noticed @abrl91 didn't check this fix on the Admin-UI, Which I guess will not work with the current API because it will not know how to send a JSON value to the server. 

So, if the Admin-UI is not working, I think we are back to the original state, and we need to understand what made @abrl91 do this change.

yuval-hazaz (changes requested):
before moving forward, we need to test this E2E, including the Admin UI

tupe12334 (approved):

PR #3982, opened 2022-09-28
Add parser option for static and templates

tupe12334:
This PR is adding a new set of parser options for the template files and for the statics files.
What was:
regular files - compiled and lint
templates - compiled, lint with @ts-ignore, and copy
static - compiled, lint with @ts-ignore and, copy

Now:
regular files - same
templates - not compiling, lint with **custom** configurations, and copy
static - not compiling, lint with regular configuration, and copy

On the way, we added types to the js files with @type new standard.


arielweinberger (approved):
Really well done! 👍

PR #3996, opened 2022-09-30
Install DSG plugins on-demand

arielweinberger:
## PR Details
This PR is the basis for dynamic plugin installations.

1. It uses env variables rather than argument positions for DSG, this is so we can re-use parameters across scripts.
2. Introduces an `install-plugins.ts` script that reads the spec file and dynamically runs `npm install` for plugins in the DSG project, right before running the generation script.
3. Removes all default plugins from `package.json` (we might want to include official plugins to save build times until we have a cache mechanism in place).
4.  Change the `Dockerfile` base image to non-alpine, so it works on M1 and ARM64 computers.

This will most likely break existing tests and functionality. Purely here for visibility for now.

NOTE: Follow-up tickets will be created regarding caching opportunities, as installation times are quite long. However, this, at the bare mininum, will allow us to go live with community plugins loaded on-demand.

https://user-images.githubusercontent.com/4976416/193320668-0943562e-3d80-4d8a-9f0e-226e6c763a66.mov

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/.dockerignore
+++ b/packages/amplication-data-service-generator/.dockerignore
@@ -4,3 +4,4 @@
 !src/server/static/package-lock.json
 !src/admin/static/package.json
 !src/admin/static/package-lock.json
+temp
--- a/packages/amplication-data-service-generator/.gitignore
+++ b/packages/amplication-data-service-generator/.gitignore
@@ -7,3 +7,4 @@ generated/
 *.map
 # ESLint
 .eslintcache
+temp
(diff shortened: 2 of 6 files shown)


yuval-hazaz:
@arielweinberger, what is the current status of this PR? is that still updated, given the recent changes on the plugin packaging?

arielweinberger:
@yuval-hazaz I suggest closing this one. I believe the bundled approach is better than NPM installing in every parameter.

yuval-hazaz:
@tupe12334 @arielweinberger if this PR is not needed, please close it

PR #4046, opened 2022-10-04
Change kafka brokers env vars to string

shaharblanksela:
As part of the helm-charts improvement, the environment variable 'KAFKA_BROKERS' will be changed to string type, so that in the future we can pass the variable as a reference and not manually.


arielweinberger:
Let's wait with merging this as I am currently refactoring for Nx Workspaces.

ShyamGit01:
KAFKA_BROKERS env changed to string

Issue #4016, opened 2022-10-02
🐛 Bug Report: Duplicate label in Topic UI


arielweinberger:
### What happened?

When creating a topic and giving it a user-friendly name such as "Payment Received", we get two input fields.

- The first input field is the display name, a human-friendly name (in this case, `Payment Received`).
- The second input field is the actual topic name as Kafka sees it. It will become `payment.received`.

Both these fields are currently labeled "Name".

### What you expected to happen

The first field should be labeled "Display Name".
<img width="736" alt="Screenshot 2022-10-02 at 11 31 50" src="https://user-images.githubusercontent.com/4976416/193447656-ed4b4237-51d7-4835-abab-3c9c16241f88.png">
 

### How to reproduce

1. Create a resource of type Message Broker
2. Go to Message Broker
3. Navigate to the "Topics" view on the left-hand nav menu
4. Create a new topic

### Amplication version

1.0.0

Ekaji:
Hi, can i work on this issue?

souravjain540:
> Hi, can i work on this issue?

Assigned. Ask any questions in order to make a PR if needed:)

Ekaji:
unfortunately i have run into some issues setting up this project on my system, as vscode crashes after running ``` npm run setup:dev``` at step 3, please assign the issue to someone else while i figure out whats going on

souravjain540:
> unfortunately i have run into some issues setting up this project on my system, as vscode crashes after running ` npm run setup:dev` at step 3, please assign the issue to someone else while i figure out whats going on

you can ask that in discord server. we would love to help.

Utkarsh212:
@souravjain540, may you assign this issue to me so that I can work on it

wiseaidev:
It looks like an easy issue to grab for me; lemme work on it. But, take a look at how fast my internet connection is:
![Screenshot from 2022-10-07 20-56-07](https://user-images.githubusercontent.com/62179149/194620737-d83c44c0-fb83-4fb8-b250-57a507cdb64e.png)
Blazingly fast!

yuval-hazaz:
> Seems like it's already solved here?

@ranrib see the discussion in #4078

Abhay5855:
@yuval-hazaz I would love to take up this issue can you please assign it to me.Thanks

yuval-hazaz:
sorry @Abhay5855 this issue is already taken

souravjain540:
any updates @wiseaidev

[...]

Issue #4015, opened 2022-10-02
Topic naming validation message is not consistent with our styleguide


arielweinberger:
### Feature description

Topic name validation is not consistent with how we do validation at Amplication.

<img width="898" alt="image" src="https://user-images.githubusercontent.com/4976416/193447359-0944ea46-769e-495b-bc47-244d5ca91fda.png">

Notice the tooltip/popover.

### Use case

This is what error messages should look like:
![image](https://user-images.githubusercontent.com/4976416/193447380-6a6ed7b0-09ca-4115-b367-571e89f37e34.png)

You can see the code for this in the Project Settings or Service Settings view.

Asmit2952:
Hey @arielweinberger I can work on this issue.

arielweinberger:
@Asmit2952 Go ahead! Let us know if you need our assistance.

wiseaidev:
Another easy one. Oh wait, it is assigned! Good luck!

yuval-hazaz:
hi @Asmit2952 I just saw that @wiseaidev 's  work on another issue would also resolve this issue 
This is the PR that handles the other issue
https://github.com/amplication/amplication/pull/4078

Sorry if you already spent time working on this. 
I will unassign you and will assign the @wiseaidev

yuval-hazaz:
@Asmit2952 let me know if you see anything else that you want to work on and I would love to assign it to you

Asmit2952:
> @Asmit2952 let me know if you see anything else that you want to work on and I would love to assign it to you

Hey @yuval-hazaz I was going to make a PR today. But no problem, I'll be happy to work on another issue. Can you assign me any beginner friendly issue related to docs or frontend? I'll be happy to work on it.

yuval-hazaz:
@Asmit2952 I created a new issue that may fit your needs. 
https://github.com/amplication/amplication/issues/4083
please comment on the issue and I will assign it to you

wiseaidev:
This ticket is fixed in [pull/4084](https://github.com/amplication/amplication/pull/4084) which is merged into the `next` branch in [pull/4084](https://github.com/amplication/amplication/pull/4126) by @barshimi.

PR #4078, opened 2022-10-07
Assign `TopicNameField` label from props

wiseaidev:
Issue Number: #4016

## PR Details

Pass the value to label from props instead of static value.

## PR Checklist
- [X] `npm test` doesn't throw any error (Due to resources limitation, I ran this command in my brain, and it passes.)

### DIFF
--- a/packages/amplication-client/src/Topics/TopicForm.tsx
+++ b/packages/amplication-client/src/Topics/TopicForm.tsx
@@ -8,6 +8,7 @@ import { validate } from "../util/formikValidateJsonSchema";
 
 import FormikAutoSave from "../util/formikAutoSave";
 import TopicNameField from "../Components/TopicNameField";
+import { DisplayNameField } from "../Components/DisplayNameField";
 
 type Props = {
   onSubmit: (values: models.Topic) => void;
@@ -64,12 +65,12 @@ const TopicForm = ({ onSubmit, defaultValues }: Props) => {
     >
       <Form childrenAsBlocks>
         <FormikAutoSave debounceMS={1000} />
-        <TopicNameField
+        <DisplayNameField
           name="displayName"
           label="Display Name"
-          minLength={1}
+          required
         />
-        <TopicNameField name="name" label="Name" />
+        <TopicNameField label="Name" name="name" />
         <TextField name="description" label="Description" textarea rows={3} />
       </Form>
     </Formik>


packages/amplication-client/src/Components/TopicNameField.tsx
```diff
@@ -41,7 +41,6 @@ const TopicNameField = ({ name, ...rest }: Props) => {
       <TextInput
         {...field}
         {...rest}
-        label="Name"
```
yuval-hazaz, line 44: this can stay "Name"

packages/amplication-client/src/Topics/TopicForm.tsx
```diff
@@ -69,7 +69,7 @@ const TopicForm = ({ onSubmit, defaultValues }: Props) => {
           label="Display Name"
```
yuval-hazaz, line 69: replace this intance of <TopicNameField/> with <DisplayNameField/>

yuval-hazaz (changes requested):
The fix should be replacing the first element to use <DisplayNameField/> instead of <TopicNameField/>

packages/amplication-client/src/Components/TopicNameField.tsx
wiseaidev: The label property is deconstructed from rest: `...rest` <- label is in here.

packages/amplication-client/src/Topics/TopicForm.tsx
wiseaidev: The issue is not here. This is a correct label.

wiseaidev:
I don't think your requested changes are reasonable. Try these changes and lemme know. It should fix the issue.

yuval-hazaz:
@wiseaidev thanks for your comment. 
I agree with you that the fix you sent will fix the issue described original about the label. 
but, while reviewing your code I noticed that the root cause for the issue is in that fact that we use <TopicNameField/> for both the name and the display name fields.
This impact the behavior not just in the label (which is indeed wrong) but also in the validation of the input. 
The display name have a different validation then the topic name. 

So, to fix the issue, we just need to replace the wrong usage of <TopicNameField/> with <DisplayNameField/>
this will resolve the label issue, and also al other side effects

yuval-hazaz:
@wiseaidev I just saw this issue https://github.com/amplication/amplication/issues/4015 which will be resolved by your fix.

wiseaidev:
Hey @yuval-hazaz, Ready for review as well.

yuval-hazaz (approved):

yuval-hazaz:
thanks @wiseaidev for the good work and your contribution to Amplication

Issue #4079, opened 2022-10-08
Automatically add the contributers images in readme file and make it more beautifull


Ankurrawat-12:
### Feature description

it will automatically add the images of the contributes to the readme file.

### Are you willing to submit PR?
Yes
_No response_

Beast71421:
can I have this issue, I would like to work on it.

yuval-hazaz:
Hi @Beast71421 we would love to do that. 
Before we move forward, can you please describe the solution you have in mind - or send us a link to a similar solution/workflow?

Beast71421:
![image](https://user-images.githubusercontent.com/78417225/194696108-b8a19882-afaa-4966-a958-6b67420e3ddc.png)
this will be the outcome of the solution that I have in mind and it will automatically update the contributed image's in readme file

MuhammadJufri:
Can you assign this issue to me please?

Beast71421:
@yuval-hazaz can I have this issue now? and can you also add hecktober fest label : )

05him:
I would love to solve this issue, and you will like my work too

yuval-hazaz:
The idea of auto-generating the image of all contributors is great, but we need to consider the impact.
Right now, we are following the [all-contributors](https://github.com/all-contributors/all-contributors) specification, where all types of contributors get credit for things like ideas, bugs, docs, and so on.

There is a problem with manually updating the list right now, but replacing it with the solution you suggest is not good enough.

If we can get an automatic workflow to update the README when needed, it will be better.

I am adding @MichaelSolati to the conversation to see if he can think of a process to automate it.

Beast71421:
ok thanks for the suggestion @yuval-hazaz

yuval-hazaz:
> @yuval-hazaz this will include every contributors image

I understand, but only contributors who created PR

Beast71421:
yup, but thanks for the feedback on my issue, I really appreciate it

wiseaidev:
`all-contributors` already has [a bot](https://github.com/all-contributors/app) for this. Just invoke on a PR or issue:

```sh
@all-contributors please add @wiseaidev for code
```

yuval-hazaz:
Yep, the problem was that we needed to remember to do that with every PR, issue, etc... 
if we can do it automatically at least for PRs, that will be one step forward, then we just need to remember to do it for issues 

any ideas?

yuval-hazaz:
also, the changes in the README are also hard to track.. 
[...]

PR #4080, opened 2022-10-08
Fix: Automatically add the contributers images

MuhammadJufri:
Issue Number: #4079 

## PR Details
Automatically add the contributors images in the readme file

## PR Checklist
- [ ] tests for the changes have been added
- [ ] npm test doesn't throw any error (Due to resources limitation, I ran this command in my brain, and it passes.)


yuval-hazaz:
thanks @MuhammadJufri 
The idea of auto-generating the image of all contributors is great but we need to consider the impact. 
Right now, we are following the [all-contributors](https://github.com/all-contributors/all-contributors) specification, where all types of contributors get credit for things like ideas, bugs, docs, and so on.

There is a problem with manually updating the list right now, but replacing it with the solution you suggest is not good enough. 

If we can get an automatic workflow to update the README when needed it will be better

I am adding @MichaelSolati to the conversation to see if he can think of a process to automate it. for now I will not merge the PR until we decide wether we want to use it or not

MuhammadJufri:
Okay, Thanks

Issue #3962, opened 2022-09-26
🐛 Plugin Event: inconsistent type for updateValues in CreatePackageJson event


yuval-hazaz:
### What happened?

The type of `updateValues` in `CreateServerPackageJsonParams` is inconsistent with other events that provide the same functionality. 

In other events, we expose an array that allows the plugin writer to push new values to the processes without knowing anything about the current structure or other plugins. Each push is handled separately. 

In the `CreateServerPackageJsonParams`, we expose a single object, which means that one plugin writer can mistakenly overwrite the value provided by previous plugins. 

### CreateServerPackageJsonParams
```
export interface CreateServerPackageJsonParams extends EventParams {
  updateValues: { [key: string]: JsonValue };
}
```

### Other events

```
export interface CreateServerDockerComposeParams extends EventParams {
  fileContent: string;
  updateProperties: { [key: string]: any }[];
  outputFileName: string;
}

export interface CreateServerDockerComposeDBParams extends EventParams {
  fileContent: string;
  updateProperties: { [key: string]: any }[];
  outputFileName: string;
}
```

### What you expected to happen

The type of updateValues should be `updateProperties: { [key: string]: any }[];`

In addition, when changing the type, we should update all plugins that uses this event 

### How to reproduce

n/a

yuval-hazaz:
@arielweinberger I assigned this to you to get your opinion before moving to one of the eng for implementation

arielweinberger:
@yuval-hazaz This is an interesting one. On the one hand, I like the idea of `JsonValue` because it will disallow stuff like functions, so it is more type safe, and feels more appropriate.

On the other hand, `JsonValue` comes from an external package (`type-fest`), which means external plugins will need to have it installed, and I am not a great fan of that.

I'll vote for being consistent and using `{ [key: string]: any }[];`, and I am okay with this change moving to our sprint/backlog.

abrl91:
Fixed by https://github.com/amplication/amplication/pull/4131

PR #4131, opened 2022-10-11
Fix plugin event inconsistency in the type of `CreateServerDockerComposeParams`

abrl91:
Issue Number: #3962

## PR Details
- update CreateServerDockerComposeParams event params and update the DSG code
- add `CreateAdminUIPackageJson` plugin event
- update code-gen-types version and publish

## PR Checklist


abrl91:
@yuval-hazaz @barshimi I will bump the amplication-code-gen-types and the plugin version  after you approve this PR

packages/amplication-data-service-generator/src/update-package-jsons.ts
```diff
@@ -24,9 +24,9 @@ export function updatePackageJSONs(
   });
 }
 
-function updatePackageJSON(module: Module, update: Record<string, any>) {
+function updatePackageJSON(module: Module, update: { [key: string]: any }[]) {
   let pkg = JSON.parse(module.code);
-  pkg = merge(pkg, update);
+  pkg = merge(pkg, ...update);
```
yuval-hazaz, line 29: In the previous cases, we used forEach (see screenshot). I just want to make sure that the behavior of merge with the spread operator will yield the same results as forEach

image

packages/amplication-data-service-generator/src/update-package-jsons.ts
```diff
@@ -49,16 +49,16 @@ function updatePackageJSON(module: Module, update: Record<string, any>) {
...
   const lockfile = JSON.parse(module.code);
   /**
    * The v2 lockfile contains a package in the packages section that is a mirror of the project so we need to update the static data there also
    */
   const pkg = lockfile.packages[""];
 
-  Object.assign(lockfile, update);
-  Object.assign(pkg, update);
+  Object.assign(lockfile, ...update);
+  Object.assign(pkg, ...update);
```
yuval-hazaz, line 61: I know this is not part of this PR but still asking - is there a reason we use object.assign instead of merge like in the other places ?

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -87,7 +87,7 @@ export interface CreateMessageBrokerClientOptionsFactoryParams
 export interface CreateMessageBrokerServiceParams extends EventParams {}
 export interface CreateMessageBrokerServiceBaseParams extends EventParams {}
 export interface CreateServerPackageJsonParams extends EventParams {
-  updateValues: { [key: string]: JsonValue };
+  updateProperties: { [key: string]: any }[];
```
yuval-hazaz, line 90: just a reminder to update the plugins that use this event

abrl91: https://github.com/amplication/amplication/pull/4131#issuecomment-1275163791

packages/amplication-data-service-generator/src/update-package-jsons.ts
abrl91: I think it's by mistake. I'll fix it

packages/amplication-data-service-generator/src/update-package-jsons.ts
```diff
@@ -12,64 +12,27 @@ import { merge } from "lodash";
...
+  const pkg = preparePackageJsonFile(module, update);
 
   if (!semver.valid(pkg.version)) {
     delete pkg.version;
   }
 
   return {
     ...module,
-    code: JSON.stringify(lockfile, null, 4),
+    code: pkg,
```
barshimi, line 36: what is the difference between pkg and JSON.stringify(pkg, null, 2) ?

packages/amplication-data-service-generator/src/util/prepare-yaml-file.ts
```diff
@@ -6,8 +6,7 @@ export function prepareYamlFile(
   updateProperties: { [key: string]: any }[]
 ): string {
   const parsed = YAML.parse(yamlFileContent);
-  updateProperties &&
-    updateProperties.forEach((updateProperty) => merge(parsed, updateProperty));
+  updateProperties && merge(parsed, ...updateProperties);
```
barshimi, line 9: we need to make sure it's not nested objects cause merge is flat merge and not deep merge. maybe it's better to use deep merge then merge

packages/amplication-data-service-generator/src/util/preparePackageJsonFile.ts
```diff
@@ -0,0 +1,15 @@
...
+  module: Module,
+  updateProperties: { [key: string]: any }[]
+): any {
+  const parsedPkg = JSON.parse(module.code);
+  updateProperties &&
+    updateProperties.forEach((updateProperty) =>
+      merge(parsedPkg, updateProperty)
+    );
+
+  return JSON.stringify(parsedPkg, null, 2);
```
barshimi, line 14: don't we want to return the module as Module (cade & path) ?
cause at the moment we return only code.

packages/amplication-data-service-generator/src/update-package-jsons.ts
abrl91: It isn't, but pkg returns it so I use it instead

packages/amplication-data-service-generator/src/util/prepare-yaml-file.ts
abrl91: I should revert it. it was just for test

packages/amplication-data-service-generator/src/util/preparePackageJsonFile.ts
abrl91: The function that is calling this function returns the module with the path and code.

packages/amplication-data-service-generator/src/update-package-jsons.ts
```diff
@@ -12,64 +12,27 @@ import { merge } from "lodash";
...
-  let pkg = JSON.parse(module.code);
-  pkg = merge(pkg, update);
-
-  if (!semver.valid(pkg.version)) {
-    delete pkg.version;
-  }
+    const updateModule =
+      module.path === `${baseDirectory}/package.json`
+        ? updatePackageJSON(module, update)
+        : module;
```
yuval-hazaz, line 21: since we remove the logic for package-lock I don't think we need this condition at all. 

```suggestion
        updatePackageJSON(module, update);
```

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/update-package-jsons.ts
abrl91: Actually, it seems like we do have to check the path.
After I made this change, I ran the tests and they failed. 
I looked at it with the debugger and it seems like it run this function on other paths that we don't want

image

abrl91: with the check, it calls the function only when the path is `${baseDirectory}/package.json` and then the tests pass
image

yuval-hazaz: @abrl91 there is no reason to send this function any file other than the files that needed to be manipulated. 

I looked at the code and noticed that in the admin-ui we send the function ALL the static files, instead of filtering the files.
This behavior need to be fixed - and then we will be able to remove the condition  

image

abrl91: Got it. I will fix it

yuval-hazaz (approved):

Issue #4136, opened 2022-10-12
Add "createResolver" event to DSG


barshimi:

GreenMachine01:
@abrl91 @barshimi I talked with Yuval and there is nothing preventing us from advancing with this item.
You need to look at his feedback and proceed with the item.

abrl91:
Fixed by https://github.com/amplication/amplication/pull/4137

Issue #4097, opened 2022-10-10
Docs: Plugin Events Article - CreateEntityService


MoFoGo:
### Describe the issue with documentation

This event needs to be included in the docs. 

### How to solve the problem

Create an article describing the Plugin Event. 

### The following can be used as examples for the documentation. 

https://github.com/amplication/amplication/issues/4095

https://github.com/amplication/amplication/issues/4094

### For reference, the Plugin Events tree is shown here:

https://www.notion.so/amplication/Plugins-Events-bbb45d7062c64c5ba5dcfba3fa0bcb4b

abrl91:
# Create Entity Service

## Description

Creates the service of an entity

## Event Name:
`CreateEntityService`

## Event Params

```tsx
export interface CreateEntityServiceParams extends EventParams {
  entityName: string;
  templateMapping: { [key: string]: any };
  passwordFields: EntityField[];
  serviceId: namedTypes.Identifier;
  serviceBaseId: namedTypes.Identifier;
  template: namedTypes.File;
}
```

### entityName

### templateMapping

### passwordFields

### serviceId

### serviceId

### template

MoFoGo:
@MoFoGo (note to self) - See this comment. The content above had been transferred to docs, but make sure this is acted upon: 

@MoFoGo this parameter and the previous parameter are commonly used in a lot of events. we should create a separate document that explains about this process and the options of how to manipulate these parameters. 
We should also use a common language to describe these parameters for all events.

PR #4137, opened 2022-10-12
add createResolver & createResolverBase events

barshimi:
Issue Number: #4136 

## PR Details
Add CreateEntityResolver and CreateEntityResolverBase events support to DSG

## PR Checklist


packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -225,3 +226,24 @@ export type PrismaDataSource = {
...
+  UPDATE_MUTATION: ASTNode | undefined;
+  DELETE_MUTATION: ASTNode | undefined;
+  CREATE_ARGS: ASTNode | undefined;
+  UPDATE_ARGS: ASTNode | undefined;
+  DELETE_ARGS: ASTNode | undefined;
+  FIND_MANY_ARGS: ASTNode | undefined;
+  FIND_ONE_ARGS: ASTNode | undefined;
+  CREATE_DATA_MAPPING: ASTNode | undefined;
+  UPDATE_DATA_MAPPING: ASTNode | undefined;
+}
```
yuval-hazaz, line 249: Instead of using this type, we want to use a generic type in order to allow the plugin writer to add a mapping.

The event parameter should be identical to the other events like CreateEntityController, and we should use 

templateMapping: { [key: string]: any };

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -93,3 +96,28 @@ export interface CreateServerPackageJsonParams extends EventParams {
...
   modulesFiles: Module[];
 }
+
+export interface CreateEntityResolverParams extends EventParams {
+  templatePath: string;
+  entityName: string;
+  entityServiceModule: string;
+  serviceId: namedTypes.Identifier;
+  resolverBaseId: namedTypes.Identifier;
+  mapping: ResolverMapping;
```
yuval-hazaz, line 106:
```suggestion
  templateMapping: { [key: string]: any };
```

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -93,3 +96,28 @@ export interface CreateServerPackageJsonParams extends EventParams {
 export interface CreateServerAppModuleParams extends EventParams {
   modulesFiles: Module[];
 }
+
+export interface CreateEntityResolverParams extends EventParams {
+  templatePath: string;
```
yuval-hazaz, line 101: instead of sending the template Path, we should send the actual template - see the examples of createEntityController

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -93,3 +96,28 @@ export interface CreateServerPackageJsonParams extends EventParams {
...
+  entityServiceModule: string;
+  entity: Entity;
+  entityDTO: NamedClassDeclaration;
+  serviceId: namedTypes.Identifier;
+  resolverBaseId: namedTypes.Identifier;
+  createArgs: NamedClassDeclaration | undefined;
+  updateArgs: NamedClassDeclaration | undefined;
+  createMutationId: namedTypes.Identifier;
+  updateMutationId: namedTypes.Identifier;
+  mapping: ResolverMapping;
```
yuval-hazaz, line 122: same as above about the path and the mapping

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -93,3 +96,28 @@ export interface CreateServerPackageJsonParams extends EventParams {
...
+}
+
+export interface CreateEntityResolverBaseParams extends EventParams {
+  templateBasePath: string;
+  entityName: string;
+  entityType: string;
+  entityServiceModule: string;
+  entity: Entity;
+  entityDTO: NamedClassDeclaration;
```
yuval-hazaz, line 115: this shoud be taken from the context (can be remove from here)

yuval-hazaz (changes requested):

packages/amplication-code-gen-types/src/plugin-events-params.ts
yuval-hazaz: see this relevant issue about another event that should also be fixed
https://github.com/amplication/amplication/issues/4156

yuval-hazaz (changes requested):

packages/amplication-code-gen-types/src/code-gen-types.ts
barshimi: I tried to avoid that, cause the inner function is expecting this specific map. if the user will add whatever he wants, it will fail

packages/amplication-code-gen-types/src/plugin-events-params.ts
barshimi: why ? it can be replaced in the after event function. I think we should stick to our logic when it related to our function and dynamic in the after function (it make more sense)

packages/amplication-code-gen-types/src/plugin-events-params.ts
barshimi: see my above comment

packages/amplication-code-gen-types/src/code-gen-types.ts
yuval-hazaz: we must use a generic type, otherwise the plugin writer will not be able to do anything 
please copy the behavior from 
have a look at this issue with the draft for the doc

https://github.com/amplication/amplication/issues/4097

packages/amplication-code-gen-types/src/plugin-events-params.ts
yuval-hazaz: in the after, the template is no longer relevant. 
see the draft for the doc here
https://github.com/amplication/amplication/issues/4097

packages/amplication-code-gen-types/src/plugin-events-params.ts
```diff
@@ -111,3 +113,28 @@ export interface CreateEntityModuleBaseParams extends EventParams {
...
+  template: namedTypes.File;
+  entityName: string;
+  entityServiceModule: string;
+  serviceId: namedTypes.Identifier;
+  resolverBaseId: namedTypes.Identifier;
+  templateMapping: { [key: string]: any };
+}
+
+export interface CreateEntityResolverBaseParams extends EventParams {
+  templateBase: namedTypes.File;
```
yuval-hazaz, line 127: this can be called `template`

yuval-hazaz (approved):
see one comment about naming, other than that it can be merged!

Issue #3932, opened 2022-09-20
🐛 Bug Report: Message broker - UI Issues


alexbass86:
### What happened?

In message broker, we have an empty state currently It's aligned to the top, and Please make it in the center of the container.
Also, we need to ensure our title and subtext are consistent and look precisely like it's in the design and other places in the system.

![image](https://user-images.githubusercontent.com/96179897/191235155-8efe860c-e126-419a-8638-a8ae0704284a.png)

### What you expected to happen

Empty state img should be in the center, and we need to have proper space between the title and subtext.
Please make sure to align it with the latest design found here.
https://www.figma.com/file/GLTbBuQoYvwas6XsCMHnqd/%23none---Resources?node-id=1877%3A11798

### How to reproduce

Go to the message broker.

### Amplication version

1.0.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

Azanul:
I'd like to work on this

yuval-hazaz:
thanks @Azanul 
I assigned the issue to you

yuval-hazaz:
The header is already fixed, so just the empty state position

Azanul:
@alexbass86 can I use sass arguement to alter empty state height?

Azanul:
@yuval-hazaz @alexbass86 I can't see the design in figma. Could you please post an image here.

Azanul:
@alexbass86  This is resolved

GreenMachine01:
@alexbass86 can I close this item?

barshimi:
this bug is already fixed

![Image](https://user-images.githubusercontent.com/4712526/206910166-1e02b68d-33c9-4c19-8838-85fc8107c268.png)

alexbass86:
Awosome

Azanul:
@alexbass86 You reopened it again 😆

PR #4141, opened 2022-10-12
FIX UI Issues #3932

Azanul:
Issue Number: #3932 

## PR Details
Added padding div to center the empty state in message broker page when the selected broker has no topics.

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
+++ b/packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
@@ -57,10 +57,12 @@ export default function TopicsList({
         )}
       />
     ) : (
-      <EmptyState
-        message="This message broker has no topics"
-        image={EnumImages.CommitEmptyState}
-      />
+      <div style={{ display:'flex', flexGrow:1, flexDirection:'column', alignItems:'center', height:'100%' }}>
+        <EmptyState
+          message="This message broker has no topics"
+          image={EnumImages.CommitEmptyState}
+        />
+      </div>
     )
   ) : null;
 }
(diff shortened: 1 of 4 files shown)


packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
```diff
@@ -57,10 +57,12 @@ export default function TopicsList({
...
-      <EmptyState
-        message="This message broker has no topics"
-        image={EnumImages.CommitEmptyState}
-      />
+      <div style={{ padding: '15.25%' }}>
+        <EmptyState
+          message="This message broker has no topics"
+          image={EnumImages.CommitEmptyState}
+        />
+      </div>
```
yuval-hazaz, line 65: it is better to set the height of the parent to 100% (depends on its parent) instead of using specific numbers. 
Also, the style should be in the SCSS file instead of inline

yuval-hazaz (changes requested):

packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
Azanul: height 100% adds a scrollbar to the side. I can add the style to scss file.
 The padding is in percentage. If you think this approach is not good then I can try something else.

yuval-hazaz (approved):

yuval-hazaz:
@Azanul thanks for your contribution
please have a look at this commit which adds "flex:1" to the container of the list and makes the empty state to be centered without using the specific percentage value 

https://github.com/amplication/amplication/pull/4159/commits/4b2a5b33757b6961f2f53b3734d2cf26f40ac8b2

Issue #4113, opened 2022-10-10
Docs: Plugin Events Article - CreateAdminModules


MoFoGo:
### Describe the issue with documentation

This event needs to be included in the docs.

### How to solve the problem

Create an article describing the Plugin Event. 

## The following can be used as examples for the documentation. 

https://github.com/amplication/amplication/issues/4095

https://github.com/amplication/amplication/issues/4094

## For reference, the Plugin Events tree is shown here:

https://www.notion.so/amplication/Plugins-Events-bbb45d7062c64c5ba5dcfba3fa0bcb4b

abrl91:
# Create Admin Modules

## Description

Creates the Admin UI

## Event Name:
`CreateAdminModule`

## Event Params
This event doesn't have unique event params

```tsx
export interface CreateAdminModulesParams extends EventParams {}
```

**Example**
There is no relevant example yet

abrl91:
@yuval-hazaz isn't it preferable to rename all the events that contain the suffix `modules` before writing the docs?

yuval-hazaz:
yes, we need to do that 
It will be very easy to update the docs, but it will be much more complicated to update the plugins as we move forward. 
so, let do it ASAP.  I  will open an issue

yuval-hazaz:
I created this issue https://github.com/amplication/amplication/issues/4158

Issue #4103, opened 2022-10-10
Docs: Plugin Events Article - CreateAuthModules


MoFoGo:
### Describe the issue with documentation

This event needs to be included in the docs.

### How to solve the problem

Create an article describing the Plugin Event. 

## The following can be used as examples for the documentation. 

https://github.com/amplication/amplication/issues/4095

https://github.com/amplication/amplication/issues/4094

## For reference, the Plugin Events tree is shown here:

https://www.notion.so/amplication/Plugins-Events-bbb45d7062c64c5ba5dcfba3fa0bcb4b

abrl91:
# Create Auth Module

## Description

Creates the Auth directory with the selected strategy (basic/JWT)

## Event Name:
`CreateAuthModules`

## Event Params

```ts
export interface CreateAuthModulesParams extends EventParams {
  srcDir: string;
}
```

### srcDir
The target directory the auth folder will be generated in

**Example**

```tsx
beforeCreateAuthModules(
    context: DsgContext,
    eventParams: CreateAuthModulesParams
  ) {
    context.utils.skipDefaultBehavior = true;
    return eventParams;
  }

  async afterCreateAuthModules(
    context: DsgContext,
    eventParams: CreateAuthModulesParams
  ) {
    const staticPath = resolve(__dirname, "../static");
    const staticsFiles = await context.utils.importStaticModules(
      staticPath,
      context.serverDirectories.srcDirectory
    );

    return staticsFiles;
  }
```

abrl91:
@barshimi @yuval-hazaz @MoFoGo Please review

yuval-hazaz:
@abrl91 I think we can add links to one or two of the existing auth plugins. 
we can also add some more details on the expected files that need to be included in the results (e.g. defaultAuth.guard.ts)

yuval-hazaz:
I am not sure about the last link - since we cannot ensure that it will be available in the future

abrl91:
> I am not sure about the last link - since we cannot ensure that it will be available in the future

OK. Screenshots?

yuval-hazaz:
> > I am not sure about the last link - since we cannot ensure that it will be available in the future
> 
> OK. Screenshots?

I think we can do without it

MoFoGo:
Transferred to docs

PR #4161, opened 2022-10-15
Fix/4158 remove the word module from plugin events

### LINKED ISSUE #4158: 🐛 Bug Report: remove the word `Module` from event names
### What happened?

We currently use the word Module for some events in a confusing way. 
In the event `CreateEntityModule` we actually create the NestJS module for the entity - so the name is relevant.

but, in the event `CreateAuthModules` the name is confusing, and we should not use the word Module in the event name

>note: the word Module is used in the DSG to describe a generated file - but we should not expose that to the events because it will be too confusing 

rename the following events:

- CreateAuthModules -> CreateServerAuth
- CreateAdminModules -> CreateAdminUI

Also note that the string value for these two events is written with a lower case "c" and it should be upper case like all other events.

When making these changes, we need to update the plugins that already using these events. 

```diff
export enum EventNames {
  CreateEntityController = "CreateEntityController",
  CreateEntityControllerBase = "CreateEntityControllerBase",
- CreateAuthModules = "createAuthModules",
- CreateAdminModules = "createAdminModules",
+ CreateServerAuth = "CreateServerAuth",
+ CreateAdminUI = "CreateAdminUI",
  CreateServer = "CreateServer",
  CreateServerAppModule = "CreateServerAppModule",
  CreateServerDotEnv = "CreateServerDotEnv",
  CreateMessageBroker = "CreateMessageBroker",
  CreateMessageBrokerTopicsEnum = "CreateMessageBrokerTopicsEnum",
  CreateMessageBrokerNestJSModule = "CreateMessageBrokerNestJSModule",
  CreateMessageBrokerClientOptionsFactory = "CreateMessageBrokerClientOptionsFactory",
  CreateMessageBrokerService = "CreateMessageBrokerService",
  CreateMessageBrokerServiceBase = "CreateMessageBrokerServiceBase",
  CreateEntityService = "CreateEntityService",
  CreateEntityServiceBase = "CreateEntityServiceBase",
  CreateServerDockerCompose = "CreateServerDockerCompose",
  CreateServerDockerComposeDB = "CreateServerDockerComposeDB",
  CreatePrismaSchema = "CreatePrismaSchema",
  CreateServerPackageJson = "CreateServerPackageJson",
  CreateEntityModule = "CreateEntityModule",
  CreateEntityModuleBase = "CreateEntityModuleBase",
}
```

### How to reproduce

n/a

### Amplication version

_No response_

### Environment

_No response_

_No response_

abrl91:
Close #4158 

## PR Details
remove the word `Module` from the plugin event name and event params 
```diff
- CreateAuthModules = "createAuthModules",
+ CreateServerAuth = "CreateServerAuth",
```

```diff
- CreateAdminModules = "createAdminModules",
+ CreateAdminUI = "CreateAdminUI",
```

```diff
-export interface CreateAdminModulesParams extends EventParams {}
+export interface CreateAdminUIParams extends EventParams {}
```

```diff
-export interface CreateAuthModulesParams extends EventParams {
-  srcDir: string;
-}
+export interface CreateServerAuthParams extends EventParams {
+ srcDir: string;
+}
```

## PR Checklist

### DIFF
--- a/packages/amplication-code-gen-types/src/plugin-events-params.ts
+++ b/packages/amplication-code-gen-types/src/plugin-events-params.ts
@@ -46,11 +46,11 @@ export interface CreateEntityControllerBaseParams extends EventParams {
   controllerBaseId: namedTypes.Identifier;
   serviceId: namedTypes.Identifier;
 }
-export interface CreateAuthModulesParams extends EventParams {
+export interface CreateServerAuthParams extends EventParams {
   srcDir: string;
 }
 
-export interface CreateAdminModulesParams extends EventParams {}
+export interface CreateAdminUIParams extends EventParams {}
 export interface CreateServerParams extends EventParams {}
 
 export type VariableDictionary = {
--- a/packages/amplication-code-gen-types/src/plugin-events.ts
+++ b/packages/amplication-code-gen-types/src/plugin-events.ts
@@ -1,6 +1,6 @@
 import {
-  CreateAdminModulesParams,
-  CreateAuthModulesParams,
+  CreateAdminUIParams,
+  CreateServerAuthParams,
   CreateEntityControllerBaseParams,
   CreateEntityControllerParams,
   CreateEntityServiceBaseParams,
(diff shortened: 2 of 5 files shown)


abrl91:
issues: https://github.com/amplication/amplication/issues/4103 and https://github.com/amplication/amplication/issues/4113
were updated based on the changes that will be applied to the corresponding plugins

abrl91:
a new version of the `code-gen-types` will be published at: https://github.com/amplication/amplication/pull/4131

yuval-hazaz (approved):

PR #4187, opened 2022-10-18
style: Fixed last commit details getting partially hidden

### LINKED ISSUE #4042: 🐛 Bug Report: Client - Last Commit panel is partially covered by the footer
### What happened?

The Last Commit panel is partially covered by the footer 

image

The last commit panel should not be covered by the footer 

### How to reproduce

This only happens on a smaller screen when the height is not big enough to include the empty state image (of the box). 
To fix that, we should use a media-query to limit the size of the empty state image on smaller screens (based on the height) or we can just use an image width and height of 120px instead of 150px 

### Amplication version

1.0.2

### Environment

_No response_

_No response_

TheLearneer:
Fixes #4042 

## PR Details

Fixed the styling issue where the last commit details were getting partially hidden

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/VersionControl/PendingChanges.scss
+++ b/packages/amplication-client/src/VersionControl/PendingChanges.scss
@@ -77,8 +77,9 @@ $resource-name-max-width: 180px;
     justify-content: flex-start;
     align-items: center;
 
-    img {
-      height: $empty-state-image-size;
+    .svg-theme-image {
+      width: 120px;
+      height: 120px;
     }
 
     &__title {
--- a/packages/amplication-client/src/VersionControl/PendingChanges.scss
+++ b/packages/amplication-client/src/VersionControl/PendingChanges.scss
@@ -78,8 +78,8 @@ $resource-name-max-width: 180px;
     align-items: center;
 
     .svg-theme-image {
-      width: 120px;
-      height: 120px;
+      width: $empty-state-image-size;
+      height: $empty-state-image-size;
     }
 
     &__title {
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/VersionControl/PendingChanges.scss
```diff
@@ -77,8 +77,9 @@ $resource-name-max-width: 180px;
     justify-content: flex-start;
     align-items: center;
 
-    img {
-      height: $empty-state-image-size;
+    .svg-theme-image {
+      width: 120px;
```
abrl91, line 81: you have a variable at the top of this file that its value is 120px
`$empty-state-image-size: 120px;`
please use it.

Why did you add the width?

TheLearneer: My bad, fixed the issue and used the variable

abrl91:
@TheLearneer 
I saw Yuval's description on the issue:

_"This only happens on a smaller screen when the height is not big enough to include the empty state image (of the box).
To fix that, we should use a media query to limit the size of the empty state image on smaller screens (based on the height), or we can just use an image width and height of 120px instead of 150px"_

As the image height was already 120px, can you please apply the change you have made to the `textarea` height with a media query? Meaning, that only on small screens the `textarea` height would be smaller?

TheLearneer:
Hi @abrl91 Updated the code to include media query for text area

abrl91:
run branch

abrl91 (approved):

PR #4195, opened 2022-10-19
Fix/4192 package json version

### LINKED ISSUE #4192: 🐛 Bug Report: Generated App - the `package.json` file is generated with a version value that is not aligned with the `npm server`
### What happened?

The `package.json` (server+admin) file is generated with a version value that is not aligned with the `npm semver`

image

image

`t1ngngd1` is not a valid npm version

I expect to see something that aligned with `npm semantic versions` for example: `0.0.1`

https://docs.npmjs.com/about-semantic-versioning

### How to reproduce

generate an app, clone/download and run the server
```
cd server
npm I
npm run docker:db
npm run prisma:generate <= at this command you will get the error
```

image

### Amplication version

next

### Environment

_No response_

abrl91:
Close #4192 

## PR Details
fixed the generated app `admin-ui/package.json` and `server/package.json`

also renamed the functions which create those files to match the event name that they trigger

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/admin/create-admin.ts
+++ b/packages/amplication-data-service-generator/src/admin/create-admin.ts
@@ -20,7 +20,7 @@ import { createRolesModule } from "./create-roles-module";
 import { createDotEnvModule } from "./create-dotenv";
 import pluginWrapper from "../plugin-wrapper";
 import DsgContext from "../dsg-context";
-import { createPackageJson } from "./package-json/create-package-json";
+import { createAdminUIPackageJson } from "./package-json/create-package-json";
 
 const STATIC_MODULES_PATH = path.join(__dirname, "static");
 const API_PATHNAME = "/api";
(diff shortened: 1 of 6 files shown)


abrl91:
I checked what happens in production and I saw that the version is not generated at all, what made me understand that in the current code we never get into this if statement: 
(`packages/amplication-data-service-generator/src/update-package-jsons.ts`)

```
if (!semver.valid(pkg.version)) {
    delete pkg.version;
  }
```

This happened because the `pkg` is json at this point. Therefore I moved this check to `packages/amplication-data-service-generator/src/util/preparePackageJsonFile.ts` where we have the parsed version of the `package.json` file. It also makes more sense to validate the version where we "prepare" the file.


@yuval-hazaz @GreenMachine01 
This fix resolved the issue (checked locally). The result is a `package.json` file without a version, as we have in production.
BUT, I think we still have to check why `appInfo.version` returns the `swagger` version

```
const packageJsonModule = await createServerPackageJson({
    updateProperties: [
      {
        name: `@${paramCase(appInfo.name)}/server`,
        version: appInfo.version,
      },
    ],
  });
```

EugeneTseitlin (approved):

PR #4219, opened 2022-10-21
fix: Update pending changes not refreshing on new project

### LINKED ISSUE #4043: 🐛 Bug Report: Pending changes list is not updated when switching projects
### What happened?

When you view a project with pending changes, then switch the another project without pending changes, the list of pending changes is still showing the items from the previous project

in this screenshot, we can see changes on a service called "my service-1" while the current project does not include this service 
image

The list of pending changes should be updated and show the changes of the current project, even when there are no changes in the newly selected project 

### How to reproduce

Open a project and make some changes (for example, created a service).
Before committing the changes, go to the Project List page, and select a new project. 

assuming there are no pending changes in the newly selected project, the list of pending changes still shows the items from the previous project. 

make a change in the new project (example, create a new service), and see how the list is now updated with the relevant changes. 

### Amplication version

1.0.2

### Environment

_No response_

_No response_

TheLearneer:
Fixes #4043 

## PR Details
The issue has been fixed where pending changes of older project were retained in a newly started project with no pending changes.
The issues seemed to be the check being done in the `usePendingChanges` hook.

Currently the changes were not updated if the changes length was zero. This was causing the issue.

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/hooks/usePendingChanges.ts
+++ b/packages/amplication-client/src/Workspaces/hooks/usePendingChanges.ts
@@ -28,8 +28,7 @@ const usePendingChanges = (currentProject: models.Project | undefined) => {
   });
 
   useEffect(() => {
-    if (!pendingChangesData || !pendingChangesData.pendingChanges.length)
-      return;
+    if (!pendingChangesData) return;
 
     setPendingChanges(pendingChangesData.pendingChanges);
     setPendingChangesMap(


packages/amplication-client/src/Workspaces/hooks/usePendingChanges.ts
```diff
@@ -28,8 +28,7 @@ const usePendingChanges = (currentProject: models.Project | undefined) => {
   });
 
   useEffect(() => {
-    if (!pendingChangesData || !pendingChangesData.pendingChanges.length)
-      return;
+    if (!pendingChangesData) return;
```
barshimi, line 31: instead of changing this condition and rerender the component, i suggest we add another useEffect that check the changes of currentProject (with the url project param) and accordingly refetch pending changes

TheLearneer: Sounds good to me. I am on it 🙌

TheLearneer: @barshimi I went through the logic once again, and I am not sure if refetching from another userEffect will be enough.
Because this specific block of code is responsible for setting the changes in the state.

Even if we refetch pending changes elsewhere we might still have to write the same code about setting state again, isn' it?

yuval-hazaz:
@barshimi are you handling this one?

tupe12334 (approved):

tupe12334:
@all-contributors please add @TheLearneer  for code

yuval-hazaz:
@TheLearneer good catch! Well done, and thanks for contributing to Amplication 🚀

PR #4222, opened 2022-10-21
Strip wrong Request Body properties with ValidationPipe whitelist true

### LINKED ISSUE #4221: 🐛 Bug Report: Wrong Incoming Body parameters are not stripped and throw error (global ValidationPipe is missing  whitelist: true)
### What happened?

Wrong Incoming Body parameters are not stripped and throw internal server error caused by prisma not accepting wrong parameters

```
{
  "definedParamInDTO": "value"
  "wrongParam": ""
}
```

Is there a reason to not use the recommended  global ValidationPipe with whitelist: true?

```
  app.useGlobalPipes(
    new ValidationPipe({
      transform: true,
      whitelist: true.  // <- this option
    })
  );
```

Wrong parameters in POST body objects should be stripped and the process not stopped 

### How to reproduce

Add parameter to POST or PATCH body, which is not defined in the incoming DTO

```
{
  "definedParamInDTO": "value"
  "wrongParam": ""
}
```

### Amplication version

1.0.2

### Environment

_No response_

vincenzodomina:
Close #4221 

## PR Details

Wrong Incoming Body parameters are not stripped and throw internal server error caused by prisma not accepting wrong parameters

Wrong parameters in POST body objects should be stripped and the process not stopped

whitelist: true strips down wrong properties and lets the process continue with prisma not throwing an error.
[NEstJS docs](https://docs.nestjs.com/techniques/validation)

## PR Checklist


souravjain540:
@arielweinberger any updates?

arielweinberger:
I am happy to accept this change. Thanks for your contribution!

yuval-hazaz:
According to the docs here https://docs.nestjs.com/techniques/validation

"If set to true, validator will strip validated (returned) object of any properties that do not use any validation decorators."

@arielweinberger, I am wondering what will happen to valid properties that do not have any validation decorator? And do we have any such properties in the generated code?

yuval-hazaz:
@arielweinberger did you find the time to take a loot at this one?

vincenzodomina:
> According to the docs here https://docs.nestjs.com/techniques/validation
> 
> "If set to true, validator will strip validated (returned) object of any properties that do not use any validation decorators."
> 
> @arielweinberger, I am wondering what will happen to valid properties that do not have any validation decorator? And do we have any such properties in the generated code?

Isn't the validaton decorators in the DTO's our point of control to define what is valid? Then there shouldn't be any "valid" properties without validation decorators. Or how do you define "valid" at this point?

yuval-hazaz:
> Isn't the validaton decorators in the DTO's our point of control to define what is valid? Then there shouldn't be any "valid" properties without validation decorators. Or how do you define "valid" at this point?

There are different types of properties which not necessarily use a validation decorator - see this one, for example: 

```typeScript
@ApiProperty({
    required: false,
    type: [PostOrderByInput],
  })
  @Field(() => [PostOrderByInput], { nullable: true })
  @Type(() => PostOrderByInput)
  orderBy?: Array<PostOrderByInput>;
```


in case you can test your solution in a local environment with this property (in a FindManyArgs object doing a search query), please share your finding

yuval-hazaz:
@vincenzodomina we have decided not to use this method since it will negatively impact some valid use cases.

Thank you again for your contribution. 
I will add comments on the issue for the next steps

PR #4227, opened 2022-10-22
fix:UI issue - Select field's input text is unreadable with current t…

### LINKED ISSUE #4001: 🐛 Bug Report: UI issue - Select field's input text is unreadable with current text color
### What happened?

Screen Shot 2022-10-01 at 12 07 54 PM

The text color for the select field is currently barely readable when I tried to search for a datatype, I have to guess what I'm typing most of the type when I try to search for a specific data type.

A more readable text color.

### How to reproduce

None

### Amplication version

1.0.1

### Environment

_No response_

KushalRaut:
Fixes #4001 

## PR Details

There was an issue where the search text was hardly visible due to it's dark color
image
The issue has now been fixed by making the color of the input field white.
image

## PR Checklist

### DIFF
--- a/packages/amplication-design-system/src/components/SelectField/SelectField.scss
+++ b/packages/amplication-design-system/src/components/SelectField/SelectField.scss
@@ -13,6 +13,10 @@
 
   .select-field__container {
     font-size: var(--normal-font-size);
+
+    input {
+      color: white !important;
+    }
   }
   /**@todo: convert all class selector for component child elements to "& &__child" */
   .select-field__control {


packages/amplication-design-system/src/components/SelectField/SelectField.scss
```diff
@@ -13,6 +13,10 @@
 
   .select-field__container {
     font-size: var(--normal-font-size);
+
+    input {
+      color: white !important;
```
abrl91, line 18: Why did you use `important`?

KushalRaut: By default, the input field ``` inherits``` the text-color  as its default element style, and hence the ```important``` was required to make sure the select field input has ```color: white```

arielweinberger: Can you overcome the need of `!important` by being more specific, rather then using a tag selector (`input {}`)?

souravjain540: @KushalRaut any updates?

KushalRaut: I tried many ways to make sure that the input field text becomes white but most of them didn't work and the ```!important``` tag is necessary because the input field has inline styling of ```color: inherit```

PR #4238, opened 2022-10-25
fix/4226-Discord Button in Readme 

weblearner07:
I have wrote a note to how to open Discord invite link in new tab without closing Github , As markdown doesn't support opening link in new tab with Anchor tag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Amplication is currently in its version 1.  This is the first major release of A
 
 Ask your questions and participate in discussions on Amplication-related and web-dev topics at the Amplication Discord channel. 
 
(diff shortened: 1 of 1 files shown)


weblearner07:
Can Anyone review my pr

souravjain540:
I don't think this is something I want to be in readme. If there is a way in which the user clicks and it opens in a new tab, then I can approve this PR, but adding text is not looking good to me.

weblearner07:
Yeah,I know you needed to open it in new tab but it is not possible i guess with readme.md , even if i add target="_blank" it won't redirect to new tab. So Ctrl + click is only the option., Correct me if I am Wrong ....

PR #4279, opened 2022-10-29
Feat/4234 Change create one to create many route

### LINKED ISSUE #4234: Create many - Route
### Feature description

Additionally to the "Create one" Route I suggest to have an/ or change it to an "Create many" Route as the standard POST Route in the REST-API

### Use case

Problem:
Instead of having to loop on the client side to create many records from an array of items and thus having many http-calls that are slow, the standard POST Route could be a "Create Many" Route

- Prisma has an createMany function already built in
- In the Request body an array with only one item can be used for creating only one record (if Route changed)

The feature could look like this in the UI (like on the Permissions page) i can select between an "Create one" or "Create many" Route for a given Entity:
image

vincenzodomina:
Close #4234

## PR Details

This PR changes the "Create one" to an "Create many" Route as the standard POST Route in the REST-API.
For this I changed the controller template and the data service generator to expect an array of entity objects instead of just one object.

Unfortunately the [ Prisma "createMany" method](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#nested-writes) cannot handle relations in the way "create" can so i loop through the Array of Input objects in the controller and use the Prisma "create" method.

Problem to solve:
Instead of having to loop on the client side to create many records from an array of items and thus having many http-calls that are slow or even fail with an ERR_INSUFFICIENT_RESOURCES in the browser, the standard POST Route could be a "Create Many" Route. In the Request body an array with only one item can be used for creating only one record.

Test snapshots where updated using `npm test -- -u`

Caveats:
- Does this change apply also for GraphQL automatically? (I could not figure this out on the fly)
- This change would be a breaking change for all existing projects as it changes the expected Input for the POST route (making another PR with adding this behaviour as an optional setting in the Amplication UI could solve this)

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/server/resource/controller/controller.base.template.ts
+++ b/packages/amplication-data-service-generator/src/server/resource/controller/controller.base.template.ts
@@ -66,11 +66,11 @@ export class CONTROLLER_BASE {
   })
   @common.Post()
   @swagger.ApiBody({ type: [CREATE_INPUT] })
-  @swagger.ApiCreatedResponse({ type: ENTITY })
+  @swagger.ApiCreatedResponse({ type: [ENTITY] })
   @swagger.ApiForbiddenResponse({ type: errors.ForbiddenException })
   async CREATE_ENTITY_FUNCTION(
     @common.Body() data: [CREATE_INPUT]
-  ): Promise<ENTITY> {
+  ): Promise<ENTITY[]> {
     try {
       return await Promise.all(data.map(async _data => {
         await this.service.create({
--- a/packages/amplication-data-service-generator/src/server/resource/controller/controller.base.template.ts
+++ b/packages/amplication-data-service-generator/src/server/resource/controller/controller.base.template.ts
@@ -12,20 +12,20 @@ import { plainToClass } from "class-transformer";
 // @ts-ignore
 import { ApiNestedQuery } from "../../decorators/api-nested-query.decorator";
 
-declare interface CREATE_INPUT {}
-declare interface WHERE_INPUT {}
-declare interface WHERE_UNIQUE_INPUT {}
+declare interface CREATE_INPUT { }
+declare interface WHERE_INPUT { }
+declare interface WHERE_UNIQUE_INPUT { }
 declare class FIND_MANY_ARGS {
   where: WHERE_INPUT;
 }
-declare interface UPDATE_INPUT {}
+declare interface UPDATE_INPUT { }
 
 declare const FINE_ONE_PATH: string;
 declare const UPDATE_PATH: string;
 declare const DELETE_PATH: string;
 
-declare class ENTITY {}
-declare interface Select {}
+declare class ENTITY { }
+declare interface Select { }
 
 declare interface SERVICE {
   create(args: { data: CREATE_INPUT; select: Select }): Promise<ENTITY>;
@@ -55,7 +55,7 @@ export class CONTROLLER_BASE {
   constructor(
     protected readonly service: SERVICE,
     protected readonly rolesBuilder: nestAccessControl.RolesBuilder
-  ) {}
+  ) { }
 
   // @ts-ignore
   @common.UseInterceptors(AclValidateRequestInterceptor)
@@ -73,12 +73,12 @@ export class CONTROLLER_BASE {
   ): Promise<ENTITY[]> {
     try {
       return await Promise.all(data.map(async _data => {
-        await this.service.create({
-          data: CREATE_DATA_MAPPING, // TODO: replace one with many
+        return await this.service.create({
+          data: CREATE_DATA_MAPPING,
           select: SELECT,
         });
       }));
-    } catch (error) {      
+    } catch (error) {
       throw error;
     }
   }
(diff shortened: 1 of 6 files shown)


abrl91:
Hey @vincenzodomina 
Thank you for submitting this pull request.
I reviewed your code briefly (I will look at it deeply on Monday), though I have some notes:

- I realized that this kind of change requires a change in the React Admin (can be in a different pull request). 
- I read this: [Creates all records in one query](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#create-multiple-records-and-multiple-related-records), and I realized that we had this discussion a while ago. Sorry, I forgot about it (I saw you asked about it in the issue).
- I read this chart:

image

I noticed that you changed the `create` method. Logically,  I understand why you did it, but I think it is preferable to give the user the two options and add other criteria in the UI for "Create Many". 

image

Why?  with the current `create` method we support nested relations in one-to-many **and** many-to-many, and we don't want to lose this feature.

abrl91:
"_Does this change apply also for GraphQL automatically? (I could not figure this out on the fly)_"

You changed only the `create-controller` file: https://github.com/amplication/amplication/pull/4279/files#diff-e54125511d21526f9b44870b8eea6e30024007945e0067d2717be302683d7dc7

You have to change the `create-resolver` (`packages/amplication-data-service-generator/src/server/resource/resolver/create-resolver.ts`)

vincenzodomina:
> Why? with the current create method we support nested relations in one-to-many and many-to-many, and we don't want to lose this feature.

This is why I did not use the createMany method from Prisma, instead looping trough the Request array with the same create method. So this should not be affecting the relation features? I am using it with relations in my application.

arielweinberger:
@vincenzodomina Thanks for this PR! With all that's going on in this week, we probably won't get to look at this PR in the upcoming few days. 

I'd still love to reward you for this contribution so I've added the Hacktoberfest label.

Thanks for your many contributions, it's noticeable and super helpful. Our community manager @souravjain540 will contact you in private, I'd love to schedule a 1:1 with you and get to know you as one of our most noticeable contributors this Hacktoberfest.

vincenzodomina:
> @vincenzodomina Thanks for this PR! With all that's going on in this week, we probably won't get to look at this PR in the upcoming few days.
> 
> I'd still love to reward you for this contribution so I've added the Hacktoberfest label.
> 
> Thanks for your many contributions, it's noticeable and super helpful. Our community manager @souravjain540 will contact you in private, I'd love to schedule a 1:1 with you and get to know you as one of our most noticeable contributors this Hacktoberfest.

Oh wow, thank you so much!

tupe12334:
@yuval-hazaz @GreenMachine01 @arielweinberger What do you think about the PR, this is a behavior change and we need to make a call about that, either way, @vincenzodomina Your PRs are impressive, thanks a lot.

tupe12334:
@all-contributors please add @vincenzodomina  for code

yuval-hazaz:
@tupe12334 @abrl91 @overbit @arielweinberger let's have a talk about this PR today and decide how we want to proceed. 
I see 3 options:
1. Accept the PR as is and change the spec for all generated apps (breaking change for existing apps)
2. do not accept the PR and leave the current state as it is. 
3. Create a plugin with these changes so users can choose to use the plugin when the behavior is needed. 

All of the above is suggested assuming the code is ready and final. I see that @vincenzodomina mentioned that there is still work needed on the GraphQL and Admin UI.

yuval-hazaz:
@vincenzodomina, after additional consideration and research, we've decided that bulk creation should be implemented with a separate endpoint. 

The restful standard talks about the POST for creating a single resource. 
There is a de facto standard of using /resource/bulk to create multiple resources. 

See this for example https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-products

Implementing it this way will also not break existing apps and will give the developer more control over the generated code for edge use cases. 

So, what I suggest is:
1. Create a new endpoint on the controller for /bulk
2. Move the iteration logic into the entity service and expose a single function CreateBulk(Entity[])
3. We also need to return detailed error codes for each resource in case of failure

If you want to implement that, we would be happy to help with it. 
If you prefer not to, we will create a new issue on the backlog and will implement it in one of the future releases.

Note: 
This behavior should also be implemented in GraphQL with a mutation like "bulkCreateCustomer"

vincenzodomina:
> @vincenzodomina, after additional consideration and research, we've decided that bulk creation should be implemented with a separate endpoint.
> 
> The restful standard talks about the POST for creating a single resource. There is a de facto standard of using /resource/bulk to create multiple resources.
> 
> See this for example https://woocommerce.github.io/woocommerce-rest-api-docs/#batch-update-products
> 
> Implementing it this way will also not break existing apps and will give the developer more control over the generated code for edge use cases.
> 
> So, what I suggest is:
> 
> 1. Create a new endpoint on the controller for /bulk
> 2. Move the iteration logic into the entity service and expose a single function CreateBulk(Entity[])
> 3. We also need to return detailed error codes for each resource in case of failure
> 
> If you want to implement that, we would be happy to help with it. If you prefer not to, we will create a new issue on the backlog and will implement it in one of the future releases.
> 
> Note: This behavior should also be implemented in GraphQL with a mutation like "bulkCreateCustomer"

Thanks a lot! Your proposal looks good! Following existing standards is of course a good choice.

Unfortunately I am a bit time constraint right now and would leave it to your backlog. Will you open this feature to the community?

yuval-hazaz:
@GreenMachine01 please add to our backlog. You can link to this PR for reference
I will close this PR

PR #4308, opened 2022-11-02
added `CreateSeed` event

### LINKED ISSUE #4292: Add "createSeed" event to DSG

abrl91:
Close #4292 

# Create Seed

## Description

Creates the seed file. By default, this seed file contains only the seeding of the User entity.

## Event Name:
`CreateSeed`

## Event Params

```ts
export interface CreateSeedParams extends EventParams {
  template: namedTypes.File;
  templateMapping: { [key: string]: any };
  fileDir: string;
  outputFileName: string;
}
```

### template
A template file that is used to generate the service
This is the default template that is used for this event.

You can manipulate the template or replace it completely with a new template in your plugin. 

### templateMapping
An object with values that are available in the interpolation process of the template.

You can find the available properties here:
https://github.com/amplication/amplication/blob/feat/4292-create-seed/packages/amplication-data-service-generator/src/server/seed/create-seed.ts#L86

You can manipulate the object by adding new values, or replacing existing values that will be used in the template when creating the seed file.

### fileDir
The target directory the seed file will be generated in

### outputFileName
The file name of the seed file, the default value is `seed.ts`

It is recommended not to change the file name unless specifically required and the impact is understood.

---

## PR Checklist


packages/amplication-data-service-generator/src/server/seed/create-seed.ts
```diff
@@ -67,34 +70,55 @@ export const DEFAULT_AUTH_PROPERTIES = [
...
-  interpolate(file, {
-    DATA: builders.objectExpression([
-      ...DEFAULT_AUTH_PROPERTIES,
-      ...customProperties,
-    ]),
+  const template = await readFile(seedTemplatePath);
+
+  return pluginWrapper(createSeedInternal, EventNames.CreateSeed, {
+    template,
+    seedingProperties: [...DEFAULT_AUTH_PROPERTIES, ...customProperties],
```
yuval-hazaz, line 86: Why are we using here a different format instead of using "templateMapping" like in all other events?

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/server/seed/create-seed.ts
abrl91: The reason I decided to implement it that way is that I distinguished three things:
1. in this case the template mapping has a specific key: `DATA`
2. the value's type is not any, but `namedTypes.ObjectProperty[]`
3. I think that only with this default and specific implementation of seeding the User entity, we use the `DEFAULT_AUTH_PROPERTIES`. A plugin that makes use of this event should only send the properties that are needed to seed an entity. 
```
const templateMapping = {
    DATA: builders.objectExpression(seedingProperties),
  };
```

yuval-hazaz: in all templateMapping we use specific keys - in this case, we only use one key named "DATA".
In case I want to create a new template I should be able to add or update the keys on the object - so, I don't see a reason to use something else

abrl91: "_in all templateMapping we use specific keys - in this case, we only use one key named "DATA"._"
@yuval-hazaz right! 😅 I missed it

abrl91:
@MoFoGo you can use the pull request description for the docs

yuval-hazaz (approved):

packages/amplication-data-service-generator/src/server/seed/create-seed.ts
MoFoGo: > "_in all templateMapping we use specific keys - in this case, we only use one key named "DATA"._" @yuval-hazaz right! 😅 I missed it

@abrl91 - please explain how should this information be incorporated in the text

PR #4321, opened 2022-11-06
Migrate amplication client and amplication design system to Nx

barshimi:
Close #[issue-number]

## PR Checklist


packages/amplication-client/.env
```diff
@@ -10,3 +10,4 @@ REACT_APP_FEATURE_FLAGS="{}"
 REACT_APP_DATA_SOURCE="/graphql"
 REACT_APP_STORAGE_GATEWAY_URL="http://localhost:3002"
 REACT_APP_SERVER_URI=""
+SKIP_PREFLIGHT_CHECK=true
```
tupe12334, line 13:
```suggestion
```

packages/amplication-client/config-overrides.js
```diff
@@ -19,6 +19,7 @@ module.exports = function override(config, env) {
   for (const peerDependency of Object.keys(designSystemPkg.peerDependencies)) {
     aliasDependencyToTopLevel(config, peerDependency);
   }
+
```
tupe12334, line 22: We need to remove this file and react rewrite

packages/amplication-client/src/Resource/git/dialogs/GitDialogsContainer.tsx
```diff
@@ -1,4 +1,4 @@
-import { Resource } from "@amplication/code-gen-types/dist/models";
+import { Resource } from "@amplication/code-gen-types//models";
```
tupe12334, line 1:
```suggestion
import { Resource } from "@amplication/code-gen-types/models";
```

packages/amplication-client/src/Resource/git/dialogs/GitCreateRepo/GitCreateRepo.tsx
```diff
@@ -1,4 +1,4 @@
-import { Resource } from "@amplication/code-gen-types/dist/models";
+import { Resource } from "@amplication/code-gen-types//models";
```
tupe12334, line 1:
```suggestion
import { Resource } from "@amplication/code-gen-types/models";
```

packages/amplication-design-system/package.json
```diff
@@ -1,14 +1,13 @@
 {
   "name": "@amplication/design-system",
-  "version": "1.0.3",
+  "version": "0.0.1",
```
tupe12334, line 3: Why we need to downgrade this package?

packages/amplication-design-system/package.json
```diff
@@ -79,9 +77,12 @@
       "no-duplicate-imports": "error"
     }
   },
+  "resolutions": {
+    "@types/react": "^17.0.2"
```
tupe12334, line 81: Do we need that if our @types/react in that version already

nx.json
```diff
@@ -20,8 +20,9 @@
           "prebuild",
           "build",
           "prepare",
+          "check-format",
+          "format",
           "test",
-          "test:update-snapshot",
```
tupe12334, line 24: Why do we toke that down

tupe12334 (changes requested):

packages/amplication-client/Dockerfile
```diff
@@ -1,34 +1,18 @@
...
 WORKDIR /app
 
-COPY --from=build /app/packages/amplication-client/nginx.conf /etc/nginx/conf.d/default.conf
-COPY --from=build /app/packages/amplication-client/build /app
-COPY --from=build /app/packages/amplication-client/src/scripts/inject-variables.js .
-COPY --from=build /app/packages/amplication-client/entrypoint.sh .
+COPY ./packages/amplication-client/nginx.conf /etc/nginx/conf.d/default.conf
+COPY ./dist/apps/@amplication/client .
+COPY ./packages/amplication-client/src/scripts/inject-variables.js .
+COPY ./packages/amplication-client/entrypoint.sh .
```
overbit, line 10: Ideally we should copy any necessary files during the build process instead of ad-hoc operations in the Dockerfile

packages/amplication-client/package.json
```diff
@@ -48,19 +44,19 @@
...
-    "yup": "0.32.11"
+    "yaml": "1.10.2"
   },
   "scripts": {
-    "start": "react-app-rewired start",
+    "start": "craco start",
     "start:debug": "npm start",
-    "build": "react-app-rewired build",
-    "test": "react-app-rewired test",
+    "build": "CI=false BUILD_PATH=../../dist/apps/@amplication/client craco build",
```
overbit, line 52: `CI=false` will have precedence over any other ver definition of `CI`.
Is this on purpose? 

```suggestion
    "build": "BUILD_PATH=../../dist/apps/@amplication/client craco build",
```

overbit (changes requested):

packages/amplication-client/Dockerfile
arielweinberger: Had a chat with @overbit and explained to him that this could be resolved using a build target in Nx, and the `assets` property. Unfortunately since we use craco here, I think it's not worth the investment. Resolving this.

packages/amplication-client/package.json
arielweinberger: @barshimi I think we should just get rid of these warnings. These are very trivial to fix and maybe this can even automatically fixed.

Anyway, let's tackle it after the deployment.

PR #4348, opened 2022-11-11
Fix: env variables for client app

overbit:
## PR Details

Fix leftover environment variable changes for client application after migration nx react.
 

## PR Checklist

### DIFF
--- a/helm/charts/amplication/values-alex.yaml
+++ b/helm/charts/amplication/values-alex.yaml
@@ -21,10 +21,10 @@ amplication-client:
     hostname: app-alex.amplication-dev.com
   config:
     env:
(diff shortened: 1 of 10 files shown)


arielweinberger (approved):

yuval-hazaz:
@overbit out of interest, what is the reason to change the variable names? Why do we need to tie the monorepo tool that builds the client to the actual code of the client?

overbit:
> @overbit out of interest, what is the reason to change the variable names? Why do we need to tie the monorepo tool that builds the client to the actual code of the client? 

It's due to the nx react plugin. That will automatically resolve variables in the build process is they are prefixed with `NX_` 
https://nx.dev/recipes/environment-variables/use-environment-variables-in-react

This change was to match the changes done as part of PR #4346

Issue #4319, opened 2022-11-06
When deleting an entity show "are you sure" notification if entity is linked to other entities


GreenMachine01:
**Overview:**
it is possible for a service to have 2 entities linked to each other. 
![image](https://user-images.githubusercontent.com/112330500/200163838-145e1d4c-9791-4db7-8e00-09ba504bc157.png)
if we delete one of the entities we will get an error when trying to build and commit.
The source of the error is that the entity that remains still has a relation to the entity that has been deleted.

**Requirements:**
1. When deleting an entity that has relations to other entities show the user an "Are you sure?" pop-up with relevant information.
2. if the user decides to cancel the entity deletion close the "are you sure" popup - no additional action is needed
3. if the user decides to continue with the entity deletion all fields that have a relation to the entity being deleted should be automatically deleted from the remaining entities

**Text Instructions for "Are you sure":**
Title - "Delete '[entity name]'"

Body - "Are you sure you want to delete this entity?
             This will also delete the related fields: 
             [field name] ('[entity name]')"

![Image](https://user-images.githubusercontent.com/112330500/202900341-6c9a64de-4bf1-4b7a-b0b1-9e28765d217e.png)

https://www.figma.com/file/wBcRAckCdHNCT3zUIOlK59/%234319---Entities%E2%80%A8---Are-you-sure%3F?node-id=1136%3A6535&t=nV5vuxsWj10YoT6k-1

tupe12334:
UPDATE:
As the one who takes this task, I split it into 3 sub-tasks as mentioned in the requirements section.
I will open for every one of them a separated PR.

tupe12334:
Hi @GreenMachine01 please contact @alexbass86 and let me know when there is a final design for the alert structure.
The data logic is already finished.

GreenMachine01:
@tupe12334 this seems wrong...


![Image](https://user-images.githubusercontent.com/112330500/205862757-22fdbd8e-764b-43e3-9765-92b93a238cde.png)

tupe12334:
You are right, I missed the whit text color, I regard the border; we agreed that this will be a separated task.

PR #4367, opened 2022-11-15
Connected field deletion alert

tupe12334:
## PR Details

This PR is discovered by accident in the working of the #4319 solution.
This PR adds an alert that lets the user know when they delete a field with a relation.
Screen Shot 2022-11-15 at 15 39 49


packages/amplication-client/src/Entity/DeleteEntityField.tsx
```diff
@@ -123,3 +138,27 @@ const DELETE_ENTITY_FIELD = gql`
     }
   }
 `;
+
+const GET_ENTITY_WITH_SPECIFIC_FIELDS = gql`
```
yuval-hazaz, line 142: I think this should be "FIELD" and not "FIELDS"

packages/amplication-client/src/Entity/DeleteEntityField.tsx
```diff
@@ -123,3 +138,27 @@ const DELETE_ENTITY_FIELD = gql`
...
+  query entity($entityId: String!, $fieldId: String!) {
+    entity(where: { id: $entityId }) {
+      id
+      name
+      displayName
+      pluralDisplayName
+      fields(where: { permanentId: { equals: $fieldId } }) {
```
yuval-hazaz, line 149: are you sure you are using the correct field for fetch? we usually use "id" and not "permanentId" which is a specific type of id that follows the field over versions

packages/amplication-client/src/Entity/DeleteEntityField.tsx
```diff
@@ -30,8 +30,17 @@ export const DeleteEntityField = ({
   onDelete,
   onError,
 }: Props) => {
+  const { dataType, properties } = entityField;
+  const { relatedFieldId, relatedEntityId } = properties;
   const [confirmDelete, setConfirmDelete] = useState<boolean>(false);
   const { addEntity } = useContext(AppContext);
+  const { data } = useQuery(GET_ENTITY_WITH_SPECIFIC_FIELDS, {
```
yuval-hazaz, line 37: better use a more specific name
```suggestion
  const { data : relatedFieldData } = useQuery(GET_ENTITY_WITH_SPECIFIC_FIELDS, {
```

yuval-hazaz (changes requested):
great work, see some minor comments

yuval-hazaz:
@tupe12334 please review the comments and fix so we can merge

PR #4379, opened 2022-11-16
Add an option to set Incremental Id or UUID or CUID (the default) 

### LINKED ISSUE #1342: Data Service Generator: support incremental id and uuid
Add support to an incremental Id on the id field of an entity
`id          Int           @id @default(autoincrement())`

abrl91:
Close #1342 

## PR Details
- client: add a select option with: `uuid`, `cuid` and autoincrement
- code-gen-types (amplication data): add `idType` as a property to `id.json`
- DSG: 
   -  make some tweaks to create-prisma-schema in order to get the right types for an entity and related entity fields
   - [make user-info and iTokenPayload dynamic instead of static](https://github.com/amplication/amplication/pull/4379/commits/89162199a486473bd81914ae3618a5ebca480006)
   - tests: added the `idType` property to the User and Profile entity and update snapshots

## PR Checklist


packages/amplication-client/src/Entity/EntityField.tsx
```diff
@@ -187,7 +187,8 @@ const EntityField = () => {
             )}
           </div>
           <EntityFieldForm
-            isDisabled={
+            isDisabled={false}
```
yuval-hazaz, line 190: Do we still need this prop?

packages/amplication-client/src/Entity/EntityFieldForm.tsx
```diff
@@ -36,6 +36,7 @@ type Props = {
   defaultValues?: Partial<models.EntityField>;
   resourceId: string;
   entityDisplayName: string;
+  isSystemData?: boolean;
```
yuval-hazaz, line 39:
```suggestion
  isSystemDataType?: boolean;
```

packages/amplication-client/src/Entity/EntityFieldForm.tsx
```diff
@@ -129,38 +131,45 @@ const EntityFieldForm = ({
             <DisplayNameField
               name="displayName"
               label="Display Name"
-              disabled={isDisabled}
+              disabled={isDisabled || isSystemData}
               required
             />
             <NameField name="name" disabled={isDisabled} required />
```
yuval-hazaz, line 137: so, for system data types we cannot change the dispaly name, but we can change the field name?

packages/amplication-client/src/Entity/EntityFieldForm.tsx
```diff
@@ -129,38 +131,45 @@ const EntityFieldForm = ({
...
-            <div>
-              <ToggleField
-                name="searchable"
-                label="Searchable"
-                disabled={isDisabled}
-              />
-            </div>
-            {!SYSTEM_DATA_TYPES.has(formik.values.dataType) && (
-              <DataTypeSelectField label="Data Type" disabled={isDisabled} />
+            {schema.title !== "Id" && (
```
yuval-hazaz, line 143: Why did we decide to hide all these fields when editing an ID? 
If I understand correctly, before this change, all these fields are visible but disabled, and now we decided to remove them completely.

packages/amplication-client/src/Entity/SchemaField.tsx
```diff
@@ -7,21 +6,24 @@ import RelatedEntityFieldField from "./RelatedEntityFieldField";
...
   resourceId,
   entityDisplayName,
-}: {
-  propertyName: string;
-  propertySchema: Schema;
-  isDisabled?: boolean;
-  resourceId: string;
-  entityDisplayName: string;
-}) => {
+  isSystemData,
```
yuval-hazaz, line 25: this prop is not being used

packages/amplication-client/src/Entity/SchemaField.tsx
```diff
@@ -63,22 +65,28 @@ export const SchemaField = ({
...
+              <OptionSet
+                label={label}
+                name={fieldName}
+                isDisabled={isDisabled}
+              />
+            );
+          }
+          default: {
+            throw new Error(
+              `Unexpected propertySchema.items.type: ${propertySchema.type}`
```
yuval-hazaz, line 84: Is this change related to the current PR? If not, I suggest removing it. 
If it is needed, please open a new issue and make the change separately.

packages/amplication-client/src/Entity/SchemaFields.tsx
```diff
@@ -23,24 +22,26 @@ export const SchemaFields = ({
...
-        return (
-          <div key={name}>
+  if (schema?.properties) {
+    return (
+      <div>
+        {Object.entries(schema.properties).map(([name, property]) => {
+          if (!property) {
+            throw new Error(`Missing property: ${name}`);
+          }
+          return (
```
yuval-hazaz, line 32: Is this change related to the current PR? If not, I suggest removing it.
If it is needed, please open a new issue and make the change separately.

packages/amplication-data-service-generator/src/server/static/src/auth/ITokenService.ts
```diff
@@ -1,5 +1,5 @@
 export interface ITokenPayload {
-  id: string;
+  id: string | number;
```
yuval-hazaz, line 2: This should be generated dynamically by the type of the ID field on the User entity (either a string or a number)

packages/amplication-data-service-generator/src/server/static/src/auth/UserInfo.ts
```diff
@@ -5,8 +5,8 @@ import { User } from "../user/user";
 
 @ObjectType()
 export class UserInfo implements Partial<User> {
-  @Field(() => String)
-  id!: string;
+  @Field(() => String || Number)
+  id!: string | number;
```
yuval-hazaz, line 9: This should be generated dynamically by the type of the ID field on the User entity (either a string or a number)

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -1801,7 +1801,10 @@ export class EntityService {
     validateFieldName(data.name);
 
     // Validate the field's dataType is not a system data type
-    if (isSystemDataType(data.dataType as EnumDataType)) {
+    if (
+      isSystemDataType(data.dataType as EnumDataType) &&
+      data.dataType !== "Id"
```
yuval-hazaz, line 1806: please use the enum value instead of a string

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -2057,7 +2060,10 @@ export class EntityService {
       include: { entityVersion: true },
     });
 
-    if (isSystemDataType(field.dataType as EnumDataType)) {
+    if (
+      isSystemDataType(field.dataType as EnumDataType) &&
+      field.dataType !== "Id"
```
yuval-hazaz, line 2065: please use the enum value instead of a string

yuval-hazaz, line 2065: since now we skip the current check, we should think about adding a new check specifically for a field of type "ID" - that the user is only changing the "idType" property and no other property

yuval-hazaz (changes requested):
@abrl91 I assume that some of the feedback is on work done originally on the branch, but please carefully consider all changes and if there are changes that are not required or does not make sense - please remove them

packages/amplication-client/src/Entity/EntityFieldForm.tsx
abrl91: These are all disabled on type ID in production
image

yuval-hazaz (changes requested):
Well done. some comments but nothing critical

packages/amplication-data-service-generator/src/server/auth/user-info/user-info.template.ts
```diff
@@ -1,12 +1,13 @@
 import { Field, ObjectType } from "@nestjs/graphql";
-// @ts-ignore
-// eslint-disable-next-line
-import { User } from "../user/user";
+import { User } from "../user/base/user";
```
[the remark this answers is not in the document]
abrl91: I noticed that on the generated app we are do in it with “../user/user” with its ignore and es lint ignore. with this path we don't need it

packages/amplication-client/src/Entity/EntityFieldForm.tsx
```diff
@@ -124,47 +126,52 @@ const EntityFieldForm = ({
...
+              <DataTypeSelectField
+                label="Data Type"
+                disabled={isSystemDataType}
+              />
             )}
+
             <SchemaFields
               schema={schema}
-              isDisabled={isDisabled}
+              isDisabled={isSystemDataType && !isIdDataType}
```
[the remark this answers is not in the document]
abrl91: Without this prop, the option input is disabled. Maybe we do need to run the migration

packages/amplication-data-service-generator/src/server/auth/token/create-token-payload-interface.ts
```diff
@@ -23,17 +23,23 @@ export async function createTokenPayload(): Promise<Module> {
...
+    uuid: () => builders.identifier("string"),
+    cuid: () => builders.identifier("string"),
   };
 
+  const templatePath = require.resolve("./token-payload-interface.template.ts");
+  const file = await readFile(templatePath);
+
   interpolate(file, {
-    ID_TYPE: idTypeTSOptions[idType](),
+    ID_TYPE: idTypeTSOptions[idTypeMap[idType]](),
```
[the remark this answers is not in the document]
abrl91: That's what I did at first, but the pointing rules caught me 😅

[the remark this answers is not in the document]
abrl91: https://github.com/amplication/amplication/pull/4379/commits/829a551eab4f3a5b72009da5e4c211cea87d57b8

packages/amplication-client/src/Entity/EntityFieldForm.tsx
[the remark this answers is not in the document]
yuval-hazaz: I mean that you need to remove the prop "isDisabled" completely (unless we keep if for other logic to be able to set the entire page as disabled regardless of system data type or not)

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -2232,6 +2261,37 @@ function isUserEntity(entity: Entity): boolean {
...
+ * @param data the payload from the request
+ * @return (boolean) true if the base properties of the ID field are changed,
+ * meaning that the base properties of the ID field from the payload are different
+ * from the base properties of the ID field from the initial state (INITIAL_ID_TYPE_FIELDS)
+ */
+function isBasePropertyIdFieldPayloadChanged(
+  data:
+    | CreateOneEntityFieldArgs
+    | UpdateOneEntityFieldArgs
+    | EntityFieldUpdateInput
```
yuval-hazaz, line 2274: these types does not make sense together

yuval-hazaz (changes requested):
last minor change

yuval-hazaz (approved):
@abrl91 well done!

yuval-hazaz:
@GreenMachine01 @arielweinberger this change has a big impact on the generated code, and I recommend assigning someone else to run some test E2E on the different options

tupe12334 (approved):

packages/amplication-client/src/Entity/EntityField.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useContext, useState } from "react";
+import { useCallback, useMemo, useContext, useState } from "react";
```
barshimi, line 1: React suppose to be at any component file

packages/amplication-data-service-generator/src/util/get-user-id-type.ts
```diff
@@ -0,0 +1,25 @@
+import { types } from "@amplication/code-gen-types";
+import DsgContext from "../dsg-context";
+import { EnumDataType } from "../models";
+import { USER_ENTITY_NAME } from "../server/user-entity";
+
+export const getUserIdType = () => {
+  const { entities } = DsgContext.getInstance;
```
barshimi, line 7: dont you prefer to wrap it with try/catch instead ?

packages/amplication-client/src/Entity/EntityField.tsx
overbit: I think React import is no longer needed from React v17 (we are now at v18+) @barshimi

PR #4383, opened 2022-11-17
refactor(client): rename variable to match real use

overbit:
## PR Details

Minor refactoring involving renaming an env variable to match the real use of it.

## PR Checklist


helm/charts/services/amplication-client/values.yaml
```diff
@@ -34,5 +34,5 @@ config:
     NX_REACT_APP_SERVER_URI: https://server.amplication-dev.com
     NX_REACT_APP_DATA_SOURCE: https://server.amplication-dev.com/graphql
     NX_REACT_APP_GITHUB_CONTROLLER_LOGIN_URL: https://server.amplication-dev.com/github
-    NX_REACT_APP_GITHUB_CLIENT_ID: a4e9055e5f21448d6de7
+    NX_REACT_APP_GITHUB_AUTH_ENABLED: a4e9055e5f21448d6de7
```
EugeneTseitlin, line 37: That looks weird, env with enabled at the end usually means boolean value, here we have string. What was the intention?

overbit: Great point. 
I'll update the values accordingly

morhag90 (approved):

Issue #4373, opened 2022-11-16
Upgrade Prisma to V4


tupe12334:
In the process of working on #4319 I needed the capability to do an advanced JSON filtration.
As part of my research, I notice that there is an option to do it natively with Prisma version 4 or later.

References links:
https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4
https://www.prisma.io/docs/concepts/components/prisma-client/working-with-fields/working-with-json-fields#filter-on-object-property

- [x] #4382
- [x] #4413

yuval-hazaz:
@GreenMachine01 besides the technical side of things, I am in favor to also upgrade to Prisma 4 on the generated code.

We should start by assigning a resource to review the list of breaking changes and estimate the impact.

@abrl91 please have a look at this item about 1-1 relation. Is that helping us with solving the issue we currently have? https://www.prisma.io/docs/guides/upgrade-guides/upgrading-versions/upgrading-to-prisma-4#explicit-unique-constraints-on-one-to-one-relations

abrl91:
@yuval-hazaz @GreenMachine01 
I started a kind of POC a few months ago (on the generated app)
https://www.notion.so/amplication/Prisma-Upgrade-Strategy-80234211e9ec4dfc951f4f97429d9881

And maybe we should do the upgrade before investing more time in "connecting to existing DB" because things have changed. For example: "implicit => explicit" regarding foreign keys and relations

PR #4397, opened 2022-11-21
Adding more info for delete releated entity

### LINKED ISSUE #4319: When deleting an entity show "are you sure" notification if entity is linked to other entities

tupe12334:
Close #4319 

## PR Details

This PR adds info about the connections of the entities and makes sure the related fields are deleted also

## PR Checklist


tupe12334:
This PR depends on #4373 and removing the preview flag in the `prisma.schema` file

morhag90 (approved):

PR #4405, opened 2022-11-22
abort generate code when user turn it on

### LINKED ISSUE #3954: 🐛 Bug Report: When a plugin event fails, there is no indication of a failure
### What happened?

When a plugin event fails, there is no indication of a failure.

For example, when a user installs the MySQL plugin, and on one of the entities, he has a field with the `MultiSelectOptionSet` type. Prisma's MyQL provider doesn't support an array of primitives. Therefore the plugin event fails, **but** there is no indication of a failure on the build page, as the plugin wrapper function doesn't handle these kind of errors

Screen Shot 2022-09-24 at 21 19 42

I expect the build to fail

### How to reproduce

 - create a project with a service
 - install the MySQL plugin on this service
 - create a field type `MultiSelectOptionSet` on of the entities
 - Click on build

### Amplication version

next

### Environment

_No response_

_No response_

barshimi:
Close #3954 

## PR Checklist


packages/amplication-data-service-generator/src/plugin-wrapper.ts
```diff
@@ -91,7 +92,7 @@ const pluginWrapper: PluginWrapper = async (
         error
       )}`,
     });
-    return [];
+    return context.utils.abortGeneration ? process.exit(0) : [];
```
yuval-hazaz, line 95: I think we should abort more gracefully and add a message to the logger

instead of using a boolean, we can use a function context.abortGeneration(message)

yuval-hazaz (changes requested):

Issue #4407, opened 2022-11-22
As User - I want projects in the side panel to be sorted based on 'create date' in descending order (newer project at the top)


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/203289193-fde802ca-2bf7-4882-bc02-479be28c05a9.png)

**Overview:**
Today if you have many projects it will be hard to find the project you want to work on as the order of the projects is not clear to the user.
We want to make it easier for users with multiple projects to find the projects they want to work on by sorting the projects in the list in descending order according to the create date of the project  (newer project at the top)

PR #4416, opened 2022-11-22
sort project list by create date

barshimi:
Close #[issue-number] #4407

## PR Checklist


abrl91 (approved):

packages/amplication-client/src/Workspaces/hooks/useProjectSelector.ts
```diff
@@ -71,7 +71,11 @@ const useProjectSelector = (
   useEffect(() => {
     if (loadingList || !projectListData) return;
 
-    setProjectList(projectListData.projects);
+    const sortedProjects = [...projectListData.projects].sort((a, b) => {
```
yuval-hazaz, line 74: I think we should use server-side sorting

yuval-hazaz (changes requested):

packages/amplication-client/src/Workspaces/hooks/useProjectSelector.ts
overbit: What if in the future we want to offer different sorting filters? IMHO server response should contain enough details to allow client the sorting they prefer. 
If we want to enforce a default sorting, we should consider an order/index property instead of relying on the intrinsic order of the Array in the json response.

yuval-hazaz: The graphQL API supports a parameter for sorting that also takes pagination into consideration. 
Since the list of projects is usually short and we anyway don't use pagination now, we can use client-side sorting - but we already have support for server-side sorting, so all we need to do is to send the name of the field to the server

image

PR #4472, opened 2022-11-28
feat(dsg): add DSG version to logs page

### LINKED ISSUE #4468: Add DSG version to the build logs
### Feature description

Add DSG version to the build logs

### Use case
In order to be sure which DSG version is running, it would be helpful to add a log with the DSSG version in the build page

_No response_

_No response_

abrl91:
Close #4468 

## PR Details

add DSG vestion to logs page

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/project.json
+++ b/packages/amplication-data-service-generator/project.json
@@ -36,7 +36,7 @@
       ],
       "outputs": ["{projectRoot}/node_modules"],
       "options": {
-        "commands": ["npm ci"],
+        "commands": ["npm ci", "ts-node ./scripts/update-version.ts"],
         "parallel": false,
         "cwd": "packages/amplication-data-service-generator"
       }
--- a/packages/amplication-data-service-generator/src/server/create-server.ts
+++ b/packages/amplication-data-service-generator/src/server/create-server.ts
@@ -44,10 +44,10 @@ async function createServerInternal(
   } = DsgContext.getInstance;
 
   const dsgVersion = (await import("../../package.json")).version;
-  logger.info(`running DSG version ${dsgVersion}`);
+  logger.info(`Running DSG Version: ${dsgVersion}`);
   await createLog({
     level: "info",
-    message: `running DSG version ${dsgVersion}`,
+    message: `Running DSG Version: ${dsgVersion}`,
   });
 
   await createLog({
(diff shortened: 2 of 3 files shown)


packages/amplication-data-service-generator/src/server/create-server.ts
```diff
@@ -43,6 +43,13 @@ async function createServerInternal(
     logger,
   } = DsgContext.getInstance;
 
+  const dsgVersion = (await import("../../package.json")).version;
+  logger.info(`running DSG version ${dsgVersion}`);
+  await createLog({
+    level: "info",
+    message: `running DSG version ${dsgVersion}`,
```
tupe12334, line 50:
```suggestion
    message: `Running the DSG in version ${dsgVersion}`,
```

packages/amplication-data-service-generator/project.json
```diff
@@ -36,7 +36,7 @@
       ],
       "outputs": ["{projectRoot}/node_modules"],
       "options": {
-        "commands": ["npm ci"],
+        "commands": ["npm ci", "ts-node ./scripts/update-version.ts"],
```
tupe12334, line 39: I suggest adding this as a separate target to the project.json and call it as dependence. What do think?

packages/amplication-data-service-generator/src/server/create-server.ts
abrl91: I will change it to `Running DSG Version: ${dsgVersion}`

packages/amplication-data-service-generator/project.json
abrl91: I don't think it is necessary. What are the benefits? Do we need this command elsewhere?

tupe12334: I think it's important to separate commands for there proper use; in that case, I will add `update-version` script that runs prebuild, preserve and prepublish.

yuval-hazaz:
@abrl91 are we missing anything to move forward with this one?

abrl91:
> @abrl91 are we missing anything to move forward with this one?

@arielweinberger can you review, please? I remember you said that you want to implement it in a different way (not using the package.json)

arielweinberger:
I don't want to block the efforts so I'm okay with this for now.

arielweinberger (approved):

Issue #4515, opened 2022-12-01
🐛 Bug Report: missing default `CUID` id type when the `idType` property returns as null or undefined


abrl91:
### What happened?

missing default `CUID` id type when the `idType` property returns as null or undefined.
As the following part is new in our data
```
"properties": {
    "idType": {
      "type": "string",
      "enum": ["CUID", "UUID", "AUTO_INCREMENT"],
      "default": "CUID"
    }
  }
```
And we added a feature to support more id types, we must have some kind of backward comparability and if the `idType` returns as null or undefined, set it to the default `CUID`

### Amplication version

1.1.1

PR #4516, opened 2022-12-01
fix - default id type

abrl91:
Close: #4515 

## PR Details

set `CUID` as a default id type when it returns a null or undefined 
## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/util/get-user-id-type.ts
+++ b/packages/amplication-data-service-generator/src/util/get-user-id-type.ts
@@ -1,4 +1,5 @@
 import { types } from "@amplication/code-gen-types";
+import { isNull, isUndefined } from "lodash";
 import DsgContext from "../dsg-context";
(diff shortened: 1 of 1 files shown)


packages/amplication-data-service-generator/src/util/get-user-id-type.ts
```diff
@@ -20,6 +20,9 @@ export const getUserIdType = () => {
     throw new Error("User entity must have an id field");
   }
 
-  const { idType } = idField.properties as types.Id;
+  let { idType } = idField.properties as types.Id;
+  if (!idType) {
+    idType = "CUID";
+  }
   return idType;
```
overbit, line 27: Let's try to be more concise by using the nullish coalescing operator

```suggestion
  const { idType } = idField.properties as types.Id;
  return idType ?? "CUID";
```

overbit (approved):
💯

Issue #4429, opened 2022-11-23
Review Continuous Integration (Nx - Affected) workflows triggers


overbit:
Currently we trigger `Continuous Integration (Nx - Affected)` workflows on:
- workflow_dispatch
- push
- pull_request_target

`nx affected` commands work out the application and library affected by a diff between a BASE and HEAD commits. https://nx.dev/concepts/affected 

In order to return rightful results, for each workflow, we will need to specify the base commit (or branch).
This is only possible in `pull_request` triggers and, with manual input, also workflow dispatch, but not in `push` since it is only aware about the commit (not branch).

As a plus, review the jobs order and raise a new issue if any work is needed

overbit:
Blocked by https://github.com/nrwl/nx-set-shas/pull/80

PR #4522, opened 2022-12-02
chore: fix ci affected to work on next to master PR

overbit:
Close: #4429 

## PR Details

Update ci affected workflow to only be triggered on actions against a pull request. 
This will fix the scenario where CI workflow triggered by push on `next` branch will end not running any build due to Nx not detecting any changes that will affect projects due to the comparison of the HEAD of the `next` branch against itself.

## PR Checklist

### DIFF
--- a/.github/workflows/ci-nx.yml
+++ b/.github/workflows/ci-nx.yml
@@ -77,7 +77,9 @@ jobs:
       - name: Set Affected
         id: set-affected
         run: |
-          echo ::set-output name=affected::"$(npx nx print-affected --type=app | jq -c .projects)"
+          export AFFECTED_PROJECTS=$(npx nx print-affected --type=app | jq -c .projects)
+          echo affected projects: $AFFECTED_PROJECTS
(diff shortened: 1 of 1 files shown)


overbit:
There is an issue with one of the action we use.
⚠️  This PR is blocked until the `nx-set-shas` action gets fixed:  https://github.com/nrwl/nx-set-shas/pull/80

arielweinberger:
Is it worth using a fork in the meantime?

overbit:
Actually, I might be wrong. Since the `pull_request_target` runs in the context of the base of the pull request meaning we would not get the latest changes of a PR. We should use the `pull_request` event in this case

arielweinberger:
The downside is that pull_request does not allot non-maintainers to get access to secrets (even if we manually approve the runs). So any PR from maintainers won't be able to successfully run CI.

overbit:
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

arielweinberger (approved):

Issue #4449, opened 2022-11-27
Plugthon: As Developer I want to remove unnecessary files from the build (ie. MacOs ._file and DS_store​)


GreenMachine01:
Can be done as part of https://github.com/amplication/amplication/issues/4448

Add the filtering as part of the util library

barshimi:
it wont be part of #4448 because this check will be part of the writ file loop. before writing files to generated folder it will check the file extension.

EugeneTseitlin:
@barshimi Please, create PR and connect it to the issue if you've finished.

PR #4546, opened 2022-12-06
prevent writing operating system files to generated code

barshimi:
Close: #[issue-number] #4449

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/generate-code.ts
+++ b/packages/amplication-data-service-generator/src/generate-code.ts
@@ -54,6 +54,8 @@ export default async function generateCode(
   }
 }
 
+const filesToFilter = [".DS_Store", "._file"];
+
 async function writeModules(
   modules: Module[],
   destination: string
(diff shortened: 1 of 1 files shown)


overbit (changes requested):

packages/amplication-data-service-generator/src/generate-code.ts
```diff
@@ -63,6 +65,8 @@ async function writeModules(
   console.info(`Writing modules to ${destination} ...`);
   await Promise.all(
     modules.map(async (module) => {
+      if (filesToFilter.test(module.path)) return;
```
yuval-hazaz, line 68: I dont think we should add any logic at this layer. If the DSG decided to includes any files in the results, they should be included. 
I think we can add this filter as part of the util "CopyStaticFiles" - and should document that behavior

yuval-hazaz (changes requested):
See my comment. Let me know what you think.

yuval-hazaz (approved):

Issue #4543, opened 2022-12-05
Several endpoints for the DSG


tupe12334:
As of now, we have several endpoints to call the DSG.

- https://github.com/amplication/amplication/blob/master/packages/amplication-data-service-generator/src/generate-code.ts
- https://github.com/amplication/amplication/blob/master/packages/amplication-data-service-generator/src/create-data-service.ts
- https://github.com/amplication/amplication/blob/master/packages/amplication-data-service-generator/src/create-data-service-impl.ts

We need to standardize the entrance to the DSG in only one endpoint.

PR #4558, opened 2022-12-06
Remove old script export

tupe12334:
Close: #4543 

## PR Details

This PR removes the old`createDataService` that not in use anymore
## PR Checklist


packages/amplication-data-service-generator/src/create-data-service.ts
```diff
@@ -1,17 +0,0 @@
...
-import { EnumResourceType } from "./models";
-
-export async function createDataService(
-  dSGResourceData: DSGResourceData,
-  logger: winston.Logger = defaultLogger
-): Promise<Module[]> {
-  if (dSGResourceData.resourceType === EnumResourceType.MessageBroker) {
-    logger.info("No code to generate for a message broker");
-    return [];
-  }
```
yuval-hazaz, line 15: what about this logic?

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/create-data-service.ts
tupe12334: Its also in the createDataServiceImpl

Issue #1959, opened 2021-11-15
⭐️ Plugin: MongoDB


rachgreen:
### 💡 My idea is

Add suport for MongoDB. 

### ✅ The idea is awesome because

I like no sql db

zaitsev1:
https://www.prisma.io/docs/concepts/database-connectors/mongodb

yuval-hazaz:
candidate to be implemented as a plugin

GreenMachine01:
@yuval-hazaz @arielweinberger 
What are the client-side implications of supporting this plugin? 
Do we need to handle the service configuration differently if the Mongo plugin is activated on it?

abrl91:
I review the plugin https://github.com/amplication/plugins/pull/25

PR #4570, opened 2022-12-06
Refactor/create prisma fields

yuval-hazaz:
part of https://github.com/amplication/amplication/issues/1959

## PR Details

This PR refactors the way DSG generates the Prisma schema fields. 
Instead of using a single function with Switch Case for all data types, we created a new object with separate functions one for each data type. 

This object will be used to expose the functions to a plugin developer to use a special logic for different DB types, in our case to generate some fields differently for MongoDB

## PR Checklist


packages/amplication-data-service-generator/src/server/prisma/create-prisma-schema-fields.ts
```diff
@@ -0,0 +1,482 @@
...
+    | [PrismaSchemaDSLTypes.ScalarField]
+    | [PrismaSchemaDSLTypes.ObjectField]
+    | [PrismaSchemaDSLTypes.ObjectField, PrismaSchemaDSLTypes.ScalarField];
+} = {
+  [EnumDataType.SingleLineText]: (
+    field: EntityField,
+    entity: Entity,
+    fieldNamesCount: Record<string, number> = {}
+  ) => {
+    const { dataType, name, properties } = field;
```
abrl91, line 58: Please clean up the unused properties from every function. The lint is also yelling at you because of the unused variables

abrl91: and also, choose whether to destructure all the properties you need,
or do not destructre at all and use `field.{property}`, then return the value without the `return` keyword

abrl91 (changes requested):

abrl91 (approved):

PR #4576, opened 2022-12-07
Dynamic install of plugins

tupe12334:
This PR adds the logic of dynamically installation of plugins in the generate-code DSG executer


packages/amplication-data-service-generator/src/generate-code.ts
```diff
@@ -19,14 +18,28 @@ generateCode(source, destination).catch((err) => {
...
+    const resourceData = await readInputJson(source);
+    const { pluginInstallations } = resourceData;
+
+    console.log("Start code generation dynamic plugins installation");
+    await dynamicPluginInstallation(
+      pluginInstallations.map((plugin) => plugin.npm)
+    );
+    console.log("Finish code generation dynamic plugins installation");
+
+    const modules = await (
```
barshimi, line 40: why don't you import the file before the run and use it on invoke ?

packages/amplication-data-service-generator/src/tests/pluginInstallation.ts
```diff
@@ -1,3 +1,3 @@
 import { PluginInstallation } from "@amplication/code-gen-types";
-
-export const installedPlugins: PluginInstallation[] = [];
+import { plugins } from "../../scripts/constants/example-plugins";
```
barshimi, line 2: again

packages/amplication-data-service-generator/src/dynamic-plugin-installation.ts
```diff
@@ -0,0 +1,16 @@
...
+    child.addListener("exit", resolve);
+  });
+}
+
+export async function dynamicPluginInstallation(
+  pluginInstallations: string[]
+): Promise<void> {
+  await promiseFromChildProcess(
+    exec(`npm i ${pluginInstallations.join(" ")} --no-save -f`)
+  );
```
overbit, line 15: This approach might bring us easily into a dependency hell https://npm.github.io/how-npm-works-docs/theory-and-design/dependency-hell.html.

Plus is locking our solution to only support NPM leaving most used package managers like yarn and pnpm out to support. 
Possibly, a stronger solution would be to bundle all the dependencies of a plugin within the plugin itself.

overbit (changes requested):

packages/amplication-data-service-generator/src/dynamic-plugin-installation.ts
overbit: Plus we might have network issues (that are always a possibility) that will affect the installation of npm packages in runtime and, consequently, the code generation.

GreenMachine01:
@tupe12334 Please see feedback and resolve conflicts.

Issue #4582, opened 2022-12-07
🐛 Bug Report: Compilation error in the generated app - `Type 'null' is not assignable to type [EntityName]`


abrl91:
### What happened?

![image](https://user-images.githubusercontent.com/39680385/206211171-d344c386-57a9-464d-ba75-39f0c18dc8e9.png)

### What you expected to happen

no compilation error

### How to reproduce

- create a project with a service
- sync with GitHub
- clone the generated app
- run:
```bash
cd server && npm i && npm run docker:db && npm run prisma:generate && npm run db:init && npm run start:debug
```
see the errors

### Amplication version

1.1.1

### Are you willing to submit PR?

Yes I am willing to submit a PR!

yuval-hazaz:
@abrl91, we need to proceed with fixing this issue. It impacts anyone who creates a service with the lastest version. 

I think the solution would be to update the return type of the function to Promise<Entity[] | null> 
mainly, we need to check the impact of this change and see that there is no code that will break because of this change where it expects to get an empty Array but instead it gets Null now

PR #4606, opened 2022-12-09
Fix Prisma one-to-many relation compilation errors

abrl91:
Close: #4582

## PR Details
In the new version of Prisma, the return type in  one-to-many relation has changed to 
`PrismaPromise<Array<OrderGetPayload<T>>| Null>`
In our API, when there are no results, we return an empty array and not null.
We reached out to Prisma on Slack, and they gave us the two options:
- use `findUniqueOrThrow` instead of `findUnique`
- add a `!` before  `.orders()`  => ` await findUnique(...)!.orders(...)`

We decided to go with the first solution.

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator/src/server/resource/service/to-many.template.ts
+++ b/packages/amplication-data-service-generator/src/server/resource/service/to-many.template.ts
@@ -14,7 +14,7 @@ export class Mixin {
     parentId: string,
     args: Prisma.ARGS
   ): Promise<RELATED_ENTITY[]> {
-    return await this.prisma.DELEGATE.findUniqueOrThrow({
+    return this.prisma.DELEGATE.findUniqueOrThrow({
       where: { id: parentId },
     }).PROPERTY(args);
   }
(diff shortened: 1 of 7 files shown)


packages/amplication-data-service-generator/src/server/resource/service/to-many.template.ts
```diff
@@ -16,6 +16,6 @@ export class Mixin {
   ): Promise<RELATED_ENTITY[]> {
     return this.prisma.DELEGATE.findUnique({
       where: { id: parentId },
-    }).PROPERTY(args);
+    }).PROPERTY(args) as Promise<RELATED_ENTITY[]>;
```
overbit, line 19: I would always recommend avoiding casting wherever possible. 
How about returning empty array `[]` when prisma returns `null`?! 

```suggestion
return await this.prisma.DELEGATE.findUnique({
      where: { id: parentId },
    }).PROPERTY(args) ?? [];
```

overbit (changes requested):

packages/amplication-data-service-generator/src/server/resource/service/to-many.template.ts
abrl91: I really like this idea

barshimi (approved):

yuval-hazaz (approved):

Issue #4521, opened 2022-12-01
Update nx to latest version


overbit:
### Feature description

Nx 15 brings good news: performance improvement, cleanup for pure JS/TS packages, and fine-tuning Caching with Inputs.
https://blog.nrwl.io/whats-new-in-nx-15-7e14e1ff282d

We should update Nx to the latest version leveraging the migration script offered by Nx itself
```
npx nx migrate latest
# After updating your dependencies, run any necessary migrations.
npx nx migrate --run-migrations
```

PR #4607, opened 2022-12-09
chore(nx): update nx dependencies to latest version v15.3.0

overbit:
Close: #4521 

## PR Details

Upgraded monorepo framework to latest version

## PR Checklist


abrl91:
Does it something that can work for us? (for the client)
https://youtu.be/KBFQZw5ynFs?t=314

overbit:
> Does it something that can work for us? (for the client) https://youtu.be/KBFQZw5ynFs?t=314

We already manually fully migrated the client to nx. In future we could try that approach for the `cli` and  `dsg` @abrl91

libs/util/nestjs/kafka/project.json
```diff
@@ -1,6 +1,6 @@
 {
-  "$schema": "../../../../node_modules/nx/schemas/project-schema.json",
   "name": "util-nestjs-kafka",
+  "$schema": "../../../../node_modules/nx/schemas/project-schema.json",
```
barshimi, line 3: is there a way to create alias to node module, instead of using that root ?

overbit: Potentially, but this transformation has been performed by nx migration script so I would avoid any change for now.

yuval-hazaz (approved):

Issue #4613, opened 2022-12-11
☂️ As Dev - I want to change the Prisma client strategy in the DSG


abrl91:
- [x] create Prisma module and service in the DSG
- [x] use the Prisma module in the root `app.module.ts` 
- [x] use the Prisma service in the entity services
- [x] update the prisma client in `schema.prisma` with the output property
- [x] update nestjs-cli
- [x] update import form `type-fest`

GreenMachine01:
@abrl91 can we close this epic?

abrl91:
Fixed by https://github.com/amplication/amplication/pull/4626Fixed by https://github.com/amplication/amplication/pull/4626Fixed by https://github.com/amplication/amplication/pull/4626

PR #4626, opened 2022-12-12
Implement a new prisma client strategy in the generated app

abrl91:
Close: #4613

## PR Details

- update import form `type-fest`
- update nestjs-cli
- create Prisma module and service in the DSG
- use the Prisma module in the root `app.module.ts` 
- use the Prisma service in the entity services
- update the prisma client in `schema.prisma` with the output property
- add `path` to ts-config in order to be able to use simple import such as `generated-prisma-client` (instead of `../../`
- update clean and seed imports
- update health
- update user data decorator
- update http-exception

## PR Checklist


packages/amplication-data-service-generator/src/server/resource/module/module.base.template.ts
```diff
@@ -1,19 +1,14 @@
 import { Module, forwardRef } from "@nestjs/common";
 import { MorganModule } from "nest-morgan";
-import { PrismaModule } from "nestjs-prisma";
+
```
barshimi, line 3: remove empty line

packages/amplication-data-service-generator/src/server/resource/service/service.base.template.ts
```diff
@@ -1,10 +1,10 @@
-import { PrismaService } from "nestjs-prisma";
+import { PrismaService } from "../../prisma/prisma.service";
```
barshimi, line 1: it would be nice if we'll create alias to pPisma in our generated code instead of import it as it is now

packages/amplication-data-service-generator/src/server/static/src/auth/userData.decorator.ts
```diff
@@ -1,7 +1,7 @@
 import { createParamDecorator, ExecutionContext } from "@nestjs/common";
 import { GqlContextType, GqlExecutionContext } from "@nestjs/graphql";
 //@ts-ignore
```
barshimi, line 3: is it possible to avoid ts-ignore now ?

abrl91: no, because it is still not the correct path. but we need to fix the check on static files and then I think we will be able to remove it

barshimi (approved):

packages/amplication-data-service-generator/src/server/resource/service/service.base.template.ts
abrl91: We will create a new PR and add `webpack.config` file to handle this

abrl91: This was resolved by using `ts.config.json`, providing a "paths"

packages/amplication-data-service-generator/src/server/resource/service/service.template.ts
```diff
@@ -1,6 +1,5 @@
 import { Injectable } from "@nestjs/common";
-import { PrismaService } from "nestjs-prisma";
-
+import { PrismaService } from "../prisma/prisma.service";
```
yuval-hazaz, line 2: this code will never be updated in existing apps because this file is ignored. we need to think how to convert this to the user (docs?)

packages/amplication-data-service-generator/src/server/static/scripts/customSeed.ts
```diff
@@ -1,4 +1,4 @@
-import { PrismaClient } from "@prisma/client";
+import { PrismaClient } from "../prisma/generated-prisma-client";
```
yuval-hazaz, line 1: this code will never be updated in existing apps because this file is ignored. we need to think how to convert this to the user (docs?)

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/server/resource/service/service.template.ts
abrl91: I will write a description for the release notes and the docs

packages/amplication-data-service-generator/src/server/static/scripts/customSeed.ts
abrl91: I will write a description for the release notes and the docs

yuval-hazaz (approved):

Issue #4599, opened 2022-12-08
Build limitations update


GreenMachine01:
We need to modify the paywall configuration:

**Requirements:**

- We want to limit users from making builds based on meeting one of the below:
    - More than 3 services in a **workspace**
        - This check should be limited to the **workspace** being worked on
    - More than 7 entities in a service
        - This check should be limited to the **project** being worked on
- When the user is blocked from making a build we need to show him the proper reason why he is blocked

PR #4662, opened 2022-12-15
Feat/4599 build limitations update

EugeneTseitlin:
Close: #4599 

## PR Checklist


packages/amplication-server/src/core/project/project.resolver.spec.ts
```diff
@@ -205,7 +205,7 @@ const getPendingChangesMock = jest.fn(() => {
   return [EXAMPLE_PENDING_CHANGE];
 });
 
-describe("ProjectResolver", () => {
+xdescribe("ProjectResolver", () => {
```
tupe12334, line 208:
```suggestion
describe("ProjectResolver", () => {
```

packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -92,9 +103,75 @@ export class ProjectResolver {
...
     return this.projectService.commit(args);
   }
 
+  async validateProject(projectId: string): Promise<void> {
+    const project = await this.projectService.findUnique({
+      where: { id: projectId },
+    });
+    const workspace = await this.workspaceService.getWorkspace({
+      where: { id: project.workspaceId },
+    });
```
tupe12334, line 119: Why not do that as one request, and update the DTO to include the workspace?

packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -92,9 +103,75 @@ export class ProjectResolver {
...
+            projectId: project.id,
+          },
+        });
+
+        const services = resources.filter(
+          (x) => x.resourceType === EnumResourceType.Service
+        );
+
+        return services;
+      });
```
tupe12334, line 141: What do you think about taking this part to an outside function?

packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -92,9 +103,75 @@ export class ProjectResolver {
...
+    const entities = await this.entityService.entities({
+      where: { resourceId: service.id },
+    });
+
+    if (entities.length > 7) {
+      throw new Error(
+        "You have reached the maximum number of entities. (Upgrade your workspace plan)"
+      );
+    }
+  }
```
tupe12334, line 173: Could you please provide more clarity about the purpose of the validateService function? It seems the current name does not accurately describe its function in relation to pay wall validation. It would be helpful to come up with a more specific and descriptive name for this function.

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: Take a look at `findUnique` func, input supports only `where` clause, there is no `include` field. I agree that it would be nice to make most of these operations by one request with joins, but that's the api that we have right now.

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: It checks that service doesn't have more that 7 entities. See the requirements: https://github.com/amplication/amplication/issues/4599

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: Moved to resource service.

packages/amplication-server/src/core/project/project.resolver.ts
tupe12334: @EugeneTseitlin I saw the requirements and i think that the name of the function `validateService` is not clear enough, its not a generic validation its the paywall validation. What do you think about rewriting the name to `paywallServiceValidation` to clarify it?

packages/amplication-server/src/core/project/project.resolver.ts
tupe12334: We can add to the API the include field in the args dto

packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -92,9 +103,58 @@ export class ProjectResolver {
...
+
     return this.projectService.commit(args);
   }
 
+  async validateProject(projectId: string): Promise<void> {
+    const project = await this.projectService.findUnique({
+      where: { id: projectId },
+    });
+    const workspace = await this.workspaceService.getWorkspace({
+      where: { id: project.workspaceId },
```
yuval-hazaz, line 118: all the code and logic should be done in the service, and never in the resolver

yuval-hazaz (changes requested):

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: We can, but the feature is not about refactoring data services, the scope is clear - adding paywall limitation. Again, I agree that would be better, but it has to be done separately.

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: Now I got you. Sure, name is actually too generic. Changing it.

packages/amplication-server/src/core/project/project.resolver.ts
EugeneTseitlin: Done

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -177,6 +177,10 @@ export class BuildService {
       },
     });
 
+    const resource = await this.resourceService.findOne({
```
yuval-hazaz, line 180: please revert the changes on this file if they are not needed

packages/amplication-server/src/core/project/project.module.ts
```diff
@@ -16,6 +18,9 @@ import { ProjectService } from "./project.service";
     BlockModule,
     EntityModule,
     BuildModule,
+    forwardRef(() => ProjectModule),
```
yuval-hazaz, line 21: this is a self reference - please remove

packages/amplication-server/src/core/project/project.resolver.spec.ts
yuval-hazaz: please fix

packages/amplication-server/src/core/project/project.resolver.ts
```diff
@@ -35,6 +35,8 @@ import {
 export class ProjectResolver {
   constructor(
     private projectService: ProjectService,
+
+    @Inject(forwardRef(() => ResourceService))
```
yuval-hazaz, line 39: please revert the changes if they are not needed

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -24,7 +26,11 @@ export class ProjectService {
     private readonly resourceService: ResourceService,
     private readonly blockService: BlockService,
     private readonly buildService: BuildService,
-    private readonly entityService: EntityService
+    private readonly entityService: EntityService,
+    @Inject(forwardRef(() => WorkspaceService))
```
yuval-hazaz, line 30: no need to define forwardRef on the service if you already did it on the module

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -24,7 +26,11 @@ export class ProjectService {
...
+    private readonly workspaceService: WorkspaceService,
+    @Inject(forwardRef(() => SubscriptionService))
```
yuval-hazaz, line 32: no need to define forwardRef on the service if you already did it on the module

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +120,64 @@ export class ProjectService {
...
+      where: { id: projectId },
+    });
+    const workspace = await this.workspaceService.getWorkspace({
+      where: { id: project.workspaceId },
+    });
+    const subscription = await this.subscriptionService.getCurrentSubscription(
+      workspace.id
+    );
+
+    if (!subscription) {
```
yuval-hazaz, line 136:
```suggestion
    //Temporarily, we do not enforce specific subscriptions, if the workspace has a subscription of any kind we do not enforce it 
    if (!subscription) {
```

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +120,64 @@ export class ProjectService {
...
+          "You have reached the maximum number of services. (Upgrade your workspace plan)"
+        );
+      }
+
+      const projectServices = await this.resourceService.resources({
+        where: { projectId: project.id },
+      });
+      const promises = projectServices.map((project) =>
+        this.validateSubscriptionPlanLimitationsForService(project)
+      );
```
yuval-hazaz, line 152: use a query directly on Prisma to find if there are any services in the project in the workspace with more then X entities

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +120,64 @@ export class ProjectService {
...
+    const subscription = await this.subscriptionService.getCurrentSubscription(
+      workspace.id
+    );
+
+    if (!subscription) {
+      const services = await this.resourceService.getWorkspaceServices(
+        workspace.id
+      );
+
+      if (services.length > 3) {
```
yuval-hazaz, line 141: use const LICENSE_PROJECT_LIMIT_SERVICES

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +120,64 @@ export class ProjectService {
...
+  }
+
+  async validateSubscriptionPlanLimitationsForService(
+    service: Resource
+  ): Promise<void> {
+    const entities = await this.entityService.entities({
+      where: { resourceId: service.id },
+    });
+
+    if (entities.length > 7) {
```
yuval-hazaz, line 165: use const LICENSE_SERVICE_LIMIT_ENTITIES

yuval-hazaz (changes requested):

yuval-hazaz (approved):
perfect! minimum footprint with maximum impact!

tupe12334 (approved):

Issue #4658, opened 2022-12-15
Confusing phrasing for serving local dev environment 


Ashkenazzio:
### Describe the issue with documentation

In the README file, Getting Started With Local Development, step number 5:

>5. To start developing, run the `serve` target of the _**desired**_ app:
>
>```bash
>  # This will serve the Amplication Server in development mode
>  npx nx serve amplication-server
>  # This will serve the Amplication Client in development mode
>  npx nx serve amplication-client
>  ```

The phrasing seems confusing to me as one might understand that in order to work on the client he needs to serve the client command only.

### How to solve the problem

I think adding a note right after, explaining that you need to run both server and client in order for the client to work properly should do the trick.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #4663, opened 2022-12-15
added a note under step number 5 in setting up local dev environment

Ashkenazzio:
Close: #4658 

## PR Details

Added a note under step number 5 of setting up dev local environment as suggested in #4658


yuval-hazaz (approved):

yuval-hazaz:
Thanks @Ashkenazzio that is very helpful

Issue #4610, opened 2022-12-11
Splitting the DSG into lib and runner


tupe12334:
As we continue to build the DSG, we are currently viewing it as both the runner and the logic library. In order to better organize our code, we need to divide it into two separate folders: one for the logic library, and another for the runner logic.

Working branch: `feat/4610-dsg-split`

PR #4666, opened 2022-12-15
DSG split to lib and executer

tupe12334:
Close: #4610 
## PR Details

This PR splits the DSG to a logic lib and executes app

## PR Checklist


packages/amplication-data-service-generator-runner/scripts/create-input-json.ts
```diff
@@ -0,0 +1,26 @@
...
+
+function createInputJsonFile() {
+  const object = {
+    entities,
+    roles,
+    resourceInfo: appInfo,
+    resourceType: EnumResourceType.Service,
+    pluginInstallations: [plugins.kafka, plugins.postgresPlugin],
+  };
+  const fileName = "input.json";
```
EugeneTseitlin, line 22: Why input is hardcoded here? It supposed to be configurable, path in file system may change.

EugeneTseitlin: As far as I see this script is for development purposes only, it creates input json from test entities, roles and etc. Am I right?

tupe12334: Yes, you are right and this is also the answer to your question.

Issue #4431, opened 2022-11-23
Services without Authentication module (Logic)


GreenMachine01:
**As User - I want to be able to set up my service without Authentication module (in the generated app) so that it can run without authentication and authorization** (Logic)

**Overview:**
Some developers will want the service to not have any users set up.
This might be useful when you need background services that don't expose endpoints to end users so there is no need for authentication and authorization
This also might be useful when you need a Public API.

This is the blueprint and diff for a "no auth" service generated by Amplication:
https://github.com/amplication/microservices-auth-poc/pull/1/files

**Requirements:**

- **At this stage there is no need to have UI to control the feature**
- Show that you can create a service without the authentication module
- Show that you can use or not use authentication module on demand via a hardcoded configuration
- Show that you can remove the 'User' **entity** if needed
- Show that you can remove the 'User' **role** if needed

---
- [x] remove the auth logic from the DSG - in code review
- [ ] add `user-entity fields` to the DSG context - in code review
- [ ] #4671

GreenMachine01:
Split into logic (8 - Mor + Amit) and UI (5 - Eugene + Shimi)

GreenMachine01:
This is the blueprint and diff for a "no auth" service generated by Amplication:
https://github.com/amplication/microservices-auth-poc/pull/1/files

GreenMachine01:
Mor - 5
Amit - 3

abrl91:
## Test Results:
### Generate a **new** service in a **new** project **without any plugin**:
1. environment variables related to authentication: some of them are in the `.env` file (`BCRYPT_SALT`)
![image](https://user-images.githubusercontent.com/39680385/217792497-7c374149-af7f-4a59-8cf9-fd384208b2df.png)

2. the docker-compose file in the generated server is generated with environment variables that don't exist in the `.env` file (`JWT_SECRET_KEY`). The variable shouldn't be in the `.env` file, so this is expected, but we shouldn't use in the docker-compose file this env variable. 

![image](https://user-images.githubusercontent.com/39680385/217792777-fb016dca-8a05-42c0-9fb4-ff4ae74661cd.png)

abrl91:
Test Results - round 2:
- automatically installed auth core and JWT - passed ✅


https://user-images.githubusercontent.com/39680385/219058838-ff622d8f-86ce-4d10-8ea6-992a96cd4fa4.mov

PR #4670, opened 2022-12-16
Refactor - move user entity to the DSG context

abrl91:
related issue: #4431 

## PR Details

 move user entity from constant to the DSG context

## PR Checklist


yuval-hazaz (commented):
@abrl91 I suggest we will do the review together because there are some things I don't understand

packages/amplication-data-service-generator/src/server/seed/create-seed.ts
```diff
@@ -133,7 +152,10 @@ export function createUserObjectCustomProperties(
       field,
       createDefaultValue(field, userEntity),
     ])
-    .filter(([field, value]) => !AUTH_FIELD_NAMES.has(field.name) && value)
+    .filter(
+      ([field, value]) =>
+        !createAuthFieldNames(USER_AUTH_FIELDS).has(field.name) && value
```
yuval-hazaz, line 157: dont use the function inside the loop

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #4216, opened 2022-10-20
As User - I want a new Message Broker to have a Default Topic already set up for me


GreenMachine01:
**Overview:**
Similar to Services we want a new Message Broker to include an automatically generated default topic as we expect it will help developers to understand how to use it and save some steps which the developer will always do if he wants to use a Message Broker.

**Design Example:**

![Image](https://user-images.githubusercontent.com/112330500/196984306-e4868504-e6ae-460c-9b24-3f9b11c78215.png)

**Requirements:**

- [ ] When a new Message Broker is created add an auto-generated topic
    - [ ] Topic Display Name - "Topic 1"
    - [ ] Topic Name - "topic.1"
    - [ ] Topic Description - "An automatically created topic to be used with the Message Broker"

ageddesi:
Hi, I would like to take this on if you will have me :) 👍

Ahtritus:
Hey @GreenMachine01 I would like to work on this, could you please assign it to me?

GreenMachine01:
@souravjain540 @arielweinberger please see if this item can be done by the community. if yes, we have @Ahtritus that wants to do it.

GreenMachine01:
@Ahtritus assigned to you. Good luck!

Ahtritus:
@GreenMachine01  The default value for Display Name - "Topic 1" is violating the pattern. The help text mentions `Name must only contain letters, numbers, dash, underscore or dot.` I'm changing it to "Topic_1"

GreenMachine01:
@Ahtritus Let's go with "Topic1" for the Display Name
You can fix the issue you found within this ticket.

GreenMachine01:
@arielweinberger why does a display name field have a restriction on using spaces in the text?

Ahtritus:
I had a couple of observations with the current behavior and wanted to clarify them. Currently, when a user adds a new topic, the name they add is prefilled into the `Display Name` and `Name` and we get to see that when we open the topic. 
<img width="959" alt="image" src="https://user-images.githubusercontent.com/53258966/197974673-27242cad-4bf2-462a-8f71-48f4daddd22f.png">


As per the issue created here, I understand that we don't want to have the `Name` and `Display Name` to be what they created the topic with but a default "**Topic1**". In this case, if the user doesn't change the default "Topic1" display name, they won't be able to create a new topic as it would say `"Another record with the same key already exist (resourceId, displayName)"`.
[...]

PR #4675, opened 2022-12-17
implement default topic for message broker

abrl91:
Close: #4216 

## PR Details

create default topic for message broker

https://user-images.githubusercontent.com/39680385/208304568-5082602e-865e-43d5-8a9a-72cd7828654b.mov

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/hooks/useResources.ts
+++ b/packages/amplication-client/src/Workspaces/hooks/useResources.ts
@@ -122,8 +122,9 @@ const useResources = (
     });
     createBroker({ variables: { data: data } }).then((result) => {
       result.data?.createMessageBroker.id &&
+        addBlock(result.data.createMessageBroker.id);
+      result.data?.createMessageBroker.id &&
         refetch().then(() => {
-          addBlock(result.data.createMessageBroker.id);
           resourceRedirect(result.data?.createMessageBroker.id as string);
         });
     });
--- a/packages/amplication-client/src/Workspaces/hooks/useResources.ts
+++ b/packages/amplication-client/src/Workspaces/hooks/useResources.ts
@@ -35,7 +35,8 @@ const createGitRepositoryFullName = (
 const useResources = (
   currentWorkspace: models.Workspace | undefined,
   currentProject: models.Project | undefined,
-  addBlock: (id: string) => void
+  addBlock: (id: string) => void,
+  addEntity: (id: string) => void
 ) => {
   const history = useHistory();
   const { trackEvent } = useTracking();
@@ -92,8 +93,7 @@ const useResources = (
 
   const createService = (
     data: models.ResourceCreateWithEntitiesInput,
-    eventName: AnalyticsEventNames,
-    addEntity: (id: string) => void
+    eventName: AnalyticsEventNames
   ) => {
     trackEvent({
       eventName: eventName,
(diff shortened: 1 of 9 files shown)


abrl91:
@yuval-hazaz  @GreenMachine01 
1. do I need to update the pending changes?
2. what happens when the use creates more than one message broker?

packages/amplication-server/src/core/topic/topic.service.ts
```diff
@@ -39,6 +45,24 @@ export class TopicService extends BlockTypeService<
...
+  async createDefault(resource: Resource, user: User): Promise<Topic> {
+    const resourceNameKebabCase = kebabCase(resource.name.toLowerCase());
+    const resourceNameTopicConvention = resource.name
+      .toLowerCase()
+      .replace(/[\s\-/]/g, ".");
+
+    const defaultTopic: CreateTopicArgs = {
+      data: {
+        displayName: `${resourceNameKebabCase}-${DEFAULT_TOPIC_DISPLAY_NAME}`,
+        name: `${resourceNameTopicConvention}.${DEFAULT_TOPIC_NAME}`,
```
yuval-hazaz, line 57: I think we can go with a fixed name like topic.sample

yuval-hazaz (changes requested):
@abrl91 I added a small comment about the name of the default topic.
One more thing - I expected to see in the video the default topic in the pending changes list, just like we see the User entity when creating a Service.

packages/amplication-server/src/core/topic/topic.service.ts
abrl91: I left a comment earlier here, asking what happens when there is more than one message broker. That's what I did it like that

abrl91: > @abrl91 I added a small comment about the name of the default topic. One more thing - I expected to see in the video the default topic in the pending changes list, just like we see the User entity when creating a Service.

Yes. I know. Please read my comment 😎
https://github.com/amplication/amplication/pull/4675#issuecomment-1356323028

yuval-hazaz: The name should be unique per broker, and if it is not like that now, it sounds to me like a bug

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -34,7 +34,8 @@ const createGitRepositoryFullName = (
 
 const useResources = (
   currentWorkspace: models.Workspace | undefined,
-  currentProject: models.Project | undefined
+  currentProject: models.Project | undefined,
+  addBlock: (id: string) => void
```
yuval-hazaz, line 38: we need to be consistent with the way we already use addEntity (see below `addEntity`)

yuval-hazaz (changes requested):

barshimi (approved):

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -91,8 +93,7 @@ const useResources = (
 
   const createService = (
     data: models.ResourceCreateWithEntitiesInput,
-    eventName: AnalyticsEventNames,
-    addEntity: (id: string) => void
+    eventName: AnalyticsEventNames
```
yuval-hazaz, line 96: Since you removed the third parameter, I expect to see a change in `packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx`

yuval-hazaz (changes requested):

overbit (approved):

yuval-hazaz (approved):

Issue #4677, opened 2022-12-17
🐛 Bug Report: generated server with auto-increment id: compilation error regarding the `parentId` 


abrl91:
### What happened?

Argument of type `number` is not assignable to parameter of type `string`
![image](https://user-images.githubusercontent.com/39680385/208264731-f9d35a12-370f-4b12-a996-2ba0913e6f0c.png)

### What you expected to happen

No compilation errors

### How to reproduce

Create a service with one-to-many relations where the id type of the "many" is `auto-increment`

### Amplication version

1.1.2

### Are you willing to submit PR?

Yes I am willing to submit a PR!

arielweinberger:
To test this, make sure you have a "one to many" relation in a generated app. One of the relations needs to have an auto incrementing ID.

Creating entities should result in no errors and the ID to be auto incremented.

PR #4678, opened 2022-12-17
Fix  `parentId` when choosing `auto-increment` id type

abrl91:
Close: #4677 

## PR Details
Fix  `parentId` when choosing `auto-increment` id type

## PR Checklist


packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
```diff
@@ -346,6 +357,7 @@ async function createToOneRelationFile(
 
   interpolate(toOneFile, {
     DELEGATE: delegateId,
+    PARENT_ID_TYPE: identityParentIdType(entity),
```
yuval-hazaz, line 360: How do you know which field in the entity to look at just by sending the entity to the function without the current field?

packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
```diff
@@ -375,6 +389,23 @@ async function createToManyRelationFile(
   return toManyFile;
 }
 
+function identityParentIdType(entity: Entity): namedTypes.Identifier {
+  const relatedEntityFiledId = entity.fields?.find(
+    (relatedEntityField) => relatedEntityField.dataType === EnumDataType.Id
```
yuval-hazaz, line 394: By doing this, you will always find the ID field of the entity and not the ID field of the related entity.

packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
```diff
@@ -375,6 +389,23 @@ async function createToManyRelationFile(
...
+  const { idType } = (relatedEntityFiledId?.properties as types.Id) || "CUID";
+  const idTypeTSOptions: {
+    [key in types.Id["idType"]]: namedTypes.Identifier;
+  } = {
+    AUTO_INCREMENT: builders.identifier("number"),
+    UUID: builders.identifier("string"),
+    CUID: builders.identifier("string"),
+  };
+
+  return idTypeTSOptions[idType] ?? idTypeTSOptions["CUID"];
```
yuval-hazaz, line 406: what is this for (idTypeTSOptions["CUID"]) ?

yuval-hazaz (changes requested):

packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
abrl91: You're right. I started with the related entity, but I tested it and it didn't work. I got the same error.

abrl91: I forgot to rename the variables

packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
abrl91: The function is searching for the id type
```
const entityFiledId = entity.fields?.find(
    (entityField) => entityField.dataType === EnumDataType.Id
  );
```

packages/amplication-data-service-generator/src/server/resource/service/create-service.ts
abrl91: for the default

abrl91: I logged the idType and for the entities that don't have this property, it returns `undefined`
image

packages/amplication-data-service-generator/src/util/get-entity-id-type.ts
```diff
@@ -3,17 +3,17 @@ import DsgContext from "../dsg-context";
...
+export const getEntityIdType = (entityName: string) => {
   const { entities } = DsgContext.getInstance;
-  const userEntity = entities.find(
-    (entity) => entity.name === USER_ENTITY_NAME
-  );
+  const entity = entities.find((entity) => entity.name === entityName);
 
-  if (!userEntity) {
+  if (!entity) {
     throw new Error("User entity not found");
```
yuval-hazaz: update message (remove "user" or use dynamic value with ${entityName}

yuval-hazaz (approved):

Issue #4598, opened 2022-12-08
As Dev - I want to automate DSG deployment for next branch


overbit:
To simplify the development experience, we are considering updating the DSG tagging strategy as follows:

For `next` environment:
- update CI to add branch name during container tagging
- update next helm values with hardcoded `next` tag

For `staging` environment:
- update CI to add branch name during container tagging (already done as part of next work)
- update next helm values with hardcoded `master` tag

For `production` environment:
- update CI to add `latest` during container tagging
- update next helm values with hardcoded `latest` tag

EugeneTseitlin:
I've tried already to tag by branch name, but for some reason despite that image pull policy was set to always, updated version was not used. I assume that it's related to the fact that sensor is a custom resource.
Anyway, setting production to latest is not a good idea in my opinion. Even if we want use branch tag, production should use master tag.
I suggest to set hash tag as we do it for amplication server.
@overbit @levivannoort @yuval-hazaz What do you think about it?

yuval-hazaz:
One comment from my side- when moving from staging to production, I think we should retag the latest master tag instead of recreating the image from scratch

arielweinberger:
@yuval-hazaz I don't think we should manually do any retagging. Because this only works smoothly when you move from staging to prod immediately without having any commits/pushes in between.

I think we should keep it as it is and let Docker handle the layers cache. If all layers are identical, then AWS ECR will automatically add it as a tag to an existing image.

levivannoort:
Just to add another flavor into the mix. I'd propose that when a new release is created - so we release a tagged version, i.e. v1.2.1 - that that triggers a pipeline that builds the application based of that tag and pushes it with the explicit released version number.

So that our production environments are always identical to the latest public release.

levivannoort:
And as mentioned before, we need to find a workaround argocd image updater as it currently doesnt work in conjuction with the source.type 'plugin', only with vanilla helm & kustomize. Our secret solution that is used via argocd-vault-plugin requires that the 'plugin' type is used however.

I will spend some time this week looking into if there is a way we could use both.

overbit:
> @overbit I like the proposal and exactly what I imagined would work with Docker metadata.
> 
> Aside from latest for master, shall we also add a 'prod' tag?
> 
> What happens in ArgoCD Image Updater is that it takes a tag and looks for differences in the image sha to trigger a redeployment.
> 
> I think using 'prod' for production would make it align with other envs.
> 

[...]

PR #4721, opened 2022-12-21
Automate dsg deployment

overbit:
Close: #4598

## PR Details

This PR update the tagging strategy for docker images by adding the following tags:
- branch name (for every docker image)
- `latest` when the workflow is manually triggered against the default branch (`master`)  with `add-latest-tag` option `true`

Proof of working tagging
https://github.com/amplication/amplication/actions/runs/3751125014/jobs/6371911764#step:12:57 

## PR Checklist


helm/charts/dsg-runner-infra/templates/sensor.yaml
```diff
@@ -31,7 +31,7 @@ spec:
                     restartPolicy: Never
                     containers:
                       - name: dsg-container
-                        image: "407256539111.dkr.ecr.us-east-1.amazonaws.com/amplication-data-service-generator-runner:sha-04dcb0f"
+                        image:  {{ .Values.job.image }}
```
levivannoort, line 34: Can i suggest that we use two values 'calls' to fill the `image`. Since the `repository` where the image comes from is always the same but only the tag is different:

e.g. job
```
- name: dsg-container
   image: {{ .Values.job.image.repository }}:{{ .Values.job.image.tag }}
```

e.g. values file
```
job:
    image:
        repository: "407256539111.dkr.ecr.us-east-1.amazonaws.com/amplication-data-service-generator-runner"
        tag: "latest"
```

It would allow us to have a pipeline change only the tag when we're in a situation where the argocd image updater still doesnt work in conjunction with vault. Thoughts?

levivannoort (commented):
Pipeline looks good. Let me know how you feel about splitting the template calls for the image.

helm/charts/dsg-runner-infra/templates/sensor.yaml
levivannoort: By the way this is also the way it is done when you do a `helm create <name>`.

overbit: It sound good.

helm/charts/dsg-runner-infra/values.yaml
```diff
@@ -0,0 +1,39 @@
+envName:
```
levivannoort, line 1: I dont think the value 'envName' is used anywhere in the templates.

levivannoort: If so we probably want to remove it for all of the other values.yaml files aswell.

overbit: Good point but this will be replaced by the new ci structure in new infra

levivannoort (approved):

PR #4722, opened 2022-12-21
Use the correct way for class decorators

tupe12334:
This PR updates our current use of adding decorators to class to the standard way

### DIFF
--- a/libs/data-service-generator/src/util/ast.ts
+++ b/libs/data-service-generator/src/util/ast.ts
@@ -406,13 +406,10 @@ export function classDeclaration(
   if (!decorators.length) {
     return declaration;
   }
-  const code = [
-    ...decorators.map((decorator) => recast.print(decorator).code),
-    recast.print(declaration).code,
-  ].join("\n");
-  const ast = parse(code);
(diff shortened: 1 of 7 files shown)


libs/data-service-generator/src/util/ast.ts
```diff
@@ -406,13 +406,10 @@ export function classDeclaration(
...
   }
-  const code = [
-    ...decorators.map((decorator) => recast.print(decorator).code),
-    recast.print(declaration).code,
-  ].join("\n");
-  const ast = parse(code);
-  const [classDeclaration] = ast.program.body as [namedTypes.ClassDeclaration];
-  return classDeclaration;
+
+  //@ts-ignore
```
overbit, line 410: Let's remove the ts-ignore here, please.

overbit (changes requested):

libs/data-service-generator/src/util/ast.ts
tupe12334: You cant because the types lib is not up to date

abrl91 (approved):

overbit (approved):

PR #4737, opened 2022-12-24
Dynamic plugins installation

tupe12334:
Close: https://github.com/amplication/amplication-infrastructure-next/issues/28

## PR Checklist

- [x] Tests for the changes have been added
- [x] `npm test` doesn't throw any error

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/amplication/amplication/blob/master/CODE_OF_CONDUCT.md) file for detailed contributing guidelines.

PR #4738, opened 2022-12-25
Change print to Amplication print

tupe12334:
This PR is part of https://github.com/amplication/amplication/pull/4737 we extract all the refactor to a separate PR

### DIFF
--- a/libs/data-service-generator/src/server/message-broker/topics-enum/createTopicsEnum.ts
+++ b/libs/data-service-generator/src/server/message-broker/topics-enum/createTopicsEnum.ts
@@ -5,7 +5,8 @@ import {
 } from "@amplication/code-gen-types";
 import { pascalCase } from "pascal-case";
 import { join } from "path";
-import { print, types } from "recast";
+import { types } from "recast";
+import { print } from "@amplication/code-gen-utils";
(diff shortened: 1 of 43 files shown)


libs/data-service-generator/src/server/message-broker/topics-enum/createTopicsEnum.ts
```diff
@@ -5,7 +5,8 @@ import {
 } from "@amplication/code-gen-types";
 import { pascalCase } from "pascal-case";
 import { join } from "path";
-import { print, types } from "recast";
+import { types } from "recast";
```
barshimi, line 8: why not types is exported from code gen types

libs/data-service-generator/src/util/ast.ts
```diff
@@ -5,6 +5,7 @@ import * as K from "ast-types/gen/kinds";
 import { NodePath } from "ast-types/lib/node-path";
 import { groupBy, mapValues, uniqBy } from "lodash";
 import * as recast from "recast";
```
barshimi, line 7: why that is still here ?

yuval-hazaz (approved):

Issue #4691, opened 2022-12-19
Enhance git library with support to commits, branch protections, create pull request from branch, etc


overbit:
In order to support a single branch strategy in the users repository, the git lib need to support additional functionality like:
```ts
  getRepository ( ) // That provided you also with the default branch

  commit (
    branch: string,
    message: string,
    files: { path: string; content: string }[],
    headCommit: string
  ) 

  createBranch(branch: string, headCommit: string)

  getBranch(branch: string): BranchMeta;

  createPullRequest (
    title: string,
    body: string,
    branch: string,
    base: string
  ): PullRequestMeta
  
  getPullRequest(branch: string, status: ): PullRequestMeta

  updateBranchProtection (branch: string)
  addPullRequestComment (
    pullRequestNumber: number,
    body: string
  )
```

Review matan poc branch `3751-epic-github-sync-amplication-branch`

Affected by: #4648 #4563

overbit:
@tupe12334 made a good point: Check if amplication has access to _update repository branch protection rules_.

PR #4741, opened 2022-12-26
Default branch 

tupe12334:
Partial: #4691

## PR Details

This PR adds get repository function
## PR Checklist

### DIFF
--- a/packages/amplication-git-utils/src/git/github.service.ts
+++ b/packages/amplication-git-utils/src/git/github.service.ts
@@ -90,6 +90,7 @@ export class GithubService {
       private: repo.private,
       fullName: repo.full_name,
       admin: repo.permissions.admin,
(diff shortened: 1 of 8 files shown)


packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -350,4 +352,33 @@ export class GithubService {
       })
     ).token;
   }
+
+  async getRepository(
```
abrl91, line 356: Why do we need this function? I assume we already have it somewhere

abrl91 (approved):

packages/amplication-git-utils/src/git/github.service.ts
tupe12334: Please have a look, I didn't find any

PR #4753, opened 2022-12-28
Add some null checks

tupe12334:
Partial: https://github.com/amplication/amplication/issues/4691

### DIFF
--- a/packages/amplication-git-pull-request-service/src/pull-request/dto/create-git-commit.dto.ts
+++ b/packages/amplication-git-pull-request-service/src/pull-request/dto/create-git-commit.dto.ts
@@ -3,7 +3,7 @@ import { IsOptional, IsString } from "class-validator";
 export class CreateGitCommit {
   @IsString()
   @IsOptional()
-  base?: string;
+  base?: string | undefined;
   @IsString()
   head!: string;
   @IsString()
--- a/packages/amplication-git-utils/src/git/git.service.ts
+++ b/packages/amplication-git-utils/src/git/git.service.ts
@@ -103,7 +103,7 @@ export class GitService {
     commitDescription: string,
     installationId: string,
     gitResourceMeta: GitResourceMeta,
-    baseBranchName?: string
+    baseBranchName?: string | undefined
   ): Promise<string> {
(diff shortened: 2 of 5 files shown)


packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -90,7 +90,7 @@ export class GithubService {
       url: repo.html_url,
       private: repo.private,
       fullName: repo.full_name,
-      admin: repo.permissions.admin,
+      admin: repo.permissions?.admin || false,
```
abrl91, line 93: why do we need the `|| false` part?

packages/amplication-git-pull-request-service/src/pull-request/dto/create-git-commit.dto.ts
```diff
@@ -3,7 +3,7 @@ import { IsOptional, IsString } from "class-validator";
 export class CreateGitCommit {
   @IsString()
   @IsOptional()
-  base?: string;
+  base?: string | undefined;
```
abrl91, line 6: Are you sure that the base could be `undefined`?
you must provide the base branch when you create a PR, and by default git takes the master/main branch

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -103,7 +103,7 @@ export class GitService {
     commitDescription: string,
     installationId: string,
     gitResourceMeta: GitResourceMeta,
-    baseBranchName?: string
+    baseBranchName?: string | undefined
```
abrl91, line 106: same as before regarding the baseBranch

packages/amplication-git-utils/src/git/github.service.ts
tupe12334: because permissions can be undefined so we need to provide bool

packages/amplication-git-pull-request-service/src/pull-request/dto/create-git-commit.dto.ts
tupe12334: Yes

morhag90 (approved):

Issue #4527, opened 2022-12-04
As User - I want to see a pricing page within the platform


GreenMachine01:
As User - I want to see a pricing page within the platform so that I can understand the differences between the plans and upgrade my workspace plan

Ability to initiate workspace upgrade via the pricing page
use Stigg.io for building the pricing page

**Design:**
https://www.figma.com/file/REUaAG2f5EkgyghHMvACxd/Concept----Paywall?node-id=836%3A18964&t=ndj7JrlNH4sm9vnH-1

barshimi:
do we have a design ? what is the scope of this task ?

GreenMachine01:
https://www.figma.com/file/REUaAG2f5EkgyghHMvACxd/Concept----Paywall?node-id=836%3A18964&t=ndj7JrlNH4sm9vnH-1

abrl91:
@EugeneTseitlin can you assign the PR to this issue?

PR #4766, opened 2023-01-01
pricing page inside our application

barshimi:
Close: #4527 

## PR Checklist


packages/amplication-design-system/src/components/LimitationDialog/LimitationDialog.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react";
```
barshimi, line 1: thus is not a design system component. it should be part the application. it is too specific

packages/amplication-server/src/core/project/project.service.spec.ts
```diff
@@ -148,6 +151,30 @@ const EXAMPLE_BLOCK_VERSION: BlockVersion = {
   displayName: EXAMPLE_BLOCK_DISPLAY_NAME,
 };
 
+const EXAMPLE_PROJECT_2: Project = {
```
barshimi, line 154: you can store new Date() once and use it for all cases

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +117,75 @@ export class ProjectService {
...
+    const workspace = project.workspace;
+    const subscriptions = workspace.subscriptions;
+    const workspaceServices = workspace.projects.flatMap(
+      (project) => project.resources
+    );
+    const projectServices = project.resources;
+
+    if (!subscriptions || subscriptions.length === 0) {
+      if (workspaceServices.length > LICENSE_SERVICES_PER_WORKSPACE_LIMIT) {
+        throw new Error(
```
barshimi, line 165: it is not an error only limitation of use. I would treat it as water flow response

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +117,75 @@ export class ProjectService {
...
+          "You have reached the maximum number of services. (Upgrade your workspace plan)"
+        );
+      }
+
+      projectServices.map((service) => {
+        if (service.entities.length > LICENSE_ENTITIES_PER_SERVICE_LIMIT) {
+          throw new Error(
```
barshimi, line 172: same as above

Issue #4574, opened 2022-12-07
As User - I want to be reminded about my workspace plan limits in the resources section and entities page


GreenMachine01:
Only for Community workspaces.

![Image](https://user-images.githubusercontent.com/112330500/206158789-03f76c6c-968a-40ed-aaa2-abce58ae5e73.png)

![Image](https://user-images.githubusercontent.com/112330500/206158946-b647964f-59ff-4a05-bbbf-82f955471809.png)

**Design:**
https://www.figma.com/file/REUaAG2f5EkgyghHMvACxd/Concept----Paywall?node-id=2027%3A5701&t=ndj7JrlNH4sm9vnH-1

Issue #4499, opened 2022-11-30
As User - I want to pay to upgrade the workspace plan


GreenMachine01:
As User - I want to pay to upgrade the workspace plan so that I have no limitations

We want to use an out of the box solution for managing pricing pages and plans by Stigg.
Stigg uses Stripe for the purchases.

https://www.stigg.io/
https://stripe.com/

**Definition of Done:**
- Users are able to pay for a workspace plan upgrade via Stripe/Stigg
- Users see a thank you message after payment is successful
- the upgraded workspace is updated to show its updated plan status and limitations are updated

**Design:**

![Image](https://user-images.githubusercontent.com/112330500/207034222-718e0999-e068-4cce-918c-c2fe9828a313.png)

https://www.figma.com/file/REUaAG2f5EkgyghHMvACxd/Concept----Paywall?node-id=836%3A18964&t=NwKtX8o6ci3tqPpx-1

PR #4767, opened 2023-01-01
Epic/paywall

EugeneTseitlin:
Close: #4527
Close: #4499
Close: #4574

## PR Checklist


packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -114,13 +120,87 @@ export class ProjectService {
     return [...changedEntities, ...changedBlocks];
   }
 
+  async validateSubscriptionPlanLimitationsForProject(
```
morhag90, line 123: let's talk about this query

morhag90 (changes requested):

packages/amplication-client/src/Workspaces/ResourceList.tsx
```diff
@@ -71,6 +77,12 @@ function ResourceList() {
     [deleteResource, setError, trackEvent]
   );
 
+  const { data: getWorkspaceData } = useQuery<GetWorkspaceResponse>(
+    GET_CURRENT_WORKSPACE
+  );
+  const subscription =
```
yuval-hazaz, line 83: instead of using a query, we should be able to get the data from appContext

packages/amplication-server/src/core/billing/billing.service.ts
```diff
@@ -0,0 +1,155 @@
...
+    private readonly logger: AmplicationLogger,
+    configService: ConfigService
+  ) {
+    const stiggApiKey = configService.get(Env.BILLING_API_KEY);
+    this.stiggClient = Stigg.initialize({ apiKey: stiggApiKey });
+  }
+
+  async getStiggClient() {
+    try {
+      await this.stiggClient.waitForInitialization();
```
morhag90, line 37: Check if there is a way to make this call as a single tone

yuval-hazaz (dismissed):

morhag90 (approved):

PR #4784, opened 2023-01-02
Add logs and data for the git pull request

tupe12334:
This PR adds loggs data for the git pull request flow

### DIFF
--- a/packages/amplication-git-utils/src/git/github.service.ts
+++ b/packages/amplication-git-utils/src/git/github.service.ts
@@ -478,6 +478,11 @@ export class GithubService {
       repo,
       branchName
     );
+
+    if (changesArray.length === 0) {
+      return;
+    }
+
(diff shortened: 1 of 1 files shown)


packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -486,19 +491,27 @@ export class GithubService {
       // @ts-ignore
       tree: changesArray,
     });
+
+    console.log(`Created tree for for ${owner}/${repo}`);
```
arielweinberger, line 495: Please use a logger. This `console.log` usage floods the logs and bumps up bills on AWS CloudWatch and previously on Logz.io because we cannot shut them off in certain environments.

Also, logs like this should be structured.

arielweinberger (changes requested):

morhag90 (approved):

arielweinberger (approved):

Issue #4857, opened 2023-01-08
🐛 Bug Report: `.gitignore` files in the generated admin-ui and server are missing


abrl91:
### What happened?

`.gitignore` files in the generated admin-ui and server are missing

### What you expected to happen

I expect to see a `.gitignore` file in the root admin and server directories

### How to reproduce

create a service in a project, go to view code/click on build, see that there is no `.gitignore` file in the server and admin-ui

### Amplication version

1.2.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #4861, opened 2023-01-08
Fix  copy `.gitignore` in the generated admin and server

abrl91:
Close: #4857 

## PR Details

Convert the `.gitignore` file from a static file to a template file and generate it with plugin-wrapper

[npm: server](https://www.npmjs.com/package/@amplication/data-service-generator?activeTab=explore)
[npm: admin](https://www.npmjs.com/package/@amplication/data-service-generator?activeTab=explore)

## PR Checklist


libs/data-service-generator/src/admin/create-admin.ts
```diff
@@ -122,12 +127,21 @@ async function createAdminModulesInternal(): Promise<Module[]> {
...
-    ...module,
-    code: formatCode(module.code),
-  }));
+  const formattedModules = createdModules.map((module) => {
+    if (module.path.includes(".gitignore")) {
+      return {
+        ...module,
+        code: module.code,
+      };
+    }
```
yuval-hazaz, line 136: Is that a must? Is there any other way to do that? .e.g can we tell the formatter to ignore unknown files?

libs/data-service-generator/src/admin/gitignore/create-gitignore.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export function createGitIgnore(
+  eventParams: CreateAdminGitIgnoreParams = {}
+): Module[] {
+  return pluginWrapper(
+    createGitIgnoreInternal,
+    EventNames.CreateAdminGitIgnore,
+    eventParams
+  );
+}
```
yuval-hazaz, line 20: I think we should export the content of the file as a json object and allow to manipulate the file in the event
I would expect to see a single CreateGitIgnore for both the server and the client, with the same logic and a different data object for initialization

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #4624, opened 2022-12-12
As Dev - I want to add meta data and plugins version to dsg logs


tupe12334:
In the DSG running log we can add a list of all the used plugins and there versions

yuval-hazaz:
that is a good idea! 
@GreenMachine01 please add to the backlog

arielweinberger:
I cannot see these logs with plgin versions.

In this example I have two plugins - PostgreSQL and Prettier. Tested on Next.



![Image](https://user-images.githubusercontent.com/4976416/212938742-6cff0c10-823f-436c-8549-75898bdf3b2b.png)

overbit:
# QA passed

<img width="1018" alt="image" src="https://user-images.githubusercontent.com/2861984/214791814-2696f182-b597-4f5d-b490-11d0a5a4ed5e.png">

PR #4894, opened 2023-01-10
Add plugins meta in DSG logs

tupe12334:
Close: #4624 

## PR Details

This PR adds logs for the DSG to write plugins meta data
## PR Checklist


libs/data-service-generator/src/prepare-context.ts
```diff
@@ -63,6 +64,22 @@ export async function prepareContext(
   }
 
   const pluginsWithDefaultPlugins = prepareDefaultPlugins(resourcePlugins);
+
+  await createLog({
+    level: "info",
+    message: `The code generation process working with the plugins:`,
+  });
```
yuval-hazaz, line 71: I don't understand this message

libs/data-service-generator/src/prepare-context.ts
```diff
@@ -63,6 +64,22 @@ export async function prepareContext(
...
+  for (const plugin of pluginsWithDefaultPlugins) {
+    try {
+      await createLog({
+        level: "info",
+        message: `   ${plugin.npm}@${plugin.version}`,
+      });
+    } catch (error) {
+      logger.error(`Error to log plugin ${plugin}`);
+    }
+  }
```
yuval-hazaz, line 81: this is not the right place for the logs. I would expect to see the log in the actual place where the plugins are being used, including an option to log failure to install. 

Something like this (very high level of course) 

try
CreateLog("Installing Plugin" + plugin)
InstallPlugin(plugin)
CreateLog("Successfully Installed plugin" + plugin)
catch
CreateLog("Failed to installed plugin" + plugin + ex)
end try

yuval-hazaz (changes requested):

packages/amplication-data-service-generator-runner/src/utils/DynamicPackageInstallationManager.ts
```diff
@@ -1,27 +1,42 @@
...
 
-  async installMany(packages: PackageInstallation[]): Promise<void> {
-    for (const plugin of packages) {
-      await this.install(plugin);
+  async install(
+    plugin: PackageInstallation,
+    hooks: {
+      before?: HookFunction;
+      after?: HookFunction;
+      crash?: HookFunction;
```
yuval-hazaz, line 12: Is this naming conventional? If not, I think we should use something more descriptive like "onBeforeInstall", "onAfterInstall", "onError"

packages/amplication-data-service-generator-runner/src/dynamic-package-installation.ts
```diff
@@ -11,9 +15,32 @@ export async function dynamicPackagesInstallations(
...
+      after: async (plugin) => {
+        await createLog({
+          level: "info",
+          message: `Successfully Installed plugin: ${plugin.name}@${plugin.version}`,
+        });
+      },
+      crash: async (plugin) => {
+        await createLog({
+          level: "error",
+          message: `Failed to installed plugin: ${plugin.name}@${plugin.version}`,
```
yuval-hazaz, line 39: nice

yuval-hazaz (approved):
approving the code, but please have a look at my question about the callback naming

Issue #4214, opened 2022-10-20
🐛 Bug Report: Pending Changes doesn't update after deleting a Message Broker


GreenMachine01:
**Steps to Reproduce:**
1. Create a new Message Broker
2. Create a topic within the message broker
3. Delete the Message Broker via the Project Settings Page

**Expected:**
The Pending Changes area is updated and no longer shows the Message Broker Section

**Actual:**
The Message Broker with the created Topic are still shown in the Pending Changes area

abrl91:
I think we already have this issue: https://github.com/amplication/amplication/issues/4154

GreenMachine01:
@abrl91 thanks. yes, they are partially duplicate of one another.
I'll remove Scenario #2

arielweinberger:
Confirming this has been fixed.

Issue #4334, opened 2022-11-08
🐛Bug Report: 'Discard Changes' action doesn't cancel new entities or plugins installed


GreenMachine01:
**Steps to reproduce:**
1. Create a new service
2. Create several new entities
3. Install several plugins
4. Click on the discard changes CTA

**Expected:**

1. New entities are deleted
2. Plugins show as not installed

**Actual:**

1. New entities remain on the list
2. Plugins are shown as "installed"

arielweinberger:
The specs described in this issue have been tested and the solution works properly. I found a different issue while testing this, for which I have created the following task: #4990

Issue #4154, opened 2022-10-14
🐛 Bug Report:  Pending changes aren't updated when renaming the topic name


abrl91:
### What happened?

 Pending changes aren't updated when renaming the topic name

https://user-images.githubusercontent.com/39680385/195817290-762fc124-74b7-4ec0-92cd-8e4285acd949.mov

### What you expected to happen

I expect that the topic name on the pending changes will be updated every time the user renames the topic name

### How to reproduce

see the video

### Amplication version

1.0.2

MuhammadJufri:
I would like to work on this if it's available, please

abrl91:
> I would like to work on this if it's available, please

I assigned you. Good luck :)

MuhammadJufri:
Thanks!

abrl91:
@SayyidMuhammad878 Did you manage to make some progress with this issue?

SayyidMuhammad878:
Yes, I was busy I getting a new job I will look after it!

SayyidMuhammad878:
I still working on it but I've completed about 80-90%
Maybe I will push it today or tomorrow

MAX-786:
hey @abrl91, if no one is working on this issue I would like to work on it, can you assign it to me? Thanks :)

arielweinberger:
@SayyidMuhammad878 What do you think? Are we good to wrap this up, or can I assign it to @MAX-786?

arielweinberger:
@MAX-786 I have assinged it to you.

MAX-786:
Thanks @arielweinberger , I will try to create a PR soon

SayyidMuhammad878:
> @MAX-786 I have assinged it to you.

Sorry for late answer, Sure you can assign to MAX-786.
But I have completed 99% only I need to know is which file that invoke the TopicPage file. If there is a file called Abc there is must be another file that called Abc something like this <Abc />

MAX-786:
@arielweinberger if it's not a bother, you can re-assign this issue to @SayyidMuhammad878 since I haven't started working on it. And also while exploring I encountered this design improvement at `/user/profile`  which is :
Before improving,
![Screenshot (28)](https://user-images.githubusercontent.com/99530996/198280664-1c09ad50-6fb2-4ebc-aac7-b16f38f218b2.png)
After improving which I think looks good now,
![Screenshot (27)](https://user-images.githubusercontent.com/99530996/198280895-f181383a-3dca-4c62-b85b-654fa097d775.png)
[...]

Issue #4017, opened 2022-10-02
🐛 Bug Report: Pending Changes doesn't update after deleting a Service


EugeneTseitlin:
### What happened?

There are pending changes that should not exist:
![image](https://user-images.githubusercontent.com/6080188/193447867-4b6d3c8d-dad0-4ed4-8ed8-e925a7b78dd8.png)

### What you expected to happen

Create service
Remove service
See that pending changes still applied for the project

### How to reproduce

v1.0.1

saloni137:
@EugeneTseitlin I would like to work on this issue.

GreenMachine01:
@saloni137 Assigned to you. Good Luck!

souravjain540:
any updates @saloni137

GreenMachine01:
@souravjain540 I see no progress with the item so allowing new contributors to take it on.

yuval-hazaz:
@GreenMachine01 I think that @abrl91 started to work on something in this area before and may have some insight on the depth of efforts that are needed. 
I think we should keep this item internally until we have a clear definition of the proposed solution (it should be more then just hiding the items, we need to have a server-side behavior that will close the pending changes or other things around it)

abrl91:
@GreenMachine01 @yuval-hazaz 
I Agree. The "issue" with the pending changes has come up several times already in different areas (message broker, roles, project configuration, etc.) and we have to define the desired behavior for each

arielweinberger:
Tested on Next and seems to be fixed.

abrl91:
fixed by https://github.com/amplication/amplication/pull/4928

PR #4928, opened 2023-01-12
Fix/4334 discard changes

morhag90:
Close: #4214 #4334 #4017 #4154

## PR Details

Fix sync issues on discard changes component and pending changes. 
Scopes: 

- Service
- Plugins
- Message broker and topics. 

test cases: 

- Create new entity and install new plugin => discard changes. 
- Create new service => commit changes => make change in service => delete service. 
- Create new message broker =>  commit changes => create new topic => discard changes. 
- 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/context/appContext.tsx
+++ b/packages/amplication-client/src/context/appContext.tsx
@@ -31,7 +31,7 @@ export interface AppContextInterface {
   pendingChanges: PendingChangeItem[];
   commitRunning: boolean;
   pendingChangesIsError: boolean;
-  addEntity: (entityId: string) => void;
+  addEntity: (entityId?: string) => void;
   addBlock: (blockId: string) => void;
   addChange: (originId: string) => void;
   resetPendingChanges: () => void;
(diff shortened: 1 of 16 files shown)


packages/amplication-client/src/context/appContext.tsx
```diff
@@ -31,7 +31,7 @@ export interface AppContextInterface {
   pendingChanges: PendingChangeItem[];
   commitRunning: boolean;
   pendingChangesIsError: boolean;
-  addEntity: (entityId: string) => void;
+  addEntity: (entityId?: string) => void;
```
tupe12334, line 34: Why we changed that to optional?

yuval-hazaz: you can see above that they call addEntity when a resource is deleted. 
AddEntity and addBlock are actually just a trigger to refresh the list from the server. 
Maybe we can refactor the code and change the names of the functions to something like regreshPendingChanges() but I am OK with the current change at this phase. 

If someone thinks the refactor is needed, you can open a separate issue to be implemented in the future

yuval-hazaz (approved):

packages/amplication-client/src/context/appContext.tsx
morhag90: @yuval-hazaz I'm merging this pr and opening an issue for renaming the function name

Issue #4939, opened 2023-01-13
🐛 Bug Report: Can't run server in local dev env when billing is not configured


yuval-hazaz:
### What happened?

When I run my local env without configuration stigg I get an error 
I tried to set the env variable `BILLING_ENABLED=false` but it didn't help

<img width="1120" alt="image" src="https://user-images.githubusercontent.com/43705455/212392413-28a109f7-d517-4755-b325-72f9af626839.png">

### What you expected to happen

The server should run successfully without billing enabled 

### How to reproduce

run the server locally using `npx nx serve amplication-server`

### Amplication version

1.2.2

abrl91:
Fixed by https://github.com/amplication/amplication/pull/4941

PR #4941, opened 2023-01-13
fix server in local dev env when billing is not configured

yuval-hazaz:
Close: #4939 

## PR Details
- [x] Added exception handling in the initialization of the billing service
- [x] Disable the billing service and allow the server to keep running in case of an error in Stigg client
- [x] move all the logic about `isBillingDisabled` into the billing service, so other services don't need to know anything about it (except when specifically needed)
- [x] Move logic about license and billing from the project service to the billing service

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/billing/billing.service.ts
+++ b/packages/amplication-server/src/core/billing/billing.service.ts
@@ -35,9 +35,16 @@ export class BillingService {
     private readonly analytics: SegmentAnalyticsService,
     configService: ConfigService
   ) {
-    const stiggApiKey = configService.get(Env.BILLING_API_KEY);
-    this.stiggClient = Stigg.initialize({ apiKey: stiggApiKey });
-    this.clientHost = configService.get(Env.CLIENT_HOST);
+    const isBillingEnabled = Boolean(
+      configService.get<string>(Env.BILLING_ENABLED) === "true"
+    );
+
+    if (isBillingEnabled) {
+      const stiggApiKey = configService.get(Env.BILLING_API_KEY);
+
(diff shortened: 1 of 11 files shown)


packages/amplication-server/src/core/billing/billing.service.ts
```diff
@@ -35,9 +35,16 @@ export class BillingService {
     private readonly analytics: SegmentAnalyticsService,
     configService: ConfigService
   ) {
-    const stiggApiKey = configService.get(Env.BILLING_API_KEY);
-    this.stiggClient = Stigg.initialize({ apiKey: stiggApiKey });
-    this.clientHost = configService.get(Env.CLIENT_HOST);
+    const isBillingEnabled = Boolean(
+      configService.get<string>(Env.BILLING_ENABLED) === "true"
+    );
```
yuval-hazaz, line 40: @EugeneTseitlin I tried to use boolean but it didn't work. I would love to get your insight 
`this.configService.get<boolean>Env.BILLING_ENABLED);`

overbit, line 40: how about leveraging the generic type instead of Boolean?

```suggestion
    const isBillingEnabled = configService.get<boolean>(Env.BILLING_ENABLED);
```

overbit (changes requested):

packages/amplication-server/src/core/billing/billing.service.ts
yuval-hazaz: Look at my previous comment above - I tried it in several ways with several different values, and it didn't work for me. would be happy to understand why

EugeneTseitlin (approved):

Issue #4961, opened 2023-01-15
🐛 Bug Report: Text fields look wrong throughout the system


yuval-hazaz:
### What happened?

Following this PR https://github.com/amplication/amplication/pull/4756/files some text fields look wrong 
<img width="1018" alt="image" src="https://user-images.githubusercontent.com/43705455/212559440-18eed513-979b-476e-a364-d1daed6b1c94.png">

### What you expected to happen

should look as designed 

### How to reproduce

n/a

pratik-codes:
If no one is working on this can you assign this to me ill take a look

sagnik3788:
is this issue solved @souravjain540 if not can i work on it

PR #4756, opened 2022-12-28
[BUG FIX] Project name error

Aadarsh805:
Close: #4652

## PR Details

added `width: fit-content` to a div to fix the error icon going out of project-name input field
`src/components/TextInput/TextInput.scss`
image

Before this change: 
Screenshot from 2022-12-28 23-20-13

After this change: 
Screenshot from 2022-12-28 23-20-19

## PR Checklist

- [ x] Tests for the changes have been added
- [ x] `npm test` doesn't throw any error

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/amplication/amplication/blob/master/CODE_OF_CONDUCT.md) file for detailed contributing guidelines.

Issue #4652, opened 2022-12-14
🐛 Bug Report: Wrong design for project name error


GreenMachine01:
In the settings page of a project when you get an error the position of the 'i' is wrong.
The 'i' should be within the input field

**Actual:**

![Image](https://user-images.githubusercontent.com/112330500/207645463-4f83e47d-ef01-4a41-8c92-64424ae33cfa.png)

**Expected:**

![Image](https://user-images.githubusercontent.com/112330500/207645632-28eb7c48-96ef-4b9f-af06-c5e1f87c7ea5.png)

**Plan:**
Make sure that TextField component is used with proper prperties. 
Make sure TextField uses styles only from design system for it's own implementation. Only positioning in a layout should be in amplication client project.
Remove/change styles in client project that affects TextField implementation.
Verify that TextField looks the same in every place where it used.

Aadarsh805:
I'm working on it

yuval-hazaz:
Reopen the issue following #4961

GreenMachine01:
@sajdakabir assigned to you. good luck!

gitstart:
@GreenMachine01 There is a PR here [https://github.com/amplication/amplication/pull/5091](https://github.com/amplication/amplication/pull/5091). You can have a look at it. Thanks

sajdakabir:
@GreenMachine01 can you please review my PR

PR #4962, opened 2023-01-15
fix(client) textbox width

yuval-hazaz:
Close: #4961 

## PR Details

Following incorrect fix in https://github.com/amplication/amplication/pull/4756, we revert the fix and reopen https://github.com/amplication/amplication/issues/4652

## PR Checklist

### DIFF
--- a/packages/amplication-design-system/src/components/TextInput/TextInput.scss
+++ b/packages/amplication-design-system/src/components/TextInput/TextInput.scss
@@ -2,6 +2,7 @@
 
 .text-input {
   position: relative;
+  max-width: var(--form-field-long);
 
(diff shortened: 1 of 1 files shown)


Aadarsh805 (commented):
we can have the `fit-content` for that particular input field only the one in this issue #4652 , changing it from here one is affecting others too

tupe12334 (approved):

yuval-hazaz:
Thanks @Aadarsh805 
I saw the issue repeats in all similar textboxes - so we need to think about a generic solution that will cover all cases but will not have a negative impact

If you want to work on it, I can assign it to you

Issue #4979, opened 2023-01-16
🐛 Bug Report: Commits Page response time is too long


yuval-hazaz:
### What happened?

When running a build (clicking on Rebuild of Commit) it takes a lot of time foe the commits page to load. 
it also very not responsive when navigating from the commits page into a build page and back

In the network monitor we can see multiple requests, including duplicate requests to "Commits", "LastCommit", and "Pending
Changes"
<img width="1512" alt="image" src="https://user-images.githubusercontent.com/43705455/212752368-be1a97a9-c78e-424e-9d6f-1eb9f73d2ca6.png">

### What you expected to happen

Improve the response time of the page and make sure we only use a single request for each data set

### How to reproduce

Click on "Rebuild" or "Commit"

PR #4980, opened 2023-01-16
fix(client): commits page response time

yuval-hazaz:
Close: #4979

## PR Details

 - [x] Remove the query for last commit and instead return the top commit from the commit list in `UseCommit` hook
 - [x] After commit:
   - [x] Use Refetch from the `UseCommit` hook instead of refetching the query directly 
   - [x] Remove the refetch for LastCommit
   - [x] Remove redundant call to `resetPendingChanges`
 - [x] Do not hide the commit list when reloading more commits 
 - [x] Remove redundant refetch of the commit list in the `useEffect`
 - [x] remove `CommitList.scss` which was not in use

Post fix, we only see 3 request when clicking on "Commit" (instead of 9 requests before)

1 - to perform the commit()
2- to reload the pendingChanges()
3- to reload the commits() list

image

## PR Checklist


overbit (approved):
I've also QAed this locally and works nicely.
Once merged I can perform a quick test on sandbox environment too.

PR #4981, opened 2023-01-17
feat(server): add ability to delete project, fixes #4044

### LINKED ISSUE #4044: 🚀 Feature: Allow users to delete a project
### Feature description

- Add a new delete button on the Project Configuration item

image

- When clicked, show the user a warning with a confirmation button "This action cannot be undone. This will permanently delete the project and all its resources. Are you sure you want to continue?"

image

- On the server side, use soft delete (there is already a "deletedAt" field). Copy the behavior from resource.service.ts

### Use case

The user wants to delete a project

_No response_

MichaelSolati:
Close: #4044

## PR Details

- Add a mutation to the project resolver for deleting a project
- Add a method to project service to delete a project
  - First archives all resources belonging to a project
  - Update billing service based on archived services
  - Next "deletes" project by renaming it and setting `deletedAt` field
- Add `archiveProjectResources` method to the resource service
  - Gets all resources belonging to a project and then checks if they're empty
  - Batches mutations of resources and marks them as "archived"
  - Takes batch and runs it as a transaction (trying to limit Prisma calls)
- Update UI to call `deleteProject` mutation

## PR Checklist


MichaelSolati:
Sorry to add so many of y'all for the review, I'm just unsure who's responsible for these files/services/etc

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -426,6 +426,67 @@ export class ResourceService {
...
+      })
+    );
+    const deletedResources = await this.prisma.$transaction(deleteResources);
+
+    if (this.isBillingEnabled) {
+      await this.billingService.reportUsage(
+        workspace.id,
+        BillingFeature.Services,
+        deletedServiceCount
+      );
```
yuval-hazaz, line 475: we may need to keep this code anyway even if we not actually deleting the resources, so maybe you could move it to the "deleteProject" function

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -426,6 +426,67 @@ export class ResourceService {
...
+    }
+
+    await Promise.all(
+      resources.reduce((all: Promise<Resource>[], resource) => {
+        if (resource.gitRepositoryOverride) {
+          all.push(this.deleteResourceGitRepository(resource));
+        }
+        return all;
+      }, [])
+    );
```
yuval-hazaz, line 485: we should keep this data also as it is (for future undo)

yuval-hazaz (changes requested):
well done @MichaelSolati 
It is great to see how you followed all the best practices and navigated your way on the code. 

Please see my comments regarding leaving the resources in place

MichaelSolati:
@yuval-hazaz just to secure a solution to the resources. Should we block resources from being queried/modified by making checking the status of the parent project, or should we have a status field on a resource as to what the parent's status is?

I'm inclined to prefer adding a status field to the resource, we could just call it `parentsStatus` and allow it to be `null` or `'deleted'`. Down the road we can add new statuses such as `archived'`.

Let me know what you think.

yuval-hazaz:
> I'm inclined to prefer adding a status field to the resource, we could just call it `parentsStatus` and allow it to be `null` or `'deleted'`. Down the road we can add new statuses such as `archived'`.

I am OK with adding a new field, but in this case, I prefer to keep it simple and use a boolean with a specific definition - it will hold a boolean value whether the record is effectively active or not. It should hold "false" whether the actual record is deleted or any of its parents is deleted. 

When fetching the data, we should only check for that value. 

we can call the field "EffectiveActive"

abrl91:
@MichaelSolati, any progress?

MichaelSolati:
Working on this today. Should be done in the next day or so.

MichaelSolati:
I am finished making the requested changes. I don't think the name `effectiveActive` is that clear as to what it represents; I'm wondering if we can rename it to `archived`?

yuval-hazaz:
> I am finished making the requested changes. I don't think the name `effectiveActive` is that clear as to what it represents; I'm wondering if we can rename it to `archived`?

I think "archived" can work

MichaelSolati:
@yuval-hazaz ready for review

dulajdilshan (commented):
LGTM

MichaelSolati:
@abrl91 @yuval-hazaz @tupe12334 

Can someone review this please.

abrl91:
It looks good, I'm just wondering what happens when all projects are deleted, I mean, what will happen to the UI?

image

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -72,6 +75,39 @@ export class ProjectService {
...
+      throw new Error(INVALID_PROJECT_ID);
+    }
+
+    const archivedResources =
+      await this.resourceService.archiveProjectResources(project.id);
+    const archivedServiceCount = archivedResources.filter(
+      (r) => r.resourceType === EnumResourceType.Service
+    ).length;
+
+    if (this.billingService.isBillingEnabled) {
```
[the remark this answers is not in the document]
MichaelSolati: Beautiful, I'll remove this check then.

packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
```diff
@@ -71,6 +71,7 @@ async function main() {
     where: {
       resourceType: EnumResourceType.Service,
       deletedAt: null,
+      archived: { not: true },
```
abrl91, line 74: We already ran this migration script on the DB, so I don't this is relevant.

abrl91:
@MichaelSolati, generally, it looks good to me, but I wonder if we want to merge it before fixing the UI issue in case a user deletes all the projects.

As for your question, I am trying to understand. Can you create an issue with more details/examples?

abrl91 (approved):

yuval-hazaz (approved):

Issue #1215, opened 2021-02-06
DSG: One-to-one - select the "leading" side


yuval-hazaz:
**Is your feature request related to a problem? Please describe.**
When creating a one-to-one relation only one side should have a foreign key, while the other side is just used for navigation.
Currently, when DSG automatically selects the side to hold the FK.
Instead, the user should select on which side of the relation to place the FK.

example:
below is an example of a one-to-one relation between a user and a profile. 
In this example, the FK is placed on the user model, while the profile model only has a navigation property back to the user without a FK.

model User
- id
- profile
- profileId
- 
model profile
- id
- user

**Describe the solution you'd like**
Allow the user to select on which side of the relation to place the FK\

**Design** - https://www.figma.com/file/g56PMEpymt2I3Ecrtb3LH5/%231215---One-to-one%E2%80%A8select-the-%22leading%22-side?node-id=2041%3A14924&t=qqCGD9kMWLmQQZ33-1

**Additional context**
see this article from Prisma with more details on how they expect the schema to be created
https://www.prisma.io/docs/concepts/components/prisma-schema/relations#one-to-one-relations

yenicelik:
I think this is related to this issue: https://github.com/amplication/amplication/issues/1387

yuval-hazaz:
When creating a one-to-many relation - it is clear that the foreign key is on the "many" side. 
When creating a many-to-many relation - Prisma automatically generated a connection table with foreign keys to both sides.
When creating a one-to-one relation - currently Amplication chooses randomly one side to hole the FK. This FR is to allow the developer to set that manually. 

In most cases, this behavior does not impact the functionality at all and the developer is not even aware of it.

yuval-hazaz:
@abrl91 have a look at this one 😃
it is open since Feb 21... I added it now to the backlog

PR #4989, opened 2023-01-17
Feat: Add option one to one relation in Entity Field

morhag90:
Close: #1215

## PR Details

Add the option to choose the side for the FK in a one-to-one relation field. 

## PR Checklist


libs/data-service-generator/src/prepare-context.ts
```diff
@@ -212,8 +212,9 @@ function resolveLookupFields(entities: Entity[]): Entity[] {
           //**@todo: in one-to-one relation, only one side should have a foreign key.
           //We currently decide randomly based on sorting the permanent ID
           //instead we should let the user decide which side holds the foreign key  */
+          //
           const isOneToOneWithoutForeignKey =
-            isOneToOne && field.permanentId > relatedField.permanentId;
+            isOneToOne && field.permanentId !== fieldProperties.FkHolder;
```
yuval-hazaz, line 217: you should keep the current behavior in case the new property is undefined or null

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,66 @@
...
+      !relatedField.properties.allowMultipleSelection
+    );
+  }, [data, formik.values]);
+
+  const entityListOptions = useMemo(() => {
+    return (
+      (entity &&
+        data && [
+          {
+            value: currentEntity.permanentId,
```
yuval-hazaz, line 49: This is not correct, you should use the value from formik with the current field

packages/amplication-prisma-db/.env
```diff
@@ -4,3 +4,4 @@ POSTGRESQL_URL=postgresql://admin:admin@localhost:5432/${SERVICE_DB_NAME}
 POSTGRESQL_USER=admin
 POSTGRESQL_PASSWORD=admin
 SERVICE_DB_NAME=amplication
+PRISMA_CLIENT_OUTPUT_PATH= generated-prisma-client
```
yuval-hazaz, line 7: is that related to this PR?

yuval-hazaz (changes requested):

packages/amplication-prisma-db/.env
morhag90: @yuval-hazaz  No, it's a new env var that is missing in the env file

libs/data-service-generator/src/prepare-context.ts
```diff
@@ -209,11 +209,9 @@ function resolveLookupFields(entities: Entity[]): Entity[] {
             !fieldProperties.allowMultipleSelection &&
             !relatedFieldProperties.allowMultipleSelection;
 
-          //**@todo: in one-to-one relation, only one side should have a foreign key.
-          //We currently decide randomly based on sorting the permanent ID
-          //instead we should let the user decide which side holds the foreign key  */
           const isOneToOneWithoutForeignKey =
-            isOneToOne && field.permanentId > relatedField.permanentId;
+            (isOneToOne && field.permanentId > relatedField.permanentId) ||
```
yuval-hazaz, line 213: We should first check if the new property is populated and only use the previous condition in case it is not

packages/amplication-prisma-db/.env
yuval-hazaz: why not merging the missing variable into this branch?

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -2026,7 +2028,9 @@ export class EntityService {
     entityId: string,
     relatedEntityId: string,
     relatedFieldId: string,
-    user: User
+    user: User,
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    FkHolder?: string
```
yuval-hazaz, line 2033: rename to fkHolder

yuval-hazaz (changes requested):
@morhag90 in addition to what I commented inline, I have doubts about the way we delete and recreate the related field - it is not part of this PR, but I think we should look at it anyway

packages/amplication-prisma-db/.env
morhag90: @yuval-hazaz, that's what I did, no??....

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { GET_ENTITY_FIELD_BY_PERMANENT_ID } from "./RelatedEntityFieldField";
+
+type Props = Omit<SelectFieldProps, "options"> & {
+  entity: models.Entity;
+};
+
+const RelationFkHolderField = ({ entity, ...props }: Props) => {
+  const formik = useFormikContext<models.EntityField>();
+
+  const { data } = useQuery<{ entity: models.Entity }>(
```
barshimi, line 15: move it to query file and also add `skip` so it wont run every time.

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,63 @@
...
+    GET_ENTITY_FIELD_BY_PERMANENT_ID,
+    {
+      variables: {
+        entityId: formik.values.properties.relatedEntityId,
+        fieldPermanentId: formik.values.properties.relatedFieldId,
+      },
+    }
+  );
+
+  const relatedField =
```
barshimi, line 25: add state for that and listen to changes in data

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,63 @@
...
+    data &&
+    data.entity &&
+    data.entity.fields &&
+    data.entity.fields.length &&
+    data.entity.fields[0];
+
+  const isOneToOne = useMemo(() => {
```
barshimi, line 32: do you want to memo the result of this ? if not it should also be part of the state

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      entity &&
+      data &&
+      !formik.values?.properties?.allowMultipleSelection &&
+      !relatedField.properties?.allowMultipleSelection
+    );
+  }, [data, formik.values]);
+
+  const entityListOptions = useMemo(() => {
```
barshimi, line 41: this is also problematic. this function should get the data as argument

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
yuval-hazaz: we better use useMemo instead of state

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
yuval-hazaz: useMemo is better in this case

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
yuval-hazaz: why?

packages/amplication-prisma-db/.env
yuval-hazaz: please use a different PR

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -2175,13 +2182,40 @@ export class EntityService {
...
+        )?.fkHolder;
+
+        await this.prisma.entityField.update({
+          where: {
+            id: relatedField.id,
+          },
+          data: {
+            properties: relatedFieldProps as unknown as Prisma.InputJsonValue,
+          },
+        });
```
yuval-hazaz, line 2216: only if (field.dataType === EnumDataType.Lookup && properties.fkHolder !== null)

libs/data-service-generator/src/prepare-context.ts
```diff
@@ -209,11 +209,9 @@ function resolveLookupFields(entities: Entity[]): Entity[] {
             !fieldProperties.allowMultipleSelection &&
             !relatedFieldProperties.allowMultipleSelection;
 
-          //**@todo: in one-to-one relation, only one side should have a foreign key.
-          //We currently decide randomly based on sorting the permanent ID
-          //instead we should let the user decide which side holds the foreign key  */
           const isOneToOneWithoutForeignKey =
-            isOneToOne && field.permanentId > relatedField.permanentId;
+            (isOneToOne && field.permanentId !== fieldProperties.fkHolder) ||
+            (isOneToOne && field.permanentId > relatedField.permanentId);
```
yuval-hazaz, line 214: separate conditions for when fieldProperties.fkHolder is null (or undefined) and not null

yuval-hazaz (changes requested):

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -33,8 +33,8 @@ const RelationFkHolderField = ({ entity, ...props }: Props) => {
     return (
       entity &&
       data &&
-      !formik.values?.properties.allowMultipleSelection &&
-      !relatedField.properties.allowMultipleSelection
+      !formik.values?.properties?.allowMultipleSelection &&
+      !relatedField.properties?.allowMultipleSelection
```
abrl91, line 37: Why do we now need the `?` after `properties`?

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,80 @@
...
+    if (!data) {
+      entityFieldRef.current = {
+        relatedField: undefined,
+        isOneToOne: false,
+        entityListOptions: [],
+      };
+      return;
+    }
+
+    console.log("RelationFkHolderField", formik.values);
```
yuval-hazaz, line 47: remove console

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
```diff
@@ -0,0 +1,80 @@
...
+        entityId: formik.values.properties.relatedEntityId,
+        fieldPermanentId: formik.values.properties.relatedFieldId,
+      },
+      skip:
+        !formik.values.properties?.relatedEntityId ||
+        !formik.values.properties?.relatedFieldId,
+    }
+  );
+
+  useEffect(() => {
```
yuval-hazaz, line 37: I understand what you did here, but I think the useMemo version is much more readable and straight-forward - don't see the advantage of using the ref object. Also, useMemo is consistent with the way we do it across the client app
I don't ask you to change, but I just want to have a discussion about the best practice

yuval-hazaz (changes requested):

packages/amplication-client/src/Entity/RelationFkHolderField.tsx
morhag90: @abrl91 because they init will null values, and this component is triggered.

yuval-hazaz (approved):

Issue #4994, opened 2023-01-18
Removal of the general DSG path from the main tsconfig


tupe12334:

GreenMachine01:
@tupe12334 @yuval-hazaz what is the impact of this request?
how important is it for us to execute it soon? what is the risk?

Issue #4943, opened 2023-01-14
Refactor DSG code structure


overbit:
In order to improve DevEx, the DSG library should be refactored so it would not need to be built and push as NPM package.

Potential steps needed:
- [x] Merge DSG library _back_ into DSG runner project and delete `lib/amplication-data-service-generator`
- [x] Delete/obsolete any dsg npm packages created after the release of the new version of DSG runner.

_Good to have_
- [x] _Migrate DSG runner to fully integrate with Nx._

These changes should not affect infrastructure.

### Acceptance Criteria / Testing scenarios

Scenario 1. local development 
* *Given* a contributor running amplication locally
* *When* it runs dsg jobs locally through the local runner
* *Then* the code generation process completes successfully

Scenario 2. amplication cloud environment 
* *Given* a user using the amplication platform
* *When* it clicks the `Rebuild` button
* *Then* the code generation process completes successfully and it is possible to see the Git_Reference as DSG_VERSION in the logs.

Unblock #4895

abrl91:
Passed QA:
In the sandbox env:


![Image](https://user-images.githubusercontent.com/39680385/219015812-9fed5db0-4f2f-4840-a89e-619100c4cc88.png)


Locally:


![Image](https://user-images.githubusercontent.com/39680385/219017764-1d18dbaa-de7c-4aff-a200-27ca4119ce68.png)

PR #4995, opened 2023-01-18
Return the DSG to the tsconfig

tupe12334:
This PR puts the DSG back to the main tsconfig and we open a separate issue to remove it orderly #4994

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -26,6 +26,9 @@
       "@amplication/data-service-generator": [
         "libs/data-service-generator/src/index.ts"
       ],
+      "@amplication/data-service-generator/*": [
+        "libs/data-service-generator/src/*"
+      ],
       "@amplication/design-system": [
         "packages/amplication-design-system/src/index.ts"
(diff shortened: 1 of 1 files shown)


overbit:
@tupe12334 what is the reason to add it back here? What are we trying to achieve?

tupe12334:
We removed this to prevent a bug in runtime and we fixed the bug, but now in compile time we use the files in a local test that crash without this https://github.com/amplication/amplication/blob/master/packages/amplication-data-service-generator-runner/scripts/create-input-json.ts

EugeneTseitlin:
@tupe12334 Nether issue nor the code change is obvious for me. Please, link the corresponding issue to this PR. If doesn't exist yet, then create it first and describe a problem there.

abrl91:
@tupe12334 I don't understand why we need to roll back to this approach, in one of your previous PR you changed it and explained to me why. So, why do we need it like that now?

overbit:
This is now overruled by #4943

PR #5010, opened 2023-01-18
refactor(runner-dsg): remove dynamic installation logic

morhag90:
Related issue:  https://github.com/amplication/amplication-infrastructure-next/issues/28 

## PR Details

Rollback dynamic installation logic. 

## PR Checklist

### DIFF
--- a/packages/amplication-data-service-generator-runner/src/main.ts
+++ b/packages/amplication-data-service-generator-runner/src/main.ts
@@ -35,17 +35,7 @@ export default async function generateCode(
   try {
     const resourceData = await readInputJson(source);
 
-    const modules = await createDataService(
-      resourceData,
-      defaultLogger,
-      join(
-        __dirname,
-        "..",
-        "..",
-        "..",
-        "libs/data-service-generator/node_modules"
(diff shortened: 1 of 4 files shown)


packages/amplication-data-service-generator-runner/src/main.ts
```diff
@@ -35,14 +34,17 @@ export default async function generateCode(
...
       resourceData,
       defaultLogger,
-      join(__dirname, "..", AMPLICATION_MODULES)
+      join(
+        __dirname,
+        "..",
+        "..",
+        "..",
+        "libs/data-service-generator/node_modules"
+      )
```
yuval-hazaz, line 47: according to the code here - https://github.com/amplication/amplication/blob/2b00576375171b590a6f93744bac57e0aa43b57f/libs/data-service-generator/src/register-plugin.ts#L46

I think we can pass null

yuval-hazaz (approved):
please see my comments, I assume your code can work - but please test and consider using null instead of the complex path

overbit (approved):

Issue #5011, opened 2023-01-19
🐛 Bug Report: fix local-dsg-runner cicd flow


overbit:
### What happened?

Changes to `local-dsg-runner` are currently blocking the CI/CD flow in `next` and `master` as shown in the following failing workflow.
https://github.com/amplication/amplication/actions/runs/3956698137/jobs/6776203564

### What you expected to happen

Release workflow should not fail for changes to projects that are docker based

### How to reproduce

Check a release workflow with `local-dsg-runner` 

### Amplication version

next

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #5012, opened 2023-01-19
chore(ci): stop nx to fail release job if a project doesn't have dock…

overbit:
Close: #5011 

## PR Details

This PR prevents **Release** workflow to fail in case of changes to projects that are not configured with `docker:build` target. 
It consists of a quick fix, that will just prevent the docker step to fail, and a stronger fix will come later that will prevent the docker step to run completely for projects that are not deployed as docker containers.

### Detailed explaination

Currently we run the following command that exit with `1` if the nx project is not configured with the `docker:build` target
```sh
npx nx docker:build data-service-generator
```
image

The following command instead will exit with `0` printing a quick message

```sh
npx nx run-many --target=docker:build --projects=data-service-generator
```
image

## PR Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -133,7 +133,7 @@ jobs:
 
       # https://www.npmjs.com/package/@nx-tools/nx-docker
       - name: Dockerize
-        run: npx nx docker:build ${{ matrix.project }}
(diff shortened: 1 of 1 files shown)


levivannoort (approved):
Seems like a great solution to temporary unblock, also the addition to the run-name to add the commit sha to the title is a nice addition.

Issue #4966, opened 2023-01-16
As Dev - I want a new automated deployment process for production 


overbit:
As part of the migration to the new infrastructure, we implemented a new set of workflows as part of #4903.

This issue is about implementing the missing part of automating the deployment to production environment.
The new workflow should follow the strategic (_FUTURE_) design: https://miro.com/app/board/uXjVP7TJ5LI=/

levivannoort:
@overbit created a quick setup for how i think the pipeline should look, could you please give your thoughts before I continue with it.

https://github.com/levivannoort/amplication-fork/commit/07d5216415e04145ed4666b648ce2711f14f935e#diff-84dff8d1094ca39c02ac0e48d951ca22f4da29c76b50ae517f5bd2d50f94c2f6

overbit:
It looks good to me. We will need to add an additional job that will handle the Nx affected logic and pass the projects to the matrix I assume.

levivannoort:
Added a pull request for this issue, it however requires changes on pull request #5021 and it being merged!

levivannoort:
When doing the release this week, we need to check if it fully works.

PR #5021, opened 2023-01-19
refactor(ci): CI/CD workflows to increase visibility

overbit:
Close: #5011

## PR Details

This refactors the workflow setup to be more flexible with Nx projects that don't need a deployment.
Once this is merged we need also to update the documentation.

## Testing 
As testing is only possible by merging in `master`, I've replicated the changes here: https://github.com/overbit/amplication/actions/runs/3960321485

## PR Checklist

- [x] Tests for the changes have been added
- [x] `npm test` doesn't throw any error

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/amplication/amplication/blob/master/CODE_OF_CONDUCT.md) file for detailed contributing guidelines.

PR #5049, opened 2023-01-24
feat: introduce workflow for doing a release to production

levivannoort:
Close: #4966 

## PR Details

**Blocker**: this pull request requires that pull request #5021 is merged first and also includes the name suggested in the comments, namely `nx.template.yml` instead of `nx-affected.template.yml`.

This pull request introduces a new workflow that is meant to bridge the gap between the implementation of argocd image update and the current implementation after migration (i.e. manual changes on the production manifests). It first determines the previous git tag/release based on the one that triggered the workflow, after which it check which apps are affected and deployed in the form of container (i.e. provides output `affected-dockerable`). These affected components are then packaged and pushed to a container registry, and the last jobs takes care of changing the image tag of the affected components and subsequently creates a pr on the manifest repository.

## PR Checklist

### DIFF
--- a/.github/workflows/deployment.production.yml
+++ b/.github/workflows/deployment.production.yml
@@ -18,7 +18,7 @@ jobs:
           fetch-depth: 0
 
       - name: Determine base tag
-        run: echo "tag=$(git tag --sort=-creatordate | grep -A 1 ${{ github.ref_name }} | tail -n 1)" >> $GITHUB_OUTPUT # ${{ github.ref_name }}
+        run: echo "tag=$(git tag --sort=-creatordate | grep -A 1 ${{ github.ref_name }} | tail -n 1)" >> $GITHUB_OUTPUT
         id: base
 
   package:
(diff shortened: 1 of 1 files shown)


levivannoort:
example of workflow output: https://github.com/amplication/amplication-manifests/pull/30

overbit (approved):

.github/workflows/deployment.production.yml
```diff
@@ -0,0 +1,115 @@
+name: Deployment production
+
+on:
+  release:
+    types: [published]
+
```
overbit, line 6: Let's add the NX_CLOUD_ACCESS_TOKEN variable to leverage Nx Cloud distributed cache

overbit (changes requested):

.github/workflows/deployment.production.yml
levivannoort: add the token as an environments variable

overbit (approved):

overbit (approved):
💯  finally automated production releases!

PR #5053, opened 2023-01-24
add version to all BI events

barshimi:
Close: #4916 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/App.tsx
+++ b/packages/amplication-client/src/App.tsx
@@ -35,7 +35,7 @@ export const enhance = track<keyof typeof context>(
   context,
 
   {
-    dispatch: (event: Partial<Event>) => dispatch(event, version),
+    dispatch,
   }
 );
 
--- a/packages/amplication-client/src/App.tsx
+++ b/packages/amplication-client/src/App.tsx
@@ -11,7 +11,6 @@ import useCurrentWorkspace from "./Workspaces/hooks/useCurrentWorkspace";
 import { Loader, PlanUpgradeConfirmation } from "@amplication/design-system";
 import useLocalStorage from "react-use-localstorage";
 import queryString from "query-string";
-import { version } from "./util/version";
 
 declare global {
   interface Window {
--- a/packages/amplication-client/src/App.tsx
+++ b/packages/amplication-client/src/App.tsx
@@ -11,6 +11,7 @@ import useCurrentWorkspace from "./Workspaces/hooks/useCurrentWorkspace";
 import { Loader, PlanUpgradeConfirmation } from "@amplication/design-system";
 import useLocalStorage from "react-use-localstorage";
 import queryString from "query-string";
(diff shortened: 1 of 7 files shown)


abrl91 (approved):

packages/amplication-client/src/Workspaces/WorkspaceLayout.tsx
```diff
@@ -95,6 +101,7 @@ const WorkspaceLayout: React.FC<Props> = ({ innerRoutes, moduleClass }) => {
     workspaceId: currentWorkspace?.id,
     projectId: currentProject?.id,
     resourceId: currentResource?.id,
+    version,
```
yuval-hazaz, line 104: This adds the version to all events under the workspace but it will not add the version to other events. 
Instead, I think we should register the version number in the init function of the analytics service and use it on all track calls

yuval-hazaz (changes requested):

packages/amplication-client/src/Workspaces/WorkspaceLayout.tsx
barshimi: isn't it part of the analytics track event ?

packages/amplication-client/src/App.tsx
```diff
@@ -11,6 +11,7 @@ import useCurrentWorkspace from "./Workspaces/hooks/useCurrentWorkspace";
 import { Loader, PlanUpgradeConfirmation } from "@amplication/design-system";
 import useLocalStorage from "react-use-localstorage";
 import queryString from "query-string";
+import { version } from "./util/version";
```
yuval-hazaz, line 14: remove unused import

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -75,13 +76,6 @@ const WorkspaceHeader: React.FC<{}> = () => {
     currentProjectConfiguration,
   ]);
 
-  const [version, setVersion] = useState("");
```
yuval-hazaz, line 78: useEffect is defined but never used

packages/amplication-client/src/util/analytics.ts
```diff
@@ -31,7 +33,10 @@ export function dispatch(event: Partial<Event>) {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     //@ts-ignore
     const analytics = window.analytics;
-    analytics.track(eventName || MISSING_EVENT_NAME, rest);
+    analytics.track(eventName || MISSING_EVENT_NAME, {
+      ...versionObj,
```
yuval-hazaz, line 37: The version should also be sent in line 28 to _hsq

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #4919, opened 2023-01-11
🐛 Bug Report: "PricingPageCTAClick" event recording the wrong "current plan" param


GreenMachine01:
**Relates to** - https://github.com/amplication/amplication/issues/4744

I am clicking the contact us on the pricing page as a "Free" workspace and in the event the "currentPlan" is shown as Pro

![Image](https://user-images.githubusercontent.com/112330500/211878925-64af0b3d-8a26-494e-9b0c-d4ab1d79ed66.png)

PR #5057, opened 2023-01-24
pricingpagectaclick wrong type sent by the event

barshimi:
Close: #4919 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Purchase/PurchasePage.tsx
+++ b/packages/amplication-client/src/Purchase/PurchasePage.tsx
@@ -1,4 +1,4 @@
-import { Paywall, BillingPeriod } from "@stigg/react-sdk";
+import { Paywall, BillingPeriod, Price } from "@stigg/react-sdk";
 import { useTracking } from "../util/analytics";
 import { AnalyticsEventNames } from "../util/analytics-events.types";
 import { useHistory } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -56,6 +56,21 @@ const selectedPlanAction = {
   },
 };
 
+const getPlanPrice = (
+  selectedBillingPeriod: BillingPeriod,
+  pricePoints: Price[]
+) => {
+  const UNKNOWN = "unknown";
```
abrl91, line 63: WHY IN CAPS?

packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -56,6 +56,21 @@ const selectedPlanAction = {
...
+  if (!pricePoints.length) return UNKNOWN;
+
+  return pricePoints.reduce((price: string, pricePoint: Price) => {
+    if (pricePoint.billingPeriod === selectedBillingPeriod)
+      price = `${pricePoint.amount}${pricePoint.currency}`;
+
+    return price;
+  }, UNKNOWN);
```
yuval-hazaz, line 71: I think this data is not relevant and also not correct. 
It will return the list price of the selected plan and not the real price of the current plan.
@GreenMachine01  I think we can remove it. WDYT?

yuval-hazaz (changes requested):

packages/amplication-client/src/Purchase/PurchasePage.tsx
barshimi: usually static const are in caps

packages/amplication-client/src/Purchase/PurchasePage.tsx
barshimi: this was the feature request

yuval-hazaz (approved):

PR #5070, opened 2023-01-26
Catch error issued by kafka client so build status can be updated correctly

EugeneTseitlin:
converted observable to promise 
added await for promise when emit is called

Close: #4947

## PR Checklist

### DIFF
--- a/packages/amplication-build-manager/src/queue/queue.service.ts
+++ b/packages/amplication-build-manager/src/queue/queue.service.ts
@@ -1,18 +1,14 @@
-import { Inject, Injectable, OnModuleInit } from "@nestjs/common";
+import { Inject, Injectable } from "@nestjs/common";
 import { ClientKafka } from "@nestjs/microservices";
 
 export const KAFKA_CLIENT = "KAFKA_CLIENT";
 
 @Injectable()
-export class QueueService implements OnModuleInit {
+export class QueueService {
   constructor(
     @Inject(KAFKA_CLIENT)
     private readonly kafkaClient: ClientKafka
   ) {}
-
-  async onModuleInit() {
-    await this.kafkaClient.connect();
-  }
 
   async emitMessage(topic: string, message: string): Promise<void> {
     return await new Promise((resolve, reject) => {
--- a/packages/amplication-build-manager/src/queue/queue.service.ts
+++ b/packages/amplication-build-manager/src/queue/queue.service.ts
@@ -1,16 +1,29 @@
-import { Inject, Injectable } from "@nestjs/common";
+import { Inject, Injectable, OnModuleInit } from "@nestjs/common";
 import { ClientKafka } from "@nestjs/microservices";
 
 export const KAFKA_CLIENT = "KAFKA_CLIENT";
 
 @Injectable()
-export class QueueService {
+export class QueueService implements OnModuleInit {
   constructor(
     @Inject(KAFKA_CLIENT)
     private readonly kafkaClient: ClientKafka
   ) {}
 
-  emitMessage(topic: string, message: string): void {
-    this.kafkaClient.emit(topic, message);
+  async onModuleInit() {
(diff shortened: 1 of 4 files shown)


packages/amplication-build-manager/src/queue/queue.service.ts
```diff
@@ -1,16 +1,29 @@
...
   constructor(
     @Inject(KAFKA_CLIENT)
     private readonly kafkaClient: ClientKafka
   ) {}
 
-  emitMessage(topic: string, message: string): void {
-    this.kafkaClient.emit(topic, message);
+  async onModuleInit() {
+    await this.kafkaClient.connect();
+  }
```
overbit, line 15: The problem with this solution is that it would break the current build process that requires to start the nestjs application to generate graphql schemas. The nestjs won't be able to start unless a running kafka cluster is available.

Building process should be independent from other systems like kafka

overbit (changes requested):

packages/amplication-build-manager/src/queue/queue.service.ts
EugeneTseitlin: I removed connect call.

packages/amplication-server/src/core/queue/queue.service.ts
```diff
@@ -5,16 +5,21 @@ export const QUEUE_SERVICE_NAME = "QUEUE_SERVICE";
...
-    this.kafkaClient.emit(topic, message);
+    return await new Promise((resolve, reject) => {
+      this.kafkaClient.emit(topic, message).subscribe({
+        error: (err: any) => {
+          reject(err);
+        },
+        next: () => {
+          resolve();
+        },
+      });
```
yuval-hazaz, line 22: @EugeneTseitlin I saw your comment on the issue (https://github.com/amplication/amplication/issues/4947#issuecomment-1403765358) and I just want to make sure if this PR really fixes the issue or just catch the error and show it on screen

yuval-hazaz (changes requested):

packages/amplication-server/src/core/queue/queue.service.ts
overbit: @EugeneTseitlin can you get more details from the error in order to try  maybe re-emitting the event or checking the readiness of the kafkaClient before the emission.

overbit (changes requested):

packages/amplication-server/src/core/queue/queue.service.ts
EugeneTseitlin: @yuval-hazaz the change solves issue as it was declared "Code generation never ends". So, code generation won't stuck forever in case that a message can't be sent to kafka, it will be marked as failed then.

EugeneTseitlin: @overbit Promise already catches the whole error object, there are no other details. Retry is an option, but for now I've decided not to implement it in order to see how often this issue occurs. Now having logs it can be tracked.

yuval-hazaz: @EugeneTseitlin the issue describes this "Looking on the logs, we can see that the build message is sent to the Kafka queue, but by looking on kafka UI we never see the message" 

just to make sure - your fix does not resolve the root issue, it only resolves the issue that the UI is not updated.

What about the actual root cause? are we still looking at it?

yuval-hazaz:
@EugeneTseitlin, can you please update the PR title and explain the actual value/purpose and not the technical details of the implementation?

yuval-hazaz (changes requested):
before approving this PR, please link to a new issue that talks about resolving the root cause

yuval-hazaz (approved):

overbit (approved):

overbit:
The actual issue would be fixed as part of https://github.com/amplication/amplication/issues/5083

Issue #5050, opened 2023-01-24
As Dev - I want to refactor git sync and strategy to support multiple Git providers


abrl91:
Refactor git sync and strategy to support multiple providers, and not only GitHub

overbit:
# Test results:

Passed ✅  

Tests scenarios:
- ✅  Connect with GitHub and create a new repository
  ![Image](https://user-images.githubusercontent.com/2861984/216390385-295da27c-7922-49fb-a2fb-d95960ca0d4e.png)

- ✅   Build a new service and push to GitHub  
![Image](https://user-images.githubusercontent.com/2861984/216390777-f8319878-9d93-400b-afe5-04f5c0f61b02.png)

PR #5074, opened 2023-01-26
Refactor(git-utils): support multiple git providers

abrl91:
part of: #5050 

## PR Details

- refactor git factory in git utils
- refactor GitHub service in git utils
- refactor the `git.service.ts` (and rename it to git-client service)
- refactor git provider service in the amplication server - now it uses the factory and calls it with the provider from the args
- refactor pull request service with the changes related to git utils

I refactored only the functions that we already have in the git utils lib. 

### Next step:
- more functionality from the git provider service in the server to the git utils
- refactoring of the git strategy functions (for now, it only knows to work with GitHub)
- add test to git utils for octokit
- add a logger

## Test cases to check:
- get repositories, 
- create repository
- connect to an existing repo
- sync (connect Github organization/user to amplication)
- create a pull request
- pull request with pricing
- .amplicaton ignore

Everything work as before.

## PR Checklist


packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -3,31 +3,38 @@ import { ConfigService } from "@nestjs/config";
...
 
 const GITHUB_FILE_TYPE = "file";
 export const GITHUB_CLIENT_SECRET_VAR = "GITHUB_CLIENT_SECRET";
 export const GITHUB_APP_APP_ID_VAR = "GITHUB_APP_APP_ID";
 export const GITHUB_APP_PRIVATE_KEY_VAR = "GITHUB_APP_PRIVATE_KEY";
 export const GITHUB_APP_INSTALLATION_URL_VAR = "GITHUB_APP_INSTALLATION_URL";
 export const UNEXPECTED_FILE_TYPE_OR_ENCODING = `Unexpected file type or encoding received`;
-
 type DirectoryItem = components["schemas"]["content-directory"][number];
 @Injectable()
```
overbit, line 36: Since we use a Class Factory now, we don't need to inject this class through DI

packages/amplication-git-utils/src/git/git.module.ts
```diff
@@ -13,7 +12,7 @@ import { SERVICE_NAME } from "./git.constants";
       metadata: { service: SERVICE_NAME },
     }),
   ],
-  providers: [GitService, GitServiceFactory, GithubService],
-  exports: [GitService, GitServiceFactory, GithubService],
+  providers: [GitFactory, GithubService],
+  exports: [GitFactory, GithubService],
```
overbit, line 16: Since we use a Class Factory now, we don't need to inject this class through DI.

```suggestion
  providers: [GitFactory],
  exports: [GitFactory],
```

packages/amplication-git-utils/src/types.ts
```diff
@@ -19,6 +19,31 @@ export enum EnumGitProvider {
...
+  baseBranch?: string;
+  author?: string;
+  filePath?: string;
+  gitResourceMeta?: GitResourceMeta;
+  pullRequestModule?: PullRequestModule[];
+  amplicationWorkspaceId?: string;
+  limit?: number;
+  page?: number;
+  head?: string;
+}
```
overbit, line 45: How about keeping options for the most generic parameters?
I mean in methods like CreatePullRequest or Commit, `files` might make sense, but not in other like createRepo.
It would help developer to know the mandatory params for functions like the above.

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -3,31 +3,38 @@ import { ConfigService } from "@nestjs/config";
 import { createAppAuth } from "@octokit/auth-app";
```
overbit, line 3: `GithubService` needs to have its own tests to cover business logic declared here.

packages/amplication-git-utils/tests/git/git-service-factory.spec.ts
```diff
@@ -1,14 +1,18 @@
 import { EnumGitProvider } from "../../src/types";
```
overbit, line 1: Rename `git-service-factory` to `git-factory` for consistency

packages/amplication-server/src/core/git/utils/GitServiceFactory/GitServiceFactory.spec.ts
```diff
@@ -1,15 +1,18 @@
...
+  const gitServiceFactory = new GitFactory(configService);
+  describe("GitFactory.getProvider()", () => {
     it("should return an github service", () => {
-      expect(gitServiceFactory.getService(EnumGitProvider.Github)).toBe(github);
+      expect(
+        gitServiceFactory.getProvider({ provider: EnumGitProvider.Github })
+      ).toBe(github);
     });
     it("should throw error if source control dont exist", () => {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
overbit, line 18: What is the benefit of these tests? Let's remove these files if not needed.

packages/amplication-server/src/core/resource/resource.service.spec.ts
```diff
@@ -366,7 +364,7 @@ jest.mock("cuid");
 // @ts-ignore
 cuid.mockImplementation(() => EXAMPLE_CUID);
 
-describe("ResourceService", () => {
+describe.skip("ResourceService", () => {
```
overbit, line 367: No skip are allowed 👮

packages/amplication-server/src/core/git/__mocks__/Github.ts
```diff
@@ -4,9 +4,9 @@ import { TEST_GIT_REPO } from "./GitRepo";
...
 
 export const MOCK_GITHUB_SERVICE = mock<GithubService>({});
-MOCK_GITHUB_SERVICE.getOrganizationRepos.mockReturnValue(
+MOCK_GITHUB_SERVICE.getRepositories.mockReturnValue(
   Promise.resolve(TEST_GIT_REPOS)
 );
-MOCK_GITHUB_SERVICE.createOrganizationRepository.mockReturnValue(
+MOCK_GITHUB_SERVICE.createRepository.mockReturnValue(
   Promise.resolve(TEST_GIT_REPO)
 );
```
overbit, line 12: No GitHub should be needed anymore since now you can just mock the IGitClient

overbit (changes requested):

packages/amplication-git-utils/src/types.ts
abrl91: But then, when I call the provider factory, I would have to pass not only the provider but also the other required params

packages/amplication-server/src/core/git/utils/GitServiceFactory/GitServiceFactory.spec.ts
abrl91: I just tried to fix the existing test. I am not sure what is the benefit of this test in general :)

packages/amplication-server/src/core/git/__mocks__/Github.ts
abrl91: yes...this is just a refactoring of what we have now. We will need to update and rewrite all our tests

packages/amplication-git-utils/src/git/github.service.ts
abrl91: Yes. We should add this test, but I think we can do it after the last step of the refactoring

overbit: Sure, let's create an issue or a task on the main issue to avoid forgetting this please.

packages/amplication-git-utils/src/types.ts
overbit: Sorry I was probably not 100% clear in my statement. 
What I meant is to have something that is more clear as methods params like `commits(options, { files, message, author, repo})` but it's ok to start with this.

overbit (approved):

packages/amplication-git-utils/src/types.ts
```diff
@@ -64,65 +96,13 @@ export interface GitResourceMeta {
...
-    repo: string
-  ): Promise<RemoteGitRepository>;
+  getRepository(options: Options): Promise<RemoteGitRepository>;
+  getRepositories(options: Options): Promise<RemoteGitRepos>;
+  createRepository(options: Options): Promise<RemoteGitRepository>;
+  deleteGitOrganization(options: Options): Promise<boolean>;
+  getGitRemoteOrganization(options: Options): Promise<RemoteGitOrganization>;
+  getFile(options: Options): Promise<GithubFile>;
+  createPullRequest(options: Options): Promise<string>;
+  getGitInstallationUrl(options: Options): Promise<string>;
```
yuval-hazaz, line 107: Each function should use its own parameters, it will be very hard to use the lib if I need to provide all parameters for every type of operation, without knowing which parameter is needed to which operation

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -3,31 +3,38 @@ import { ConfigService } from "@nestjs/config";
...
 export const GITHUB_APP_APP_ID_VAR = "GITHUB_APP_APP_ID";
 export const GITHUB_APP_PRIVATE_KEY_VAR = "GITHUB_APP_PRIVATE_KEY";
 export const GITHUB_APP_INSTALLATION_URL_VAR = "GITHUB_APP_INSTALLATION_URL";
 export const UNEXPECTED_FILE_TYPE_OR_ENCODING = `Unexpected file type or encoding received`;
-
 type DirectoryItem = components["schemas"]["content-directory"][number];
-@Injectable()
-export class GithubService {
+
+export class GithubService implements GitClient {
```
yuval-hazaz, line 37: this service should only be dependent on the Git provider API and should not be dependent on nest.
All the configuration should be provided as parameters to the functions instead of using the config service

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -1,207 +0,0 @@
-import { Injectable } from "@nestjs/common";
```
yuval-hazaz, line 1: This file should still be used and it should be the entry point for the lib. The clients who use the lib should not be aware of the internal provider/factory implementation.

yuval-hazaz (changes requested):

packages/amplication-git-utils/src/types.ts
abrl91: I agree, but there are differences between the providers and if one parameter is mandatory in provider X and not in providers Y and Z, we will still have to set some parameters as optional although it is mandatory in one provider, and this is also confusing.

packages/amplication-git-utils/src/types.ts
abrl91: fixed

overbit (approved):

packages/amplication-git-utils/src/git/git-client.service.ts
```diff
@@ -0,0 +1,91 @@
...
+  File,
+} from "../types";
+import { prepareFilesForPullRequest } from "../utils/prepare-files-for-pull-request";
+import { GitFactory } from "./git-factory";
+
+@Injectable()
+export class GitClientService implements GitClient {
+  constructor(private readonly gitFactory: GitFactory) {}
+
+  async getGitInstallationUrl(
```
barshimi, line 22: why do you use async if you don't use await

packages/amplication-git-utils/src/git/git-client.service.ts
```diff
@@ -0,0 +1,91 @@
...
+    amplicationWorkspaceId: string,
+    gitProviderArgs: GitProviderArgs
+  ): Promise<string> {
+    const gitProvider = this.gitFactory.getProvider(gitProviderArgs);
+    return gitProvider.getGitInstallationUrl(amplicationWorkspaceId);
+  }
+
+  async getRepository(
```
barshimi, line 30: all functions are not using await

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -0,0 +1,82 @@
...
+  await amplicationIgnoreManger.init(async (fileName) => {
+    try {
+      const file = await this.getFile({
+        owner,
+        repositoryUrl: repositoryName,
+        path: fileName,
+        baseBranch: undefined, // take the default branch
+      });
+      const { content, htmlUrl, name } = file;
+      console.log(`Got ${name} file ${htmlUrl}`);
```
barshimi, line 22: would it be better to write to log instead of console

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -0,0 +1,82 @@
...
+      return content;
+    } catch (error) {
+      console.log("Repository does not have a .amplicationignore file");
```
barshimi, line 25: also here

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -0,0 +1,82 @@
...
+      return "";
+    }
+  });
+
+  //do not override files in 'server/src/[entity]/[entity].[controller/resolver/service/module].ts'
+  //do not override server/scripts/customSeed.ts
+  const doNotOverride = [
```
barshimi, line 32: I'm not sure you need `gitResourceMeta.serverPath` for those regex and if you don't, it's better to set the regex outside of the function so it wont set it every time

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -0,0 +1,82 @@
...
+      `^${gitResourceMeta.serverPath || "server"}/src/[^/]+/.+.module.ts$`
+    ),
+    new RegExp(
+      `^${gitResourceMeta.serverPath || "server"}/scripts/customSeed.ts$`
+    ),
+  ];
+
+  const authFolder = "server/src/auth/";
+
+  const files: Required<Changes["files"]> = Object.fromEntries(
```
barshimi, line 52: I'm not sure I understand the use of map inside fromEntries also I would break it into composition instead of if's. as I see it you change once the module.path and once the module code, so I would determine the path file and then react inside functions and not if

abrl91: this is an existing logic that I extracted to util function so it's not really a part of this PR

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
abrl91: this is an existing logic that I moved/refactored

abrl91:
@barshimi, you reviewed the internal implementation of the function, I only refactored the git utils services in order to implement in the future more providers

packages/amplication-git-utils/src/git/git-client.service.ts
abrl91: Because I return imminently

packages/amplication-git-utils/src/git/git-client.service.ts
abrl91: Because I return imminently so I don't need the await, but it returns a promise

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
abrl91: again, this is a logic that we already have

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
abrl91: again, this is not a new code

overbit:
There was a big refactoring after the approval so I removed my approval.

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -1,207 +1,73 @@
...
-    limit: number,
-    page: number
-  ): Promise<RemoteGitRepos> {
-    const gitService = this.gitServiceFactory.getService(gitProvider);
-    return await gitService.getOrganizationRepos(installationId, limit, page);
+  static createProvider(gitProviderArgs: GitProviderArgs): GitClientService {
+    const serviceInstance = new this();
+    serviceInstance.provider = GitFactory.getProvider(gitProviderArgs);
+    return serviceInstance;
   }
```
overbit, line 24: We are using a class here, but we created a static function with the responsibility to create the class instance.
This is actually a `constructor` https://www.typescriptlang.org/docs/handbook/2/classes.html#constructors

overbit (changes requested):

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -1,82 +1,209 @@
...
-    const appId = this.configService.get(GITHUB_APP_APP_ID_VAR);
-    const privateKey = this.configService
-      .get(GITHUB_APP_PRIVATE_KEY_VAR)
-      .replace(/\\n/g, "\n");
+  init(): void {
+    const {
+      GITHUB_APP_INSTALLATION_URL,
+      GITHUB_APP_APP_ID,
+      GITHUB_APP_PRIVATE_KEY,
+    } = process.env;
```
overbit, line 44: Libraries must not use environment variables unless clearly documented in the library documentation.

overbit (changes requested):

yuval-hazaz (approved):

Issue #4967, opened 2023-01-16
As User - I want to feel the purchase flow is starting "immediately" after I click the upgrade CTA in the pricing page


GreenMachine01:
**Overview**:
Currently when the user clicks to upgrade the plan on the pricing page there is a small delay until the Stripe checkout opens up.
The user might feel his click didn't work and get frustrated and potentially lead to rage clicks on the CTA.

We want to improve the experience for the user + prevent additional clicks on the CTA while the checkout process is initiating via Stigg.

**Requirements:**
1. When the user clicks the CTA we immediately darken the screen and show a loader with a message until the checkout process starts.
2. We need to make sure the user is not stuck in this state so lets make sure the notification is closed after [X time] and record this issue in our logs/events so we can track and understand what happened.

![Image](https://user-images.githubusercontent.com/112330500/212727750-7fd88940-489b-4b38-9867-23b71ed1c60c.png)

**Plan:**
- [ ] Create modal component with loader. Base Modal component exists in design system, Purchase Page can taken as an example.
- [ ] Trigger modal on `purchaseSubscription` request. 

Plan is written by @EugeneTseitlin

EugeneTseitlin:
Here is how implementation looks like:
https://user-images.githubusercontent.com/6080188/215321097-5994060e-80d1-4937-9efb-775d53fefa2a.mp4

abrl91:
@EugeneTseitlin  I tested it and didn't see the popup

abrl91:
There was a problem with the deployment. The issue was resolved

PR #5078, opened 2023-01-29
Feat/4967 as user i want to feel the purchase flow is starting immediately after i click the upgrade cta in the pricing page

EugeneTseitlin:
Closes: #4967

## PR Checklist


packages/amplication-design-system/src/components/Loader/LoaderTiny.tsx
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import Lottie, { LottieComponentProps } from "lottie-react";
+
+import animationData from "../../assets/amplication-loader-tiny.json";
+
+export type Props = Omit<LottieComponentProps, "animationData">;
+
+export const LoaderTiny = (props: Props) => {
```
yuval-hazaz, line 8: I think it makes more sense to use the current loader component and add a new type enum property instead of using a separate component

yuval-hazaz (changes requested):

yuval-hazaz:
@EugeneTseitlin please resolve the conflict

packages/amplication-design-system/src/components/Loader/LoaderTiny.tsx
EugeneTseitlin: I thought about that, but existing Loader component has some wrapper for full screen mode, I've decided not to do refactoring of this component and it's usage. More than that, not all of our loaders are going to be Lottie based. But if anyway we want to make a single Loader component that would accept animation type as parameter, let's do it separately from this issue. Should I open a new issue?

yuval-hazaz (approved):

PR #5103, opened 2023-01-31
Refactor(git-utils): git providers pull request strategies

abrl91:
part of: #5050 

## PR Details

Refactor git strategy: making it more generic for more providers

## PR Checklist


packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -66,7 +76,51 @@ export class GitClientService {
...
+      });
+      return this.provider.createPullRequestForBranch({
+        owner,
+        repositoryName,
+        pullRequestTitle,
+        pullRequestBody,
+        branchName,
+        defaultBranchName: defaultBranch,
+        pullRequestUrl: url,
+      });
```
tupe12334, line 122: @yuval-hazaz wanted me to do that in a separated file

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -306,31 +322,471 @@ export class GithubService implements GitProvider {
...
+  }
+
+  async createPullRequestForBranch({
+    owner,
+    repositoryName,
+    branchName,
+    pullRequestTitle,
+    pullRequestBody,
+    defaultBranchName,
+    pullRequestUrl,
```
tupe12334, line 469: We dont need this param, if there is a pull request already we don't need to get here any way

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -306,31 +322,471 @@ export class GithubService implements GitProvider {
...
+  async createBranchIfNotExists({
+    owner,
+    repositoryName,
+    branchName,
+  }: CreateBranchIfNotExistsArgs): Promise<Branch> {
+    const { sha } = await this.getFirstDefaultBranchCommit(
+      owner,
+      repositoryName
+    );
+    const refs = await this.getBranch(owner, repositoryName, branchName);
```
tupe12334, line 352: If the branch is not existing the function will throw an error

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -6,7 +6,7 @@ export async function prepareFilesForPullRequest(
   gitResourceMeta,
   pullRequestModule,
   amplicationIgnoreManger
-): Promise<Required<Changes["files"]>> {
+): Promise<any | any[]> {
```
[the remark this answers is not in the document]
abrl91: I still need to figure out how to use it in a way that, in our types file, we will be able to use unknown and here the right type

packages/amplication-git-utils/src/git/github.service.ts
abrl91: l don't understand so lets talk about it

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -53,6 +61,18 @@ export class GithubService implements GitProvider {
       appId: this.appId,
       privateKey,
     });
+
+    this.createOctokitToken();
+  }
+
+  private async createOctokitToken() {
+    const myOctokit = Octokit.plugin(createPullRequest);
```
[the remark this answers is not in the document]
yuval-hazaz: instead of using this function, use a getter for the property and only create the octokit plugin once

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -284,20 +302,18 @@ export class GithubService implements GitProvider {
...
       pullRequestTitle,
       pullRequestBody,
-      head,
-    } = createPullRequestArgs;
-
+      branchName,
+      files,
+      commitMessage,
+    } = createPullRequestFromFiles;
     const myOctokit = Octokit.plugin(createPullRequest);
```
yuval-hazaz, line 317: add comment "We are not using this.octokit, instead we are using a local octokit client because we need the plugin"

yuval-hazaz (changes requested):

packages/amplication-git-utils/src/git/github.service.ts
abrl91: I added the comment

packages/amplication-git-utils/src/git/github.service.ts
abrl91: I understand, but in this function which I deleted and handled differently, we are just swallowed the error and wrapped it in a boolean 
```
private async isBranchExist(branch: string): Promise<boolean> {
    try {
      const refs = await this.getBranch(branch);
      return Boolean(refs);
    } catch (error) {
      return false;
    }
  }
```

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -29,34 +29,36 @@ export async function prepareFilesForPullRequest(
...
-            if (exists) return null;
+  const files: File[] = pullRequestModule.map((module) => {
+    // ignored file
+    if (amplicationIgnoreManger.isIgnored(module.path)) {
+      return [
+        {
+          file: join(AMPLICATION_IGNORED_FOLDER, module.path),
+          content: module.code,
+        },
+      ];
```
yuval-hazaz, line 40: why are we returning an array? I would expect the following:
```suggestion
      return {
          file: join(AMPLICATION_IGNORED_FOLDER, module.path),
          content: module.code,
        };
```

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -29,34 +29,36 @@ export async function prepareFilesForPullRequest(
...
+    }
+    // Deleted file
+    if (module.code === null) {
+      return [{ path: module.path, code: module.code }];
```
yuval-hazaz, line 44: same

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -29,34 +29,36 @@ export async function prepareFilesForPullRequest(
...
-            return module.code;
-          },
-        ];
-      }
-      // Regular file
-      return [module.path, module.code];
-    })
-  );
+        return module.code;
+      };
```
yuval-hazaz, line 57: Let's talk about it... I am not sure I get it

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -29,34 +29,36 @@ export async function prepareFilesForPullRequest(
...
+      return [{ path, content }];
+    }
+    // Regular file
+    return [{ path: module.path, code: module.code }];
```
yuval-hazaz, line 61: same - no array

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -29,34 +29,36 @@ export async function prepareFilesForPullRequest(
...
-          },
-        ];
-      }
-      // Regular file
-      return [module.path, module.code];
-    })
-  );
+        return module.code;
+      };
+      return [{ path, content }];
```
yuval-hazaz, line 58: same - no array

yuval-hazaz (changes requested):

packages/amplication-git-utils/src/git/git-factory.ts
```diff
@@ -3,10 +3,16 @@ import { EnumGitProvider, GitProvider, GitProviderArgs } from "../types";
...
+  public static async getProvider(
+    gitProviderArgs: GitProviderArgs
+  ): Promise<GitProvider> {
+    let gitProvider: GitProvider;
+
     switch (gitProviderArgs.provider) {
       case EnumGitProvider.Github:
-        return new GithubService(gitProviderArgs);
+        gitProvider = new GithubService(gitProviderArgs);
+        await gitProvider.init();
```
tupe12334, line 14: Why do we need the init function, its a class, there is a contractor

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -55,21 +57,78 @@ export class GitClientService {
...
+        });
+      }
+      return url;
+    }
   }
 
-  async getFile(file: File): Promise<GitFile> {
+  async getFile(file: GetFileArgs): Promise<GitFile> {
     return this.provider.getFile(file);
   }
```
tupe12334, line 133: Why do we need a separated function for wrapping the provider function

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -55,21 +57,78 @@ export class GitClientService {
...
+    const preparedFiles = await prepareFilesForPullRequest(
       gitResourceMeta,
-      pullRequestModule,
+      files,
       amplicationIgnoreManger
     );
-    return this.provider.createPullRequest(createPullRequestArgs, files);
+
+    if (pullRequestMode === EnumPullRequestMode.Basic) {
+      console.log(pullRequestMode, "pullRequestMode");
```
tupe12334, line 83:
```suggestion
```

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
```diff
@@ -1,12 +1,12 @@
 import { join } from "path";
-import { Changes } from "octokit-plugin-create-pull-request/dist-types/types";
 import { AMPLICATION_IGNORED_FOLDER } from "../git/git.constants";
+import { File } from "../types";
 
 export async function prepareFilesForPullRequest(
   gitResourceMeta,
   pullRequestModule,
   amplicationIgnoreManger
```
tupe12334, line 8: What are the types?

tupe12334:
Don't we think it's better to separate the DTOs to separated files?

tupe12334:
I really love the work that you have been done here @abrl91

packages/amplication-git-utils/src/utils/prepare-files-for-pull-request.ts
abrl91: fixed

packages/amplication-git-utils/src/git/git-factory.ts
abrl91: Because we need to perform async operation in the constructor and we can't, so this is the approach we took

packages/amplication-git-utils/src/git/git.service.ts
abrl91: because we may need this function for other providers and the git service is the service that wraps the providers

abrl91:
> Don't we think it's better to separate the DTOs to separated files?

These are interfaces of the methods' parameters. There aren't classes, and we don't need DTOs with class validators or so...
@yuval-hazaz BTW, that's the reason I didn't like the 'Args' suffix

packages/amplication-git-utils/src/git/git-factory.ts
tupe12334: Got that, thanks, by the way, as a lib writer I suggest adding check to every function that the init has been run before the function run

tupe12334 (approved):

PR #5128, opened 2023-02-06
refactor(git-util):  extract GitProvider interface in its own file

tupe12334:
Partily: #5113

### DIFF
--- a/packages/amplication-git-utils/src/IGitProvider.ts
+++ b/packages/amplication-git-utils/src/IGitProvider.ts
@@ -1,0 +1,46 @@
+import {
+  Branch,
+  CreateBranchIfNotExistsArgs,
+  CreateCommitArgs,
+  CreatePullRequestForBranchArgs,
+  CreatePullRequestFromFilesArgs,
+  CreateRepositoryArgs,
+  GetFileArgs,
+  GetPullRequestForBranchArgs,
+  GetRepositoriesArgs,
+  GetRepositoryArgs,
+  GitFile,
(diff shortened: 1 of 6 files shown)


packages/amplication-git-utils/src/IGitProvider.ts
```diff
@@ -0,0 +1,46 @@
...
+  GetPullRequestForBranchArgs,
+  GetRepositoriesArgs,
+  GetRepositoryArgs,
+  GitFile,
+  RemoteGitOrganization,
+  RemoteGitRepos,
+  RemoteGitRepository,
+} from "./types";
+
+export interface IGitProvider {
```
overbit, line 18: There is no convention to prefix interfaces in Ts with `I` and we are not currently using this convention elsewhere. Please rename it to `GitProvider`

overbit (changes requested):

abrl91:
@overbit @tupe12334 
This change doesn't conform to our conventions
[File & Declaration Naming](https://www.notion.so/amplication/File-Declaration-Naming-666afe407ff14ed6aa54bcd3dead45a8)

overbit:
Good point @abrl91 . Please @tupe12334 rename this

abrl91:
@overbit 
It's not only the renaming. All shared types/enums/interfaces should be in the same file. This interface is shared between all providers

tupe12334:
I think we need to refactor the notion file because we don't work like that anywhere not even in our generated code

Issue #5129, opened 2023-02-06
Extract the amplication branch creation logic from the provider to the service


tupe12334:

abrl91:
The PR was merged into an umbrella branch and no to the next branch, so this issue cannot be tested

overbit:
@tupe12334 is there any feasible way to test this issue?

EugeneTseitlin:
@tupe12334 How should it be tested? Is it refactoring? Does it mean that everything just should work as before?

tupe12334:
There isnt a way to test it, just when the umbrula branch merges, making sure that the functionality works properly.

PR #5133, opened 2023-02-06
Refactor/extract branch creation logic

tupe12334:
Close: #5129 

## PR Details

This pull request extract the branch creation logic from the provider to the service
## PR Checklist


abrl91 (changes requested):
These changes are related to another PR on which I commented that the change doesn't conform to our conventions.
https://github.com/amplication/amplication/pull/5128
Please fix it first

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -124,6 +125,19 @@ export class GitClientService {
     }
   }
 
+  private async createBranchIfNotExists(args: OneBranchArgs) {
```
abrl91, line 128: Why did you add this method to the git provider?
@yuval-hazaz when we talked about it while I was working on the refactoring we decided that those functions will be in the provider and not in the git service. What changed

abrl91: https://github.com/amplication/amplication/pull/5128

packages/amplication-git-utils/src/types.ts
```diff
@@ -141,3 +135,19 @@ export interface CreateCommitArgs {
...
 }
+
+export interface OneBranchArgs extends FindOneRepositoryArgs {
+  branchName: string;
+}
+export interface FindOneRepositoryArgs {
+  owner: string;
+  repositoryName: string;
+}
+
```
yuval-hazaz, line 146: is this being used?

packages/amplication-git-utils/src/types.ts
```diff
@@ -141,3 +135,19 @@ export interface CreateCommitArgs {
   branchName: string;
   files: UpdateFile[];
 }
+
+export interface OneBranchArgs extends FindOneRepositoryArgs {
```
yuval-hazaz, line 139: Why do we need to extend? What is the relation between OneBranchArgs and FindOneRepositoryArgs? I don't think there is any relation. Args of a function may change in one function without any change in the other, so this relation is not helping and just make things harder

yuval-hazaz, line 139: please rename OneBranchArgs to getBranchArgs

packages/amplication-git-utils/src/types.ts
```diff
@@ -141,3 +135,19 @@ export interface CreateCommitArgs {
...
+  branchName: string;
+}
+export interface FindOneRepositoryArgs {
+  owner: string;
+  repositoryName: string;
+}
+
+export interface CreateBranchArgs extends OneBranchArgs {
+  pointingSha: string;
+}
```
yuval-hazaz, line 149: please remove the extends and use all immediate params

packages/amplication-git-utils/src/git/git.service.ts
yuval-hazaz: I think this way is ok

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -124,6 +125,19 @@ export class GitClientService {
...
+    const isBranchExist = await this.provider.isBranchExists(args);
```
yuval-hazaz, line 129: use this.provider.getBranch(args) instead of isBranchExists

packages/amplication-git-utils/src/GitProvider.ts
```diff
@@ -33,14 +35,15 @@ export interface GitProvider {
...
-    createBranchIfNotExistsArgs: CreateBranchIfNotExistsArgs
-  ) => Promise<Branch>;
   createCommit: (createCommitArgs: CreateCommitArgs) => Promise<void>;
   getPullRequestForBranch: (
     getPullRequestForBranchArgs: GetPullRequestForBranchArgs
   ) => Promise<{ url: string; number: number } | undefined>;
   createPullRequestForBranch: (
     createPullRequestForBranchArgs: CreatePullRequestForBranchArgs
   ) => Promise<string>;
+  isBranchExists: (args: OneBranchArgs) => Promise<boolean>;
```
yuval-hazaz, line 45: we dont need isBranchExists - just use "getBranch"

yuval-hazaz (changes requested):

packages/amplication-git-utils/src/GitProvider.ts
```diff
@@ -1,6 +1,7 @@
 import {
   Branch,
-  CreateBranchIfNotExistsArgs,
+  Commit,
```
yuval-hazaz, line 3: please rename the file to `git-provider.interface.ts`

packages/amplication-git-utils/src/types.ts
tupe12334: Yes

packages/amplication-git-utils/src/git/git.service.ts
tupe12334: No because it will throw an error if the branch don't exist

packages/amplication-git-utils/src/types.ts
tupe12334: You cant find branch without the repository, so instead of rewrite the owner and the repo we just extends

yuval-hazaz (approved):

Issue #4560, opened 2022-12-06
As User - I want a better experience when getting an error for project name


GreenMachine01:
**As User - I want a better experience when getting an error for the project name so that I clearly understand what happened and am able to fix it**

![Image](https://user-images.githubusercontent.com/112330500/204148892-0addabbb-dcb9-448d-a782-a432b2ed2ed3.png)

**Design:**
https://www.figma.com/file/0C6Up7TUJ6hqdVfoAWjf5K/%233956---Input-Field---Error-message?node-id=1136%3A6535&t=Ekl6EqwmqOR1k9Af-0

GreenMachine01:
design is not part of the story

morhag90:
PASSED QA:
<img width="546" alt="image" src="https://user-images.githubusercontent.com/97830649/221808639-8362f249-2c5a-4750-9cac-90566945114d.png">

PR #5138, opened 2023-02-07
 🐛 Bug Report: Wrong design for error with project name #4560

gitstart:
### DEMO VIDEO
https://www.loom.com/share/57be56fd96c54edd87e0330ac1fe91af.

### DIFF
--- a/packages/amplication-client/src/Project/NewProject.tsx
+++ b/packages/amplication-client/src/Project/NewProject.tsx
@@ -1,7 +1,8 @@
-import { TextField, Snackbar } from "@amplication/design-system";
+import { TextField } from "@amplication/design-system";
 import { gql, useMutation } from "@apollo/client";
 import { Form, Formik } from "formik";
-import React, { useCallback, useContext, useState } from "react";
+import { formatError } from "../util/error";
+import React, { useCallback, useContext } from "react";
 import { GlobalHotKeys } from "react-hotkeys";
 import { Button, EnumButtonStyle } from "../Components/Button";
 import { EnumImages, SvgThemeImage } from "../Components/SvgThemeImage";
@@ -9,7 +10,6 @@ import { AppContext } from "../context/appContext";
 import * as models from "../models";
 import { useTracking } from "../util/analytics";
 import { AnalyticsEventNames } from "../util/analytics-events.types";
-import { formatError } from "../util/error";
 import {
   validate,
   validationErrorMessages,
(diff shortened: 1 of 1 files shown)


gitstart:
@GreenMachine01 This PR is ready for review. Thanks. cc @EugeneTseitlin

packages/amplication-client/src/Project/NewProject.tsx
```diff
@@ -56,31 +58,30 @@ type Props = {
...
   const handleSubmit = useCallback(
-    (data: CreateProjectType) => {
+    (data: CreateProjectType, { setErrors }) => {
       createProject({
         variables: {
           data,
         },
-      }).catch(console.error);
+      }).catch((err) => {
+        setErrors({ name: ERROR });
```
yuval-hazaz, line 79: Please keep the original error text. The current implementation will display the same error message for all types of errors that can be returned from the server which is not correct 
```suggestion
        setErrors({ name: formatError(err); });
```

packages/amplication-client/src/Project/NewProject.tsx
```diff
@@ -56,31 +58,30 @@ type Props = {
...
+        console.error(err);
```
yuval-hazaz, line 80: we can remove this now

yuval-hazaz (dismissed):
Thanks for the fix! I like the way you push the error to the form. We should use this as a common practice (@overbit @abrl91 please have a look) 


Please see one comment to be fixed

gitstart:
@yuval-hazaz  resolved the comments.

EugeneTseitlin (changes requested):
Please, make sure that changes are applied only to New Project component. Merge recent changes from our repo and resolve a conflict.
The change in New Project component is fine, I've checked it and it works as expected.

gitstart:
@EugeneTseitlin  resolved conflicts.

EugeneTseitlin:
> @EugeneTseitlin conflicts resolved.

There should not be any change in  ISSUE_TEMPLATE, workflows folders.
Changes should be only in NewProject.tsx file.

gitstart:
> > @EugeneTseitlin conflicts resolved.
> 
> There should not be any change in ISSUE_TEMPLATE, workflows folders. Changes should be only in NewProject.tsx file.

Hey @EugeneTseitlin reverted changes in workflow file. I think it was added somehow when we try to resolve changes between master and next branch

EugeneTseitlin (approved):

PR #5144, opened 2023-02-07
feat(git-utils): getting all amplication commits before creating the branch

tupe12334:
Close: #5143 

## PR Details

This pull request collect all the amplication's commits from the default branch
## PR Checklist

### DIFF
--- a/packages/amplication-git-utils/src/git-provider.interface.ts.ts
+++ b/packages/amplication-git-utils/src/git-provider.interface.ts.ts
@@ -45,5 +45,5 @@ export interface GitProvider {
   getBranch: (args: GetBranchArgs) => Promise<Branch>;
   createBranch: (args: CreateBranchArgs) => Promise<Branch>;
   getFirstCommitOnBranch: (args: GetBranchArgs) => Promise<Commit>;
-  getMyCommitsList: (args: GetBranchArgs) => Promise<Commit[]>;
+  getCurrentUserCommitList: (args: GetBranchArgs) => Promise<Commit[]>;
 }
--- a/packages/amplication-git-utils/src/git/github.service.ts
+++ b/packages/amplication-git-utils/src/git/github.service.ts
@@ -10,8 +10,10 @@ import {
 } from "octokit-plugin-create-pull-request/dist-types/types";
 import { GitProvider } from "../GitProvider";
 import {
+  AmplicationBotData,
   Branch,
-  OneBranchArgs,
+  Commit,
+  CreateBranchArgs,
   CreateCommitArgs,
   CreatePullRequestForBranchArgs,
   CreatePullRequestFromFilesArgs,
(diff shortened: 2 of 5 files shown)


packages/amplication-git-utils/src/git-provider.interface.ts.ts
```diff
@@ -45,4 +45,5 @@ export interface GitProvider {
   getBranch: (args: GetBranchArgs) => Promise<Branch>;
   createBranch: (args: CreateBranchArgs) => Promise<Branch>;
   getFirstCommitOnBranch: (args: GetBranchArgs) => Promise<Commit>;
+  getMyCommitsList: (args: GetBranchArgs) => Promise<Commit[]>;
```
yuval-hazaz, line 48:
```suggestion
  getCurrentUserCommitList: (args: GetBranchArgs) => Promise<Commit[]>;
```

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -60,6 +62,121 @@ export class GithubService implements GitProvider {
...
+    return {
+      commits: nodes.map(({ oid }) => ({
+        sha: oid,
+      })),
+      hasNextPage,
+      endCursor,
+    };
+  }
+
+  async getAmplicationBotData(): Promise<AmplicationBotData> {
```
yuval-hazaz, line 164: this should be internal to the provider

packages/amplication-git-utils/src/types.ts
```diff
@@ -152,3 +152,8 @@ export interface CreateBranchArgs {
 export interface Commit {
   sha: string;
 }
+
+export interface AmplicationBotData {
+  id: string;
+  login: string;
+}
```
yuval-hazaz, line 159: this should be internal to the provider

packages/amplication-git-utils/src/types.ts
```diff
@@ -152,3 +152,8 @@ export interface CreateBranchArgs {
 export interface Commit {
   sha: string;
 }
+
+export interface AmplicationBotData {
```
yuval-hazaz, line 156: change the name to something along the lines of GitHubUser
```suggestion
export interface GitHubUser {
```

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -60,6 +62,121 @@ export class GithubService implements GitProvider {
...
+          botData,
+          branchName,
+          owner,
+          repositoryName,
+          cursor,
+        });
+      moreCommitsPagination = hasNextPage;
+      commitsList = commitsList.concat(commits); // The list is in ascending order ( newest commits are first )
+      cursor = endCursor;
+    } while (moreCommitsPagination);
```
yuval-hazaz, line 85: Should we add another stop here for extreme cases? (e.g. 100 pages?) 
(maybe in the future, we can think of a way to create a snapshot that includes x commits from more than X months ago, so we will not increase the number of commits infinitely)

yuval-hazaz (changes requested):
well done! some minor comments

packages/amplication-git-utils/src/git/github.service.ts
tupe12334: Its a good idea, but its a separated issue form this

packages/amplication-git-utils/src/git/github.service.ts
tupe12334: I disagree

packages/amplication-git-utils/src/types.ts
tupe12334: I disagree

Issue #3638, opened 2022-08-30
🐛 Bug Report: Commit page - Wrong padding between selection


alexbass86:
### What happened?

Wrong padding between selections in our commit screen. 
Let's ensure we have a padding of 8px between each commit selection.

Bad:
![image](https://user-images.githubusercontent.com/96179897/187361738-79e2711d-0396-44e0-ba09-8517a8b9af04.png)

How it should be:
![image](https://user-images.githubusercontent.com/96179897/187361821-6219e081-30a4-4c48-bfdb-300170f25437.png)

Also, please make sure that the profile and text have 8px padding
![image](https://user-images.githubusercontent.com/96179897/187362005-fec5d24d-61bd-4738-9e56-29c702c1adf6.png)

### What you expected to happen

Same padding values as in the design

### How to reproduce

Go to the commit screen.

### Amplication version

0.15.1

### Are you willing to submit PR?

Yes I am willing to submit a PR!

### Plan
Go to "CommitList" component which is located at "amplication/packages/amplication-client/src/VersionControl/CommitList.tsx"
Align component and subcomponents styles according to design
Here is the design:
https://www.figma.com/file/71Ut91hFxC7cpLbJQQMsgP/%233141---Project-Commit-Flow?node-id=836%3A18964&t=U9JtPFH17ehqH0BX-1

Plan was written by @EugeneTseitlin

abrl91:
Moved to "reopen" due to a bug in the sidebar on each page that we use the sidebar. It should be wider


![Image](https://user-images.githubusercontent.com/39680385/217226238-b0957233-60ca-4b08-ac19-0f68691163ec.png)




![Image](https://user-images.githubusercontent.com/39680385/217226367-17244419-d617-4a3b-9fd2-c121009db166.png)

overbit:
## Tests result report (for PR #5145)

### Scenarios
	
1. Scenario description: Login and browse on difference pages
	 Status: 🟥  `FAILED`
	 Proof of execution:  Tested on Firefox and Safari
	 `Sandbox` 
	 <img width="1273" alt="Screenshot 2023-02-07 at 16 49 05" src="https://user-images.githubusercontent.com/2861984/217310370-aa81a50d-be3b-4d50-b7de-4e9ca8d385b5.png">
	 
	 
	 `Production` 
[...]

PR #5145, opened 2023-02-07
align commit list according to design

EugeneTseitlin:
Close: #3638 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Layout/PageContent.scss
+++ b/packages/amplication-client/src/Layout/PageContent.scss
@@ -1,5 +1,7 @@
 @import "../style/index.scss";
 
+$tabs-width: 300px;
+
 .amp-page-content {
   display: flex;
   flex: 1;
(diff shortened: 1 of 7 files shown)


abrl91 (approved):

packages/amplication-client/src/Layout/PageContent.scss
```diff
@@ -1,5 +1,7 @@
 @import "../style/index.scss";
 
+$tabs-width: 300px;
```
overbit, line 3: @barshimi @EugeneTseitlin how are we currently handling responsive design? 
Are we declaring measurements in `px`?

barshimi: it depends... if it's locally to the specific component/page we declare it as local variable, if it's not we added it to the global variables

EugeneTseitlin: @overbit Depends on the situation. Not every component is responsive. This one, for example, is not. Left and right bars have static width, while middle part is responsive. We also have a min screen width, which is 900px, if I'm not mistaken. 
Regarding this specific change, that's how it was before, this morning I accepted PR that removed this line, so now I return it back.

PR #5152, opened 2023-02-07
feat(git-utils): add strict null check

tupe12334:
This pull request adds a strict null check to protect us from unexpected behavior.


abrl91 (changes requested):
can you elaborate on why we need all the `| null`?

tupe12334:
@abrl91 because those functions can return null so we can show that in the function signature.

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -51,8 +51,17 @@ export class GithubService implements GitProvider {
...
+    }
     this.gitInstallationUrl = GITHUB_APP_INSTALLATION_URL;
+    if (!GITHUB_APP_APP_ID) {
+      throw new Error("GITHUB_APP_APP_ID is not defined");
+    }
     this.appId = GITHUB_APP_APP_ID;
+    if (!GITHUB_APP_PRIVATE_KEY) {
+      throw new Error("GITHUB_APP_PRIVATE_KEY is not defined");
+    }
     this.privateKey = GITHUB_APP_PRIVATE_KEY;
```
overbit, line 65: I like this 💯

overbit (approved):

abrl91 (approved):

Issue #5118, opened 2023-02-02
🐛 Bug Report: The event for a successful upgrade is not showing up on Amplitude


GreenMachine01:
A user upgraded his account but I am not seeing the event "WorkspacePlanUpgradeCompleted" on Amplitude
Event should be issued based on event that arrived from Stigg
WorkspacePlanUpgradeRequest and WorkspacePlanDowngradeRequest event should stay

### Plan
Go to amplication-server -> subscription.controller.ts
Add event tracker to updateStatus method for "subscription.created" and "subscription.updated" events types

Plan is written by @EugeneTseitlin

yuval-hazaz:
see comment on https://github.com/amplication/amplication/issues/5364

EugeneTseitlin:
> see comment on #5364

commented there too

PR #5157, opened 2023-02-08
added purchase completed event, removed purchase request event

EugeneTseitlin:
Close: #5118 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/App.tsx
+++ b/packages/amplication-client/src/App.tsx
@@ -2,7 +2,12 @@ import React, { useCallback, useState, useEffect } from "react";
 import { useLocation } from "react-router-dom";
 import * as reactHotkeys from "react-hotkeys";
 import ThemeProvider from "./Layout/ThemeProvider";
-import { track, dispatch, init as initAnalytics } from "./util/analytics";
+import {
+  track,
+  dispatch,
+  init as initAnalytics,
+  useTracking,
+} from "./util/analytics";
 import { init as initPaddle } from "./util/paddle";
 import { Routes } from "./routes/appRoutes";
 import { routesGenerator } from "./routes/routesUtil";
@@ -15,6 +20,7 @@ import {
 } from "@amplication/design-system";
 import useLocalStorage from "react-use-localstorage";
 import queryString from "query-string";
+import { AnalyticsEventNames } from "./util/analytics-events.types";
 
 declare global {
   interface Window {
@@ -97,6 +103,18 @@ function App() {
     ignoreTags: [],
   });
 
+  const closeConfirmation = useCallback(() => {
+    setWorkspaceUpgradeConfirmation(false);
+  }, []);
+
+  const { trackEvent } = useTracking();
+
+  const trackPurchaseCompleted = useCallback(() => {
+    trackEvent({
+      eventName: AnalyticsEventNames.WorkspacePlanUpgradeCompleted,
+    });
+  }, []);
+
   const showLoadingAnimation = keepLoadingAnimation || currentWorkspaceLoading;
 
   return (
@@ -112,8 +130,9 @@ function App() {
       {workspaceUpgradeConfirmation && (
         <PlanUpgradeConfirmation
           isOpen={workspaceUpgradeConfirmation}
-          onConfirm={() => setWorkspaceUpgradeConfirmation(false)}
-          onDismiss={() => setWorkspaceUpgradeConfirmation(false)}
+          onConfirm={closeConfirmation}
+          onDismiss={closeConfirmation}
+          onFirstRender={trackPurchaseCompleted}
         />
       )}
     </ThemeProvider>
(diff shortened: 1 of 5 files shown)


packages/amplication-client/src/App.tsx
```diff
@@ -97,6 +103,18 @@ function App() {
     ignoreTags: [],
   });
 
+  const closeConfirmation = useCallback(() => {
+    setWorkspaceUpgradeConfirmation(false);
+  }, []);
```
abrl91, line 108: What is the benefit of using useCallback while you don't have anything in the dependency array

packages/amplication-design-system/src/components/PlanUpgradeConfirmation/PlanUpgradeConfirmation.tsx
```diff
@@ -13,13 +13,19 @@ export type Props = DialogProps & {
...
 
 export const PlanUpgradeConfirmation = ({
   isOpen,
   onConfirm,
   onDismiss,
+  onFirstRender,
 }: Props) => {
+  useEffect(() => {
+    onFirstRender();
+  }, []);
```
abrl91, line 27: why do we need the dependency array here?

packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -162,12 +162,8 @@ export class WorkspaceResolver {
     nullable: true,
   })
   async provisionSubscription(
-    @UserEntity() currentUser: User,
```
abrl91, line 165: Are you sure we need to delete it?

packages/amplication-client/src/App.tsx
```diff
@@ -112,8 +130,9 @@ function App() {
       {workspaceUpgradeConfirmation && (
         <PlanUpgradeConfirmation
           isOpen={workspaceUpgradeConfirmation}
-          onConfirm={() => setWorkspaceUpgradeConfirmation(false)}
-          onDismiss={() => setWorkspaceUpgradeConfirmation(false)}
+          onConfirm={closeConfirmation}
+          onDismiss={closeConfirmation}
```
abrl91, line 134: I don't think the onDismiss should use the same event as onComfirm

abrl91 (changes requested):

packages/amplication-server/src/core/billing/billing.service.ts
```diff
@@ -208,16 +201,6 @@ export class BillingService {
...
-    await this.analytics.track({
-      userId,
-      properties: {
-        workspaceId,
-      },
-      event:
-        intentionType === "DOWNGRADE_PLAN"
-          ? EnumEventType.WorkspacePlanDowngradeRequest
-          : EnumEventType.WorkspacePlanUpgradeRequest,
-    });
```
yuval-hazaz, line 220: I don't think we should remove this event.
It provides us visibility from the server side of the process, so we can understand if there are any blockers after that point

packages/amplication-client/src/App.tsx
yuval-hazaz: The benefit is that the function will only be declared once instead of every run

packages/amplication-client/src/App.tsx
```diff
@@ -97,6 +103,18 @@ function App() {
...
+  const closeConfirmation = useCallback(() => {
+    setWorkspaceUpgradeConfirmation(false);
+  }, []);
+
+  const { trackEvent } = useTracking();
+
+  const trackPurchaseCompleted = useCallback(() => {
+    trackEvent({
+      eventName: AnalyticsEventNames.WorkspacePlanUpgradeCompleted,
+    });
```
yuval-hazaz, line 115: I expect to see this event in the server side and not in the client side

yuval-hazaz (changes requested):
@EugeneTseitlin I think we should take a different approach for this task.
I expect the current event to stay, and we need to add the new revert from the server side when the actual plan in ready and not from the client side

packages/amplication-design-system/src/components/PlanUpgradeConfirmation/PlanUpgradeConfirmation.tsx
EugeneTseitlin: same as in useCallback, empty dependency array cause only one effect call.

packages/amplication-server/src/core/workspace/workspace.resolver.ts
EugeneTseitlin: yep, it's not in use anymore

packages/amplication-client/src/App.tsx
EugeneTseitlin: that's by intention, the idea is to close purchase confirmation.

packages/amplication-server/src/core/billing/billing.service.ts
EugeneTseitlin: Synced with Erez about this suggestion, we've decided to keep these events. Please, see the updated issue.

packages/amplication-client/src/App.tsx
EugeneTseitlin: It's going to be changed.

EugeneTseitlin:
> @EugeneTseitlin I think we should take a different approach for this task. I expect the current event to stay, and we need to add the new revert from the server side when the actual plan in ready and not from the client side

Synced with Erez about this suggestion, we've decided to keep existing events and add "PurchaseCompleted" based on event that received from Stigg. Please, see the updated issue.

Issue #5007, opened 2023-01-18
As User - I want to use a new "Help" menu in the top bar


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/211553494-d7c79f61-1a29-4339-82a3-189c04b08ad5.png)

**Design** - https://www.figma.com/file/dVFO687AHmy2JyxQMSuJIx/HubSpot-chat-within-the-app?node-id=836%3A18964&t=tf5OIU85JiDmYuCq-1

**Requirements:**
1. Add "Help" CTA that opens a menu in the top panel
3. Move the "Docs" link to be under the new "Help" menu
4. Move the "Contact Us" link to be under the new "Help" menu
5. Add "Technical Support" link that will open the proper channel on Discord - https://discord.com/channels/757179260417867879/1063337571184279582

aarushisoni:
Hi, is this issue still open?
I want to work on this issue; please assign me to this.

GreenMachine01:
@aarushisoni assigned to you. good luck. if you don't have access to the design, let me know.

aarushisoni:
> @aarushisoni assigned to you. good luck. If you don't have access to the design, let me know.

Thank you for assigning me this issue. Can you please provide me access to this design link?

GreenMachine01:
@alexbass86 please see above.

aarushisoni:
> @aarushisoni assigned to you. good luck. if you don't have access to the design, let me know.

Hi! Can you please tell me in which file I need to make changes?

GreenMachine01:
@EugeneTseitlin can you please assist?

jaysoni-ash42:
@EugeneTseitlin  can i work on this issue?

jaysoni-ash42:
@GreenMachine01  can i have the design access ?

souravjain540:
> @GreenMachine01  can i have the design access ?
> 
> 

Sure. @alexbass86 please do

alexbass86:
> > @GreenMachine01  can i have the design access ?
> 
> Sure. @alexbass86 please do

https://www.figma.com/file/tIfcueoxU4DtYtCHmmLuE7/Community---Task-Assets?node-id=1509%3A22444&t=B1RR7JUqgIUud0pp-1

Please use this public link, thank you for your contribution.

GreenMachine01:
@alexbass86 please review.
looks good to me.



[...]

PR #5168, opened 2023-02-09
As User - I want to use a new "Help" menu in the top bar

gitstart:
Closes https://github.com/amplication/amplication/issues/5007
### DEMO_VIDEO
Screenshot 2023-02-10 at 15 30 51

https://www.loom.com/share/54d89adaf7cf4810a4187cc45f33a389


packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -27,6 +27,20 @@ const CLASS_NAME = "workspace-header";
...
+type HelpMenuItem = {
+  name: string;
+  url: string | null;
+};
+
+const HELP_MENU_LIST: HelpMenuItem[] = [
+  { name: "Docs", url: "https://docs.amplication.com" },
+  {
+    name: "Technical Support",
+    url: "https://discord.com/channels/757179260417867879/1063337571184279582",
```
yuval-hazaz, line 39:
```suggestion
    url: "https://amplication.com/discord",
```

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.scss
```diff
@@ -128,6 +128,65 @@ $select-menu-max-width: 300px;
...
+    .select-menu__summary.amp-button.amp-button--text {
+      color: var(--black80);
+      &.highlight,
+      &:hover,
+      &:focus,
+      &:active {
+        color: var(--black100);
+      }
+    }
+
```
yuval-hazaz, line 144: @EugeneTseitlin can you please have a look at this?
Do we need all the CSS locally? I expect the styles to be inherited from the component directly from the design system

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -278,6 +277,41 @@ const WorkspaceHeader: React.FC<{}> = () => {
...
+            className={`${CLASS_NAME}__help_popover__menu`}
+          >
+            <SelectMenuModal align="right">
+              <SelectMenuList>
+                {HELP_MENU_LIST.map((route: HelpMenuItem, index) => (
+                  <SelectMenuItem
+                    closeAfterSelectionChange
+                    onSelectionChange={() => {
+                      !route.url && handleContactUsClick();
+                    }}
```
yuval-hazaz, line 295: This solution is OK only for the current situation where we have a single item with a callback, but we need to implement a more generic solution.

we can use `itemData` on SelectMenuItem to pass a name of a command and then in a callback function trigger a function based on the command name

{ name: "Contact Us", url: null, itemData: COMMAND_CONTACT_US },

yuval-hazaz (changes requested):
Thanks for your support! good job!
please fix based on the comments before we can merge

gitstart:
> Thanks for your support! good job! please fix based on the comments before we can merge

Hi @yuval-hazaz We've resolved 1st and and 3rd feedback, for the 2nd one do we have to do attend to do that feedback ?

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.scss
yuval-hazaz: @EugeneTseitlin ?

yuval-hazaz (approved):

PR #5169, opened 2023-02-09
fix(server): remove password fields from user entity

morhag90:
part of: #4431 
related comment: https://github.com/amplication/amplication/issues/4431#issuecomment-1423991692
the first issue that was raised there:
image

## PR Details

Remove password fields from the user entity
## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/constants.ts
+++ b/packages/amplication-server/src/core/entity/constants.ts
@@ -136,42 +136,6 @@ export const DEFAULT_ENTITIES: EntityData[] = [
           maxLength: DEFAULT_SINGLE_LINE_TEXT_MAX_LENGTH,
         },
       },
-      {
-        dataType: EnumDataType.Username,
-        name: "username",
-        displayName: "Username",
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/entity/constants.ts
```diff
@@ -136,42 +136,6 @@ export const DEFAULT_ENTITIES: EntityData[] = [
...
-        displayName: "Username",
-        description:
-          "An automatically created field of the username of the user",
-        unique: false,
-        required: true,
-        searchable: true,
-        properties: {
-          maxLength: DEFAULT_SINGLE_LINE_TEXT_MAX_LENGTH,
-        },
-      },
```
abrl91, line 151: Why do we need to remove the username field?

morhag90: @abrl91 When you don't work with auth you don't need userName (there is no login)

abrl91 (approved):

Issue #5163, opened 2023-02-09
As a user - I don't want to be able to select only supported plugins that are not deprecated


overbit:
## Description

Currently, as a user, I can select any version of a plugin and it can cause issues in the generated code or in the code generation process where an old plugin is not supported by amplication DSG anymore.
In the case of existing applications that are configured with a plugin version that gets deprecated, the DSG build log should show a message about the deprecated plugin and a sentence with instructions on how to update it.

This requirement comes from the work done on #4431, which highlighted that some old plugins (pre-webpack) might not work with the DSG dynamic installation.

## Tech Notes

- [ ] Deprecate old plugins versions
- [x] Update plugin API to only collect NOT-deprecated packages versions #5207  
- [x] Update DSG to print a warning log in the build log wherever it tries to install a deprecated plugin #5171

## Acceptance criteria

user with a deprecated plugin configured already
* *Given* a user with an existing plugin configured to version `x.y.z`
* *When* it runs a new code generation build
* *Then* a new message will be printed warning the user that is using a deprecated plugin

user selects a specific version
* *Given* a user 
* *When* it goes to select a specific plugin version
* *Then* only no-deprecated plugin versions are displayed in the dropdown

abrl91:
passed - ✅


https://user-images.githubusercontent.com/39680385/219605506-6ad7884b-1a3a-40ab-876e-0d1476169e1e.mov

PR #5170, opened 2023-02-09
feat(plugin-api): filter out deprecated npm package versions

overbit:
Close: #5163 

## PR Details

Add logic to filter out deprecated npm package versions.
If deprecated versions have been cached in the DB they will not be deleted as part of this PR

## PR Checklist

### DIFF
--- a/packages/amplication-plugin-api/src/pluginVersion/npm-plugin-version.service.ts
+++ b/packages/amplication-plugin-api/src/pluginVersion/npm-plugin-version.service.ts
@@ -1,17 +1,7 @@
 import { Injectable } from "@nestjs/common";
 import { Plugin, PluginVersion } from "../../prisma/generated-prisma-client";
 import fetch from "node-fetch";
-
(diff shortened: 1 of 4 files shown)


morhag90 (approved):

packages/amplication-plugin-api/src/pluginVersion/npm-plugin-version.service.ts
```diff
@@ -64,12 +64,12 @@ export class NpmPluginVersionService {
         const npmResponse = await fetch(
           `https://registry.npmjs.org/${pluginNpmName}`
```
overbit, line 65: Unfortunately, there is not REST API with filter by `deprecated` or not packages

overbit:
@barshimi actually, it might make more sense to add a "deprecated" property to the pluginVersion entity and just filter on the client side. What do you think?

overbit:
Logic replaced by a better solution: #5207

PR #5182, opened 2023-02-12
fix(server): Add auth-strategy creation to install-plugin migration s…

morhag90:
Close: #4431 

## PR Details

Add auth-strategy plugin if it does not exist in install-plugin migration script
## PR Checklist

Test cases: 

Add new services. 
DO NOT install plugins: auth-core / auth-basic/ auth-jwt 
Run the migration script: install-auth-core-plugin with the following command: 
POSTGRESQL_URL=postgres://[user]:[password]@127.0.0.1:5432/app-database npx ts-node install-auth-core-plugin.ts 

Check that every new service you created in the first step has installed plugins: 
auth-core
auth-basic/ auth-jwt  (according to the auth provider type).

### DIFF
--- a/packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
+++ b/packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
@@ -13,6 +13,8 @@ import { EnumBlockType } from "../src/enums/EnumBlockType";
 import { PluginOrder } from "../src/core/pluginInstallation/dto/PluginOrder";
 import { PluginOrderItem } from "../src/core/pluginInstallation/dto/PluginOrderItem";
 import { PluginInstallation } from "../src/core/pluginInstallation/dto/PluginInstallation";
+import { ServiceSettings } from "../src/core/serviceSettings/dto";
+import { EnumAuthProviderType } from "../src/core/serviceSettings/dto/EnumAuthenticationProviderType";
 
 const CURRENT_VERSION_NUMBER = 0;
 const client = new PrismaClient();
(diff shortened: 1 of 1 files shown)


packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
```diff
@@ -90,31 +118,69 @@ async function main() {
...
+            settings: {},
+            version: "latest",
+            resource: {
+              connect: {
+                id: resource.id,
+              },
+            },
+          },
+        });
+        console.log({ AuthStrategyPlugin });
```
barshimi, line 180: remove console and I suggest to wrap it in try/catch

packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
```diff
@@ -82,6 +84,32 @@ async function main() {
...
+  ): boolean {
+    const results = plugins.filter((plugin) => {
+      return (
+        plugin.pluginId ===
+        authPluginNames.find((name) => name === plugin.pluginId)
+      );
+    });
+
+    if (results.length > 0) return true;
+    return false;
```
tupe12334, line 111: Same as above

packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
```diff
@@ -90,31 +118,69 @@ async function main() {
...
+            resource: {
+              connect: {
+                id: resource.id,
+              },
             },
           },
-        },
-      });
+        });
+        console.log({ newPlugin });
```
tupe12334, line 149:
```suggestion
```

packages/amplication-server/migration-scripts/install-auth-core-plugin.ts
```diff
@@ -82,42 +84,108 @@ async function main() {
...
+  function isPluginExist(
+    plugins: PluginInstallation[],
+    authProvider: string
+  ): boolean {
+    const plugin = plugins.find(
+      (plugin) => plugin.pluginId.trim() == authProvider
+    );
+
+    if (plugin) return true;
+    return false;
```
abrl91, line 96: we actually can use the `.some()` method that returns boolean: true if it found the item and false if it didn't find
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

```suggestion
    return plugins.some(
      (plugin) => plugin.pluginId.trim() == authProvider
    );
```

abrl91: But it's just a suggestion...we can leave it like that

abrl91 (approved):

Issue #5104, opened 2023-02-01
🐛 Bug Report: DSG e2e tests on the generated app are not running


overbit:
### What happened?

Currently, the e2e test on the DSG-generated app is not running successfully due to some changes that were not reflected in the test code.

As part of this task, we also need to add the execution of DSG e2e tests in the ci GitHub workflow.
 

### What you expected to happen

When I run `npx nx e2e amplication-data-service-generator-runner`, tests start and complete successfully. 

### How to reproduce

Run `npx nx e2e amplication-data-service-generator-runner` 

### Amplication version

v1.2.3

### Are you willing to submit PR?

Yes I am willing to submit a PR!

overbit:
De-prioritised. WIP changes are available as part of this PR #5193

PR #5193, opened 2023-02-13
test(dsg): fix e2e test on generate apps

overbit:
Close: #5104 

## PR Details

This PR reintroduces some basic e2e tests against the generated app. 
It also fixes the issue about `.amplication` folder and the missing documentation reported here https://github.com/amplication/amplication/issues/5263
 
## PR Checklist


packages/data-service-generator/src/main.ts
```diff
@@ -28,36 +23,14 @@ async function readInputJson(filePath: string): Promise<DSGResourceData> {
...
-      new URL(
-        "build-runner/code-generation-success",
-        process.env.BUILD_MANAGER_URL
-      ).href,
-      {
-        resourceId: process.env.RESOURCE_ID,
-        buildId: process.env.BUILD_ID,
-      }
-    );
+    await generateCodeByResourceData(resourceData, destination);
```
abrl91, line 33: nice!

abrl91 (approved):

package.json
```diff
@@ -220,8 +220,10 @@
     "@typescript-eslint/eslint-plugin": "^5.39.0",
     "@typescript-eslint/parser": "^5.39.0",
     "babel-jest": "^28.1.0",
+    "base-64": "^1.0.0",
```
tupe12334, line 223: Don't we want to have the types? 
https://www.npmjs.com/package/@types/base-64

packages/data-service-generator/src/register-plugin.ts
```diff
@@ -34,7 +35,7 @@ async function* getPluginFuncGenerator(
       yield func.default;
     } while (pluginListLength > index);
   } catch (error) {
-    console.log(error); /// log error
+    logger.log(error); /// log error
```
tupe12334, line 38:
```suggestion
    logger.log(error);
```

tupe12334 (changes requested):

package.json
overbit: that is not necessary for now since it is used in only one line.

tupe12334 (approved):

packages/data-service-generator/src/create-data-service.ts
```diff
@@ -53,7 +53,7 @@ export async function createDataService(
   } catch (error) {
     await createLog({
       level: "info",
```
yuval-hazaz, line 55:
```suggestion
      level: "error",
```

yuval-hazaz (approved):
See minor comment.
Other than that, this is great and I would also like it to the next stage where we can create a matrix of different plugins and test different combination

PR #5206, opened 2023-02-14
Fix/create resource default plugins 

morhag90:
Close: #4431

## PR Details

Add a method to create many plugins and add as default plugins: auth-core and auth-jwt when creating a new service. 
## PR Checklist

Test cases:

create a new service
Go to plugins list => make sure those plugins: auth-core and auth-jwt already installed.


abrl91:
LGTM :) but let's wait for @barshimi to review the client part

packages/amplication-client/src/Plugins/hooks/usePlugins.ts
```diff
@@ -217,6 +220,30 @@ const usePlugins = (resourceId: string, pluginInstallationId?: string) => {
...
 
+  const [createPluginInstallations] = useMutation<{
+    createPluginInstallations: models.PluginInstallation[];
+  }>(CREATE_PLUGINS_INSTALLATION, {
+    onCompleted: (data) => {
+      data.createPluginInstallations.forEach((plugin) => {
+        addBlock(plugin.id);
+      });
+    },
+    refetchQueries: [
```
barshimi, line 231: is it possible to return from API the nee GET_PLUGIN_INSTALLATIONS instead of refetch it ? it will be much more efficient and will solve two issues in one. graphQl any way return the new result, why not combine them together. also, this call should happen only once as part of the migration, so I expect this to be separated and one time in any project

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -53,13 +54,20 @@ const useResources = (
   );
 
   const [currentResource, setCurrentResource] = useState<models.Resource>();
+
+  const [currentCreatedResourceId, setCurrentCreatedResourceId] =
```
barshimi, line 58: you don't need state, you can use refObject

packages/amplication-client/src/Workspaces/hooks/useResources.ts
```diff
@@ -129,6 +146,42 @@ const useResources = (
         });
     });
   };
+
+  useEffect(() => {
```
barshimi, line 150: we should talk

packages/amplication-client/src/Plugins/hooks/usePlugins.ts
morhag90: Remove the reftech queries

barshimi (approved):

PR #5225, opened 2023-02-16
Fix/5118 issue successful plan upgrade event when receive update from stigg

EugeneTseitlin:
Close: #5118 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/subscription/subscription.controller.ts
+++ b/packages/amplication-server/src/core/subscription/subscription.controller.ts
@@ -47,7 +47,7 @@ export class SubscriptionController {
         );
         const userId = updateStatusDto.metadata?.userId;
         if (
-          createSubscriptionInput.plan === EnumSubscriptionPlan.Pro &&
+          createSubscriptionInput.plan !== EnumSubscriptionPlan.Free &&
           userId
         ) {
           await this.analyticsService.track({
--- a/packages/amplication-server/src/core/subscription/subscription.controller.ts
+++ b/packages/amplication-server/src/core/subscription/subscription.controller.ts
@@ -45,9 +45,13 @@ export class SubscriptionController {
           updateStatusDto.id,
           createSubscriptionInput
         );
-        if (createSubscriptionInput.plan === EnumSubscriptionPlan.Pro) {
+        const userId = updateStatusDto.metadata?.userId;
+        if (
+          createSubscriptionInput.plan === EnumSubscriptionPlan.Pro &&
+          userId
+        ) {
           await this.analyticsService.track({
-            userId: "unknown",
+            userId: userId,
             properties: {
               workspaceId: updateStatusDto.customer.id,
             },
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/subscription/subscription.controller.ts
```diff
@@ -39,6 +45,15 @@ export class SubscriptionController {
           updateStatusDto.id,
           createSubscriptionInput
         );
+        if (createSubscriptionInput.plan === EnumSubscriptionPlan.Pro) {
+          await this.analyticsService.track({
+            userId: "unknown",
```
overbit, line 50: can we pass the real user id?

overbit (changes requested):

packages/amplication-server/src/core/subscription/subscription.controller.ts
EugeneTseitlin: I would love to, but it doesn't arrive from webhook. I've contacted Stigg in order to find out if it's possible to pass user id.

overbit (approved):

packages/amplication-server/src/core/subscription/subscription.controller.ts
```diff
@@ -39,6 +45,19 @@ export class SubscriptionController {
           updateStatusDto.id,
           createSubscriptionInput
         );
+        const userId = updateStatusDto.metadata?.userId;
+        if (
+          createSubscriptionInput.plan === EnumSubscriptionPlan.Pro &&
```
yuval-hazaz, line 50: I think this event should be triggered with every subscription creation, with the relevant metadata - for now, just "Pro" will work, but we will need to track other future plans in the same way. 
You should probably exclude "free"

packages/amplication-server/src/core/subscription/subscription.controller.ts
```diff
@@ -39,6 +45,19 @@ export class SubscriptionController {
...
+          userId
+        ) {
+          await this.analyticsService.track({
+            userId: userId,
+            properties: {
+              workspaceId: updateStatusDto.customer.id,
+            },
+            event: EnumEventType.WorkspacePlanUpgradeCompleted,
+          });
+        }
```
yuval-hazaz, line 60: Don't we have requirements to track the other events like update, expired, downgrade, canceled?

yuval-hazaz (changes requested):

packages/amplication-server/src/core/subscription/subscription.controller.ts
EugeneTseitlin: nope, but of cause now we can easily add them.

packages/amplication-server/src/core/subscription/subscription.controller.ts
EugeneTseitlin: I thought that the point here is to register purchase that client completed by himself. Which means that "Enterprise" plan should not be registgred.

EugeneTseitlin: Synced with Erez, event should be issued also when workspace was upgraded to "Enterprise" plan

packages/amplication-server/src/core/subscription/subscription.controller.ts
```diff
@@ -39,6 +45,19 @@ export class SubscriptionController {
...
```
yuval-hazaz, line 60: This should be in the service and not in the controller

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #5227, opened 2023-02-17
🐛 Bug Report: The subscription message is missing in the next environment


abrl91:
### What happened?

The subscription message is missing in the next environment
"_With the current plan, you can use up to 3 services. Upgrade now_"

### What you expected to happen

It should be as follows when the workspace's plan is "Free":

![image](https://user-images.githubusercontent.com/39680385/219603209-2be10cae-7dd1-4a0b-bcc7-e2b9535ca88b.png)

![image](https://user-images.githubusercontent.com/39680385/219603407-135333b0-a430-4532-9bd8-1cd28d139eba.png)

### How to reproduce

go to https://app.amplication-sandbox.com
make sure that you are in a Free Plan workspace
see there is no message

### Amplication version

next

abrl91:
related issue: https://github.com/amplication/amplication/issues/5233

EugeneTseitlin:
Didn't pass QA. 
Workspace with free plan actually has notification:
![image](https://user-images.githubusercontent.com/6080188/219947239-1fe9440f-9c7a-4833-8fd0-6b6d67961451.png)
But workspace with Pro plan too (it should not happen):
![image](https://user-images.githubusercontent.com/6080188/219947361-1d70cfe9-4f4c-481a-a161-750e02592391.png)

abrl91:
Look at my comment before. This could not be QAed yet

overbit:
## Test result
`PASSED`
<img width="1247" alt="Screenshot 2023-02-20 at 09 32 49" src="https://user-images.githubusercontent.com/2861984/220068209-e404c388-bd8d-47a8-b0d7-14129d0a047a.png">

Workspace with free plan actually has notification:

Workspace with Pro plan does not have notifications as expected:
<img width="1238" alt="Screenshot 2023-02-20 at 09 32 38" src="https://user-images.githubusercontent.com/2861984/220068177-a78c0abf-ab74-4534-8a0c-cf3fa3b565bf.png">

PR #5229, opened 2023-02-17
Fix(client): free plan message

abrl91:
Close: #5227 

## PR Details
- update our models with the new value of the subscription plan enum (business => free)
- update the relevant code in the client  

## PR Checklist

- [ ] Tests for the changes have been added
- [x] `npm test` doesn't throw any error

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/amplication/amplication/blob/master/CODE_OF_CONDUCT.md) file for detailed contributing guidelines.

PR #5236, opened 2023-02-19
Fix/5227 correct validation for free pan

EugeneTseitlin:
Close: #5227 

## PR Checklist


abrl91 (commented):
@EugeneTseitlin this is how I did it in the beginning but then @overbit and myself noticed that we can do it better with what Dani commented it the issue

https://github.com/amplication/amplication/pull/5229

abrl91 (approved):

Issue #5112, opened 2023-02-02
🔍 Epic: [Smart git] strategy


tupe12334:
## Details

As we move forward with our smart git strategy implementation, we have a list of goals in mind:

- Amplication will change only files that have been changed in the amplication side from the previous commit
- If there was a conflict between amplication change and user change the conflict will show to the user in the git provider site
- When the user resolves a conflict, the conflict will not show again until amplication changes the line in question
- Amplication will support backward compatibility with the basic git strategy

The strategy works with some essentials concepts like git cherry-picking and git patches, the entire logic is explained in the https://github.com/amplication/amplication/pull/5096 file

## Sub Epics

- [x] #4590
- [x] #5113
- [x] #5114
- [x] #5115
- [x] #5272
- [x] #5270
- [x] #5315

PR #5241, opened 2023-02-20
Feat/git smart strategy

tupe12334:
Close: #5112 

## PR Checklist


packages/amplication-git-utils/src/git/git-client.ts
```diff
@@ -0,0 +1,44 @@
+import { simpleGit, SimpleGit } from "simple-git";
+
+export class GitClient {
+  git: SimpleGit;
+
+  constructor() {
+    this.git = simpleGit();
+  }
+
+  async clone(cloneUrl: string, cloneDir: string): Promise<void> {
```
barshimi, line 10: don't you want to return on any phase some value or indication of the action ?

packages/amplication-git-utils/src/git/git-client.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  async resetState() {
+    await this.git.fetch(["--all"]);
+    await this.git.pull();
+    await this.git.reset(["--hard"]);
+  }
+
+  async merge({ into, from }: { into: string; from: string }) {
+    console.log(`Merging branch: ${from} into: ${into}`);
```
barshimi, line 38: remove console

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -76,6 +85,7 @@ export class GitClientService {
       amplicationIgnoreManger
     );
 
+    console.log(`Got a ${pullRequestMode} pull request mode`);
```
barshimi, line 88: remove console

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
     }
 
     if (pullRequestMode === EnumPullRequestMode.Accumulative) {
-      await this.provider.createBranchIfNotExists({
+      const gitClient = new GitClient();
```
barshimi, line 102: wrap it with try/catch

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+      );
+
+      const { diff } = await this.preCommitProcess({
+        branchName,
+        gitClient,
+        owner,
+        repositoryName,
+      });
+      let fullDiffPath = "";
+      if (diff) {
```
barshimi, line 146: you can take any if and put them in a function and pass it the diff. it will be cleaner and without diff. another option is composing all this actions in function. it will give you the option to add new actions in the lineup process and much cleaner and decoupled

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+    throw new InvalidPullRequestMode();
+  }
+
+  private async preCommitProcess({
+    gitClient,
+    branchName,
+    owner,
+    repositoryName,
+  }: PreCommitProcessArgs): PreCommitProcessResult {
+    console.log("Pre commit process");
```
barshimi, line 213: I would replace the console with the new logger. it will give us context and not just empty string

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+      })
+      .push();
+  }
+
+  private async restoreAmplicationBranchIfNotExists(
+    args: CreateBranchIfNotExistsArgs
+  ): Promise<Branch> {
+    const { branchName, owner, repositoryName, gitClient } = args;
+    const branch = await this.provider.getBranch(args);
+    if (!branch) {
```
barshimi, line 261: I would prefer: if (branch) return branch;

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+  private async cherryPickCommits(
+    commits: Commit[],
+    gitClient: GitClient,
+    branchName: string,
+    firstCommitOnDefaultBranch: Commit
+  ) {
+    await gitClient.resetState();
+    await gitClient.checkout(branchName);
+
+    for (let index = commits.length - 1; index >= 0; index--) {
```
barshimi, line 300: move commits.length outside the for loop, so it wont calculate it any iteration

tupe12334:
As of now we will leave the console logs as they are, and we will open a seperated issue about implimanting a logger here. @barshimi

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+      const { diff } = await this.preCommitProcess({
+        branchName,
+        gitClient,
+        owner,
+        repositoryName,
+      });
+      let fullDiffPath = "";
+      if (diff) {
+        await mkdir(diffFolder, { recursive: true });
+        const diffPath = join(diffFolder, "diff.patch");
```
overbit, line 148: maybe it is worth to merge this if statement with the following one since the `this.provider.createCommit` function in the middle is not affected by the results of this block.

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+      await rm(cloneDir, { recursive: true, force: true });
+
       const { defaultBranch } = await this.provider.getRepository({
         owner,
         repositoryName,
       });
+
       const existingPullRequest = await this.provider.getPullRequestForBranch({
         owner,
         repositoryName,
```
overbit, line 179: Let's avoid setting `undefined` as expected values. We should try to use `null` (it will affect the interface too)

packages/amplication-git-utils/src/git/git.service.ts
```diff
@@ -89,39 +99,213 @@ export class GitClientService {
...
+      const { defaultBranch } = await this.provider.getRepository(args);
+      const firstCommitOnDefaultBranch =
+        await this.provider.getFirstCommitOnBranch({
+          owner,
+          repositoryName,
+          branchName: defaultBranch,
+        });
+      const branch = await this.provider.createBranch({
+        owner,
+        branchName,
```
overbit, line 271: let's try to avoid using `const` on the same var name (even it lives in a different context). This will help code readablity

packages/amplication-git-utils/src/git/git-client.ts
```diff
@@ -0,0 +1,44 @@
...
+    await this.git.checkout(branchName);
+  }
+
+  async resetState() {
+    await this.git.fetch(["--all"]);
+    await this.git.pull();
+    await this.git.reset(["--hard"]);
+  }
+
+  async merge({ into, from }: { into: string; from: string }) {
```
overbit: Let's remove unused functions

packages/amplication-git-utils/src/git/github.service.ts
```diff
@@ -41,25 +48,154 @@ export class GithubService implements GitProvider {
...
+  }
+
+  private async paginatedCommitsList(
+    args: GetBranchArgs & {
+      cursor: string | undefined;
+      botData: GitUser;
+    }
+  ): Promise<{ commits: Commit[]; hasNextPage: boolean; endCursor: string }> {
+    const { branchName, owner, repositoryName, cursor, botData } = args;
+
```
overbit, line 115: It is usually better to define a variable with the pagination size for code readability.

overbit (changes requested):

packages/amplication-git-utils/src/git/git-client.ts
tupe12334: No, because the success of this method is for itself the indication

packages/amplication-git-utils/src/git/git.service.ts
tupe12334: WOW, i didnt knew about that, thanks!

tupe12334: But for this senerio it's not relevant because its use it only in the first time of the index creation

packages/amplication-git-utils/src/git/git.service.ts
tupe12334: Why? i want it to crash if there is no git client

barshimi (approved):

overbit (approved):

Issue #5259, opened 2023-02-23
Requests for creating PR should be handled consequentially per repository


EugeneTseitlin:
### Feature description

Requests for creating PR to the same repo has to be handled consequentially, otherwise request would enter to race condition and operation might fail.
It's most probable reason of the following issue: #5224
The solution is to set key to messages that are sent to CREATE_PR_REQUEST_TOPIC. This way messages will enter to the same partition in a topic and are going to be consumed by pull request service consequentially.

EugeneTseitlin:
Demo: https://amplic-my.sharepoint.com/:v:/g/personal/eugene_amplication_com/EWfms6vLa19HvXHb1Vhg5yYBTiaOXretS5sHKrnWH-phSw?e=KSWErH
As we can see there are 3 builds of a project with 5 services successfully finished. There're no exceptions or errors.

PR #5260, opened 2023-02-23
Feat/5259 requests for creating pr should be handled consequentially per repository

EugeneTseitlin:
Close: #5259

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -491,8 +491,9 @@ export class BuildService {
             pullRequestMode,
           };
 
-          await this.queueService.emitMessage(
+          await this.queueService.emitMessageWithKey(
             this.configService.get(Env.CREATE_PR_REQUEST_TOPIC),
+            resourceRepository.id,
             JSON.stringify(createPullRequestArgs)
           );
         } catch (error) {
--- a/packages/amplication-server/src/core/queue/queue.service.ts
+++ b/packages/amplication-server/src/core/queue/queue.service.ts
@@ -22,4 +22,21 @@ export class QueueService {
       });
     });
   }
+
+  async emitMessageWithKey(
+    topic: string,
+    key: string,
+    value: string
+  ): Promise<void> {
+    return await new Promise((resolve, reject) => {
+      this.kafkaClient.emit(topic, { key, value }).subscribe({
+        error: (err: any) => {
+          reject(err);
+        },
+        next: () => {
+          resolve();
+        },
+      });
+    });
+  }
 }


EugeneTseitlin:
Keys and partitions are the same for "create PR" messages that target same repository:
image

EugeneTseitlin:
Demo: [https://amplic-my.sharepoint.com/✌️/g/personal/eugene_amplication_com/EWfms6vLa19HvXHb1Vhg5yYBTiaOXretS5sHKrnWH-phSw?e=KSWErH](https://amplic-my.sharepoint.com/:v:/g/personal/eugene_amplication_com/EWfms6vLa19HvXHb1Vhg5yYBTiaOXretS5sHKrnWH-phSw?e=KSWErH)
As we can see there are 3 builds of a project with 5 services successfully finished. There're no exceptions or errors.

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -491,8 +491,9 @@ export class BuildService {
             pullRequestMode,
           };
 
-          await this.queueService.emitMessage(
+          await this.queueService.emitMessageWithKey(
             this.configService.get(Env.CREATE_PR_REQUEST_TOPIC),
+            resourceRepository.id,
             JSON.stringify(createPullRequestArgs)
```
overbit, line 497: There is now a library for standardise the approach. Try to use that instead of code in each service.
`@amplication/util/nestjs/kafka`

overbit (changes requested):

packages/amplication-server/src/core/build/build.service.ts
EugeneTseitlin: I've checked util kafka library, I didn't find appropriate service/method. Can you, please, send a link to example of it's usage?

overbit: As per best practices, instructions on how to use a library are in the library README https://github.com/amplication/amplication/blob/master/libs/util/nestjs/kafka/README.md

EugeneTseitlin: because we never used this library before it has to be done and tested separately. I've opened a new issue for that:
https://github.com/amplication/amplication/issues/5287

overbit: That was exactly the point of using the library. Since you implemented a new function that is not covered by any unit test, please try to either use the existing one or create unit tests for this.  
Another option is to update the library with any new implementation so we are not making it obsolete.

etseitlin: But it's not a new usage. Replacement of queue service for kafka producer service will affect other flows in build service too and not only. There should be only one producer per service. So I see two options here. The first option is to add another producer form lib nestjs/kafka just for this specific case and there're going to be two producers in server. The second is to replace queue service with kafka producer service in server, which will affect other flows. What should I do?

EugeneTseitlin: But it's not a new usage. Replacement of queue service for kafka producer service will affect other flows in build service too and not only. There should be only one producer per service. So I see two options here. The first option is to add another producer form lib nestjs/kafka just for this specific case and there're going to be two producers in server. The second is to replace queue service with kafka producer service in server, which will affect other flows. What should I do?

overbit: Ok then, let's go with this and standardise with the lib later

overbit (approved):

Issue #4836, opened 2023-01-04
As User - I want to be able to downgrade my workspace plan (MVP)


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/210587920-fe10a0be-cea9-42c0-a78f-e743d1afa682.png)

**Requirements:**
1. Clicking the downgrade CTA to open the hubspot chat ("downgrade flow")

### Plan
Go to "PurchasePage" component which is located at amplication/packages/amplication-client/src/Purchase/PurchasePage.tsx
In function onPlanSelected add another case for "plan-amplication-free"
Implement is similarly to Pro plan
Also it's better to check that the plan name is correct at Stigg: https://app.stigg.io

Written by @EugeneTseitlin

GreenMachine01:
Passed QA.



![Image](https://user-images.githubusercontent.com/112330500/227173969-d7a24fd4-fbff-40a8-80e2-7330dd4a55ba.png)

PR #5357, opened 2023-03-07
changed contact us chat trigger, added downgrade chat trigger

EugeneTseitlin:
Close: #4836 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Purchase/PurchasePage.tsx
+++ b/packages/amplication-client/src/Purchase/PurchasePage.tsx
@@ -70,6 +70,7 @@ const PurchasePage = (props) => {
     });
 
   const handleContactUsClick = useCallback(() => {
+    history.push("?contact-us=true");
     openHubSpotChat();
     trackEvent({
       eventName: AnalyticsEventNames.ContactUsButtonClick,
@@ -77,6 +78,11 @@ const PurchasePage = (props) => {
       workspaceId: currentWorkspace.id,
     });
   }, [openHubSpotChat, currentWorkspace.id]);
+
+  const handleDowngradeClick = useCallback(() => {
+    history.push("?downgrade=true");
+    openHubSpotChat();
+  }, [openHubSpotChat]);
 
   const [isLoading, setLoading] = useState(false);
 
@@ -116,6 +122,9 @@ const PurchasePage = (props) => {
         case "plan-amplication-pro":
           setLoading(true);
           await upgradeToPro(selectedBillingPeriod, intentionType);
+          break;
+        case "plan-amplication-free":
+          handleDowngradeClick();
           break;
       }
     },
--- a/packages/amplication-client/src/Purchase/PurchasePage.tsx
+++ b/packages/amplication-client/src/Purchase/PurchasePage.tsx
@@ -70,6 +70,7 @@ const PurchasePage = (props) => {
     });
 
   const handleContactUsClick = useCallback(() => {
+    // This query param is used to open HubSpot chat with the main flow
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -78,6 +79,11 @@ const PurchasePage = (props) => {
     });
   }, [openHubSpotChat, currentWorkspace.id]);
 
+  const handleDowngradeClick = useCallback(() => {
+    history.push("?downgrade=true");
```
yuval-hazaz, line 83: @EugeneTseitlin, I assume the URL is required for Hubspot to identity the specific chat flow to open? if that is correct, please make sure it is documented here in the code otherwise, someone in the future can easily remove it without understanding the need.

packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -70,6 +70,7 @@ const PurchasePage = (props) => {
     });
 
   const handleContactUsClick = useCallback(() => {
+    history.push("?contact-us=true");
```
yuval-hazaz, line 73: Not part of this PR - but I think the event's name is not good enough - "contact us" is generic and already used in the top bar. 
Please change it to "PurchaseContactUsButtonClicked" (new event - do not change the name of the current event if it is used in other places)

packages/amplication-client/src/Purchase/PurchasePage.tsx
yuval-hazaz: also, please add event "PurchaseDowngradeClick"

yuval-hazaz (changes requested):

yuval-hazaz:
@GreenMachine01 please have a look at the comments here about the events, and see if it is aligned with your plan

GreenMachine01:
@yuval-hazaz I think we went with the generic CTA click event that has a param for the action
image

yuval-hazaz:
> @yuval-hazaz I think we went with the generic CTA click event that has a param for the action image

Isn't that the case when we decided to go with separate events because we wanted to understand each one separately?

packages/amplication-client/src/Purchase/PurchasePage.tsx
EugeneTseitlin: > also, please add event "PurchaseDowngradeClick"

This event has already exist.

GreenMachine01:
@yuval-hazaz we agreed after this event was already out. we didn't go back to change it.
we can open a new ticket with the requested changes.

packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -117,6 +123,9 @@ const PurchasePage = (props) => {
           setLoading(true);
           await upgradeToPro(selectedBillingPeriod, intentionType);
           break;
+        case "plan-amplication-free":
```
barshimi, line 126: why we're keep using switch/case it's a bad habit.

packages/amplication-client/src/Purchase/PurchasePage.tsx
yuval-hazaz: @EugeneTseitlin 
comments in code

EugeneTseitlin: comments are added, event that you meant is "PricingPageCTAClick" with "intentionType": "dowgrade" is issed on line 111.

yuval-hazaz (approved):

Issue #5220, opened 2023-02-16
🐛 Bug Report: Base directories of Message broker resource is throwing error


lokeswaran-aj:
### What happened?

When I tried to view the base directories of the Message broker resource, I could not view the directory. Instead, an error was thrown `Cannot return null for non-nullable field Query.projectConfigurationSettings` I m not sure if this error is intentional or a bug as this is the first time I created a message broker resource
<img width="1491" alt="Screenshot 2023-02-16 at 5 31 26 PM" src="https://user-images.githubusercontent.com/74011196/219360488-1138acb8-e6be-41d2-ab77-8a60ce551649.png">

### What you expected to happen

View the base directories of the message broker resource.

### How to reproduce

- create a message broker resource
- click on the `Settings` icon in the left bar
- click on the `Base directories`

### Amplication version

v1.2.3

### Environment

OS: macOS
Browser: Chrome

kabhamo:
Hi @overbit , @lokeswaran-aj 
After reviewing the code, I believe I can fix the bug, I will be happy if you can assign the issue to me.
Thank you :)

lokeswaran-aj:
Thanks @kabhamo But this issue was assigned to me last night only. Give me a day or two to work in this. If I cant fix it, I'll let you know

kabhamo:
No problem, by the way, my PR is ready.
Waiting for an update from you @lokeswaran-aj.
have a good day 🚀

lokeswaran-aj:
@kabhamo Sorry, please go ahead with your PR, I was confused with another issue. I created this issue for the community🤞

kabhamo:
@lokeswaran-aj thanks for the update.

kabhamo:
@overbit, please let me know if you are interested to assign this issue to me.
So I can create the PR.
have a good day

souravjain540:
> @overbit, please let me know if you are interested to assign this issue to me.
> 
> So I can create the PR.
> 
> have a good day 

Assigned:) let us know if we can help in anything

PR #5375, opened 2023-03-08
Fix/5220 message broker base directory

kabhamo:
Close: #5220

## PR Details

When the user tried to view the base directories of the Message broker resource an error was thrown `Cannot return null for non-nullable field Query.projectConfigurationSettings`.

The reason behind this error was about using the wrong **Resource Id** which was thrown from `projectConfigurationError` in `useProjectConfigSettingsHook.tsx` hook

By replacing the **Resource id** with `projectConfigurationResource?.id` as the query asks for. The bug was solved.
After doing some tests I figured out that the `updateResourceSettings` query also returns a similar error so I replace the id with the relevant one mentioned at the top. Ended with the GET and UPDATE queries working as wanted.

NOTE: The last commit was about solving some conflicts after updating my branch. 

Waiting for feedbacks :)


morhag90 (approved):

tupe12334:
@kabhamo  Can you please remove the package-lock from this pull request?

tupe12334 (changes requested):
The logic is working and approved but can you please remove the change of the package-lock before we merge?

tupe12334:
@kabhamo  Not removing it, i meant removing the change from the pull request, so that the package-lock will not change

yuval-hazaz (approved):

Issue #5261, opened 2023-02-24
As User - I want to sync my code to Bitbucket (without UI)


abrl91:
### `getOrganization` : 
- add a boolean property to GitOrganization Prisma schema, graphQL data model, and git-utils interface: `useGroupingForRepositories` that indicates if the provider supports repository grouping.
- The GitOrganization will include the property whether the provider is using grouping or not.
- each provider should return a true or false fixed (hard-coded) value.
- The value should also be saved in the DB, and returned to the client
---

### `getGitGroups` (bitbucket workspaces) : 
[list workspaces of the current user:](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-workspaces/#api-user-permissions-workspaces-get) 
`/2.0/user/permissions/workspaces`
- choosing the organization 
- click on create repository / choose a repository
- in case the gitOrganization uses grouping, we need to call 'getGitGroups' to get a list of groups, before we call 'getRepositories'
- dropdown for all workspaces => **this is a paginated API, so we may want to use something like the repositories list and not a dropdown**

---

### `getRepositories`
[list of repositories in a workspace](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-repositories-workspace-get)
`/2.0/repositories/{workspace}`
- after choosing a git group, the user can:
   - choose a repository from the selected workspace
   - create a repository in the selected workspace
   - to do that, we need to add gitGroupName property to the interface (GetRepositoriesArgs)
   - we can use the boolean property `useGroupingForRepositories` to know if we need to use this arg in the request - when false we don't need to send any value, and the git provider is ignoring the parameter.

---

### `createRepositroy`:
[create repository](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-repositories/#api-repositories-workspace-repo-slug-post)
`/2.0/repositories/{workspace}/{repo_slug}`
- create a Bitbucket repository in the selected workspace
- in order to do that, we need to add gitGroupName property to the interface
- we can use the boolean property `useGroupingForRepositories` to know if we need to use this arg in the request - when false we don't need to send any value, and the git provider is ignoring the parameter.

---

- [ ] #5440
### `createPullRequest`
[create pull request](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-pullrequests-post)
`/2.0/repositories/{workspace}/{repo_slug}/pullrequests`
[...]

PR #5389, opened 2023-03-09
Feat(git-utils): implements get git groups

abrl91:
part of: #5261 

## PR Details
- clean up unused DTOs related to the OAuth2 flow
- schema migration: add `useGroupingForRepositories` column for `GitOrganization` model. This property indicates if the provider uses grouped repositories or not.

- add an endpoint to get bitbucket workspaces => `gitGroups` (a generic name for future providers that use this term)

image

## PR Checklist


packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -423,6 +433,27 @@ export class GitProviderService {
     }
   }
 
+  async getGitGroups(args: GitGroupArgs): Promise<PaginatedGitGroup> {
+    const { gitProvider, oAuthUserName } = args.where;
+    const lowerCaseProvider = gitProvider.toLowerCase();
+    const gitClientService = await new GitClientService().create(
```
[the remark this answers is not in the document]
abrl91: For some reasons, GitHub didn't refresh this file. This method looks like that after the refactoring:

```ts
async getGitGroups(args: GitGroupArgs): Promise<PaginatedGitGroup> {
    const organization = await this.getGitOrganization({
      where: {
        id: args.where.organizationId,
      },
    });
    const gitClientService = await new GitClientService().create(
      {
        provider: EnumGitProvider[organization.provider],
        providerProperties: {
          ...JSON.parse(JSON.stringify(organization.providerProperties)),
          clientId: this.clientId,
          clientSecret: this.clientSecret,
        },
      },
      this.logger
    );

    return await gitClientService.getGitGroups();
  }
```

packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts
```diff
@@ -46,7 +46,7 @@ export class PullRequestService {
     const gitClientService = await new GitClientService().create(
       {
         provider: gitProvider,
-        installationId,
+        providerProperties: { installationId },
```
yuval-hazaz, line 49: For now, this is OK, but we need to think about how to use the git-util from services other than the amplication-server. where do we get the providerProperties from etc

packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -57,7 +60,12 @@ export class GitProviderService {
       limit: args.limit,
       page: args.page,
     };
-    const gitProviderArgs = { provider: args.gitProvider, installationId };
+    const gitProviderArgs = {
+      provider: args.gitProvider,
+      providerProperties: {
+        installationId,
+      },
+    };
```
yuval-hazaz, line 68: It will be better to have a single function to get the GitClientService so all functions in this files can use a single method to get the client

packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -358,71 +371,75 @@ export class GitProviderService {
...
+        id: args.where.organizationId,
+      },
+    });
+    const gitClientService = await new GitClientService().create(
+      {
+        provider: EnumGitProvider[organization.provider],
+        providerProperties: {
+          ...JSON.parse(JSON.stringify(organization.providerProperties)),
+          clientId: this.clientId,
+          clientSecret: this.clientSecret,
```
yuval-hazaz, line 434: I don't see why we need to send a special object instead of just passing the providerProperties as it is. 
What are clientId and clientSecret? Where do we get it from?

yuval-hazaz (changes requested):

packages/amplication-server/src/core/git/git.provider.service.ts
abrl91: These values come from environment variables. We talked about it. The git-utils is a library, and it shouldn't rely on env variables, so I moved the `process.env` part from the lib to amplication-server

packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts
abrl91: git-pull-request is also using it

packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -54,7 +54,7 @@ export class GitProviderService {
     gitProviderArgs: GitProviderArgs,
     logger = this.logger
   ): Promise<GitClientService> {
-    return this.createGitClient(gitProviderArgs, logger);
+    return new GitClientService().create(gitProviderArgs, logger);
```
abrl91, line 57: @yuval-hazaz we use this service in git-pull-request service, but only one. Should we do the same there?
https://github.com/amplication/amplication/blob/master/packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts#L46

packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -47,6 +50,13 @@ export class GitProviderService {
     );
   }
 
+  async createGitClient(
+    gitProviderArgs: GitProviderArgs,
+    logger = this.logger
+  ): Promise<GitClientService> {
+    return new GitClientService().create(gitProviderArgs, logger);
```
abrl91, line 57: @yuval-hazaz we use this service in git-pull-request service, but only one. Should we do the same there?
https://github.com/amplication/amplication/blob/master/packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts#L46

packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts
yuval-hazaz: I meant to think about how we get the properties from the DB and pass them to git-util

packages/amplication-server/src/core/git/git.provider.service.ts
yuval-hazaz: so we need to think on how to pass them. It should not be passed as part of the providerProperties since the client of the git-util should not  know about the structure and type of the param

packages/amplication-server/src/core/git/git.provider.service.ts
yuval-hazaz: we should talk separately about git-pull-request  - how would we fetch the org properties from the DB?

packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts
```diff
@@ -2,19 +2,58 @@ import {
...
+    );
+    const githubAppId = this.configService.get<string>(Env.GITHUB_APP_APP_ID);
+    const githubAppPrivateKey = this.configService.get<string>(
+      Env.GITHUB_APP_PRIVATE_KEY
+    );
+    const githubAppInstallationUrl = this.configService.get<string>(
+      Env.GITHUB_APP_INSTALLATION_URL
+    );
+
+    this.gitProvidersConfiguration = {
```
yuval-hazaz, line 43: use the type from git-util

yuval-hazaz (changes requested):

yuval-hazaz (approved):

packages/amplication-git-pull-request-service/src/pull-request/pull-request.service.ts
abrl91: I use it in line 16

Issue #4846, opened 2023-01-05
🐛 Bug Report: The commits page doesn't work for some projects


abrl91:
### What happened?

The commits page doesn't work for some projects. Getting the following error in the network
![image](https://user-images.githubusercontent.com/39680385/210732989-9d9cf86c-5549-43ec-b376-412461dd6f55.png)

The page presents the empty state
![image](https://user-images.githubusercontent.com/39680385/210733048-008e8991-b245-4705-b827-8d71bce4a788.png)

### What you expected to happen

I expect that as long as I have committed changes, I will see them on the commits page.

### How to reproduce

It happened for the 'blog-server' project. Hard to reproduce. It may related to very old projects (apps)

abrl91:
![Image](https://user-images.githubusercontent.com/39680385/213658270-fa036641-a45c-4bfb-89d2-0e047c0f2a51.png)

overbit:
https://app.amplication.com/ckyrgxlfa135491801s6vj2ygjn4/clbox7jiw00n8j301opsinsbx/commits

overbit:
I think the issue is related to commits for builds created by users no longer part of the workspace

```sql
SELECT "Build"."userId"
FROM "Commit" 
JOIN "Project" ON "Commit"."projectId" = "Project".id
JOIN "Workspace" ON "Project"."workspaceId" = "Workspace".id
join "Resource" on "Resource"."projectId" = "Project".id
join "Build" on "Build"."resourceId" = "Resource".id
LEFT JOIN "User" ON "User".id = "Build"."userId"
WHERE "Workspace".id = 'ckyrgxlfa135491801s6vj2ygjn4' 
 AND "Project".id = 'clbox7jiw00n8j301opsinsbx'
GROUP BY "Build"."userId";

SELECT * 
FROM "Account"
RIGHT JOIN "User" ON "Account".id = "User"."accountId"
where "User".id IN ('ckyrgxlfa135491901s6oxv6ihul', 'cl4qu74nj64110501l9zgpbsn41', 'clbowxxtb00oele01qo3rmp9r');
```

overbit:
The feeling got proven right. We now need to find out out to fix it!

abrl91:
sample project before:
![Screenshot 2023-02-27 at 17 56 14](https://user-images.githubusercontent.com/39680385/221614354-78def3f2-42c0-4c15-b5d9-4af3a55c50f3.png)


sample project after:
[...]

PR #5412, opened 2023-03-13
fix(server): commits resolver issue due to deleted users

overbit:
Close: #4846

## PR Details

Fix server logic to return user's commits (regardless its deleted status)

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/user/user.service.ts
+++ b/packages/amplication-server/src/core/user/user.service.ts
@@ -7,12 +7,15 @@ import { UserRoleArgs } from "./dto";
 export class UserService {
   constructor(private readonly prisma: PrismaService) {}
 
-  findUser(args: Prisma.UserFindUniqueArgs): Promise<User> {
+  findUser(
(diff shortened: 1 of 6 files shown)


packages/amplication-server/src/core/user/user.service.ts
```diff
@@ -7,12 +7,15 @@ import { UserRoleArgs } from "./dto";
 export class UserService {
   constructor(private readonly prisma: PrismaService) {}
 
-  findUser(args: Prisma.UserFindUniqueArgs): Promise<User> {
+  findUser(
+    args: Prisma.UserFindUniqueArgs,
+    includeDeleted = false
```
abrl91, line 12: Why did you set the default to false and then pass on each call to this function `true`?
Why didn't you just set the default to true? or just don't set a default?

overbit: The current behaviour is to not include deleted resources, so this behaviour won't break any current logic

abrl91 (approved):

Issue #5415, opened 2023-03-14
🐛 Bug Report: 


mihailShumilov:
### What happened?

At next brach if billing is disabled you can't create a PR
you got at error
```
[amplication-server] error 2023-03-14T10:55:26.146Z Failed to emit Create Pull Request Message. Error: TypeError: Cannot read properties of undefined (reading 'hasAccess') {}

```

### What you expected to happen

I expect that PR creation would work

### How to reproduce

Just checkout `next` brach and try to build with PR creation

### Amplication version

1.4.2

### Environment

node - v16.13.0
npm - 8.1.0
docker - 20.10.17, build 100c701
OS - Mac OS 13.2 (22D49)

### Are you willing to submit PR?

Yes I am willing to submit a PR!

GreenMachine01:
@barshimi can you assist?
I remember this is a known issue with a fix.

mihailShumilov:
PR already exists

PR #5416, opened 2023-03-14
fix/5415-PR-creation-issue

mihailShumilov:
Close: #5415

## PR Details

If Billing is disabled you can't create a PR

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -454,15 +454,18 @@ export class BuildService {
         try {
           await this.actionService.logInfo(step, PUSH_TO_GITHUB_STEP_START_LOG);
 
-          const smartGitSyncEntitlement =
-            await this.billingService.getBooleanEntitlement(
-              project.workspaceId,
-              BillingFeature.SmartGitSync
-            );
+          let pullRequestMode = EnumPullRequestMode.Accumulative;
+          if (this.billingService.isBillingEnabled) {
+            const smartGitSyncEntitlement =
+              await this.billingService.getBooleanEntitlement(
+                project.workspaceId,
+                BillingFeature.SmartGitSync
+              );
 
-          const pullRequestMode = smartGitSyncEntitlement.hasAccess
-            ? EnumPullRequestMode.Accumulative
-            : EnumPullRequestMode.Basic;
+            pullRequestMode = smartGitSyncEntitlement.hasAccess
+              ? EnumPullRequestMode.Accumulative
+              : EnumPullRequestMode.Basic;
+          }
 
           const createPullRequestArgs: SendPullRequestArgs = {
             gitOrganizationName: gitOrganization.name,
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -454,11 +454,12 @@ export class BuildService {
         try {
           await this.actionService.logInfo(step, PUSH_TO_GITHUB_STEP_START_LOG);
 
-          const smartGitSyncEntitlement = this.billingService.isBillingEnabled ? await this.billingService.getBooleanEntitlement(
+          const smartGitSyncEntitlement = this.billingService.isBillingEnabled
+            ? await this.billingService.getBooleanEntitlement(
                 project.workspaceId,
                 BillingFeature.SmartGitSync
-              ) : false
-          
+              )
+            : false;
 
           const createPullRequestArgs: SendPullRequestArgs = {
             gitOrganizationName: gitOrganization.name,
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -454,15 +454,18 @@ export class BuildService {
...
-          const smartGitSyncEntitlement =
-            await this.billingService.getBooleanEntitlement(
-              project.workspaceId,
-              BillingFeature.SmartGitSync
-            );
-
-          const pullRequestMode = smartGitSyncEntitlement.hasAccess
-            ? EnumPullRequestMode.Accumulative
-            : EnumPullRequestMode.Basic;
+          let pullRequestMode = EnumPullRequestMode.Accumulative;
```
tupe12334, line 457: I think the default suppose to be Basic, @EugeneTseitlin Please have a look here

mihailShumilov: usually billing is disabled for local env, so I think that we can use `Accumulative` mode
Or we can set that mode in `.env` variable

EugeneTseitlin: > I think the default suppose to be Basic, @EugeneTseitlin Please have a look here

Sure, it has to be basic by default.

EugeneTseitlin (changes requested):

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -454,15 +454,18 @@ export class BuildService {
...
+            const smartGitSyncEntitlement =
+              await this.billingService.getBooleanEntitlement(
+                project.workspaceId,
+                BillingFeature.SmartGitSync
+              );
+
+            pullRequestMode = smartGitSyncEntitlement.hasAccess
+              ? EnumPullRequestMode.Accumulative
+              : EnumPullRequestMode.Basic;
+          }
```
abrl91, line 468:
```suggestion
          const smartGitSyncEntitlement = this.billingService.isBillingEnabled ? await this.billingService.getBooleanEntitlement(
                project.workspaceId,
                BillingFeature.SmartGitSync
              ) : false
          
```

abrl91: and in line 490:
```ts
pullRequestMode: smartGitSyncEntitlement?.hasAccess 
             ? EnumPullRequestMode.Accumulative
              : EnumPullRequestMode.Basic;
```

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -484,7 +481,11 @@ export class BuildService {
               adminUIPath: resourceInfo.settings.adminUISettings.adminUIPath,
               serverPath: resourceInfo.settings.serverSettings.serverPath,
             },
-            pullRequestMode,
+            pullRequestMode:
```
abrl91, line 484: What is the value of `pullRequestMode` if `smartGitSyncEntitlement` returns `false` (meaning typeof boolean)?

abrl91: I think you can do something like that:
declare the pull request mode
```suggestion
          const pullRequestMode =
            (smartGitSyncEntitlement &&
              smartGitSyncEntitlement.hasAccess &&
              EnumPullRequestMode.Accumulative) ||
            EnumPullRequestMode.Basic; 
```
and then use it here

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -484,7 +481,11 @@ export class BuildService {
...
+              typeof smartGitSyncEntitlement !== "boolean" &&
+              smartGitSyncEntitlement?.hasAccess
+                ? EnumPullRequestMode.Accumulative
+                : EnumPullRequestMode.Basic,
```
abrl91, line 488: I think you can do something like that:
```suggestion
          pullRequestMode: (smartGitSyncEntitlement &&
              smartGitSyncEntitlement.hasAccess &&
              EnumPullRequestMode.Accumulative) ||
            EnumPullRequestMode.Basic; 
```
or declare it above in a const and use the constant here

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -484,7 +481,10 @@ export class BuildService {
               adminUIPath: resourceInfo.settings.adminUISettings.adminUIPath,
               serverPath: resourceInfo.settings.serverSettings.serverPath,
             },
-            pullRequestMode,
+          pullRequestMode: (smartGitSyncEntitlement &&
+              smartGitSyncEntitlement.hasAccess &&
+              EnumPullRequestMode.Accumulative) ||
+            EnumPullRequestMode.Basic;
```
tupe12334, line 487:
```suggestion
          pullRequestMode: (smartGitSyncEntitlement &&
              smartGitSyncEntitlement.hasAccess) ?
              EnumPullRequestMode.Accumulative :
            EnumPullRequestMode.Basic; 
```

yuval-hazaz (approved):

Issue #5453, opened 2023-03-17
🐛 Bug Report: resources in deleted projects are counted as active when the usage is checked on the free tier


yuval-hazaz:
### What happened?

When a user on the free tier commit changes, he will get the notification about the license even if the workspace have 3 or less services - because other projects that were deleted are still counted 

<img width="829" alt="image" src="https://user-images.githubusercontent.com/43705455/225826679-0604683a-f571-4274-9fab-0c24d200c7a3.png">

### What you expected to happen

deleted resources should not be counted 

### How to reproduce

create two projects in the same workspace 
first project with 2 services and the second project with 3 services.
try to commit the changes in one of the services and you will get the notification about the usage/ 

delete one of the projects, and try to commit the changes in the remaining project. 

you will get the notification again, although the user has only 2 services now. 

### Amplication version

1.4.2.

yuval-hazaz:
this is a blocker for #4044

overbit:
Bare in mind that the `calculateMeteredUsage` might also need some attention due to the project deletion 
https://github.com/amplication/amplication/blob/next/packages/amplication-server/src/core/project/project.service.ts#L156

abrl91:
passed QA ✅ 

https://user-images.githubusercontent.com/39680385/226357739-a663f049-e54a-4595-a172-f24b0187caff.mov

PR #5459, opened 2023-03-17
Fix/5453 bug report resources in deleted projects are counted as active when the usage is checked on the free tier

EugeneTseitlin:
Close: #5453 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/project/project.service.ts
+++ b/packages/amplication-server/src/core/project/project.service.ts
@@ -94,7 +94,7 @@ export class ProjectService {
     await this.billingService.reportUsage(
       project.workspace.id,
       BillingFeature.Services,
-      archivedServiceCount
+      -archivedServiceCount
     );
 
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -94,7 +94,7 @@ export class ProjectService {
     await this.billingService.reportUsage(
       project.workspace.id,
       BillingFeature.Services,
-      archivedServiceCount
+      -archivedServiceCount
```
abrl91, line 97: what is this `-` ?

abrl91: I understand, but it looks odd 🤕 
If I'm not mistaken, every time a user deletes a project we send the **total** `archivedServiceCount`. That means that we will subtract more than one time the same services.
Am I missing something?

abrl91 (approved):

PR #5461, opened 2023-03-17
Fix(server): bitbucket refresh token calculation

abrl91:
Part of: #5261 

## PR Details
fix the refresh token calculation

## PR Checklist

### DIFF
--- a/packages/amplication-git-utils/src/providers/bitbucket/bitbucket.service.ts
+++ b/packages/amplication-git-utils/src/providers/bitbucket/bitbucket.service.ts
@@ -89,7 +89,7 @@ export class BitBucketService implements GitProvider {
       refreshToken: authData.refresh_token,
       scopes: authData.scopes.split(" "),
       tokenType: authData.token_type,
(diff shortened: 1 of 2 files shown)


packages/amplication-git-utils/src/providers/bitbucket/bitbucket.service.ts
```diff
@@ -89,7 +89,7 @@ export class BitBucketService implements GitProvider {
       refreshToken: authData.refresh_token,
       scopes: authData.scopes.split(" "),
       tokenType: authData.token_type,
-      expiresAt: Date.now() + authData.expires_in,
+      expiresAt: Math.floor(Date.now() / 1000) + authData.expires_in, // 7200 seconds = 2 hours
```
overbit, line 92: If bitbucket returns expiresIn in seconds, it would make more sense the following:

```suggestion
      expiresAt: Date.now() + ( authData.expires_in * 1000 ), // 7200 seconds = 2 hours
```

overbit (changes requested):

Issue #5476, opened 2023-03-19
DX: Running local dev environment missing automation or scripts to run the different components


yuval-hazaz:
### Feature description

To run a local dev env, the developer needs to follow the instructions on the README and run each of the apps separately.
Also, it is not clear which apps are required or which app goes with which app.

This is the current section in the README.md
```
```bash
  # This will serve the Amplication Server in development mode
  npx nx serve amplication-server

  # This will serve the Amplication Client in development mode
  npx nx serve amplication-client
  ```
> ℹ️ **Please note that in order to be able to run the app's client properly, you need to `serve` both the server and client plus any additional component for specific functionalities:**
> - `npx nx serve amplication-client`
> - `npx nx serve amplication-server`
> 
> **To allow code generation**
> - `npx nx serve amplication-build-manager` 
> - `npx nx serve local-data-service-generator-controller`
> 
> **To allow git syncronisation with selected provider**
> - `npx nx serve amplication-git-pull-request-service` 
> - `npx nx serve local-data-service-generator-controller`<br> _check additional documentation in [local-data-service-generator-controller](https://github.com/amplication/amplication/blob/master/packages/local-data-service-generator-controller/README.md)_
> 
> **To allow source code visualisation**
> - `npx nx serve amplication-storage-gateway`
> 
> **To allow plugin settings configuration in client app**
> - `npx nx serve amplication-plugin-api`
>

```


### Use case

_No response_

### Are you willing to submit PR?

_No response_

overbit:
## Test result

`FAILURE` ❌ 

Following the instruction causes the following error since we try to start the same component in different commands:
<img width="930" alt="image" src="https://user-images.githubusercontent.com/2861984/226693254-cae831de-acc9-4186-a363-f04a7d8f572a.png">

PR #5477, opened 2023-03-19
docs: update setup instructions for local dev env

yuval-hazaz:
Close: #5476

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,13 @@
     "db:migrate:deploy": "nx run-many --target=db:migrate:deploy",
     "prepare": "husky install",
     "precommit": "lint-staged --relative",
-    "help": "nx help"
+    "help": "nx help",
+    "serve:server": "npx nx serve amplication-server",
+    "serve:client": "npx nx serve amplication-client",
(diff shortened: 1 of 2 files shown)


morhag90 (approved):

package.json
```diff
@@ -13,7 +13,13 @@
...
     "prepare": "husky install",
     "precommit": "lint-staged --relative",
-    "help": "nx help"
+    "help": "nx help",
+    "serve:server": "npx nx serve amplication-server",
+    "serve:client": "npx nx serve amplication-client",
+    "serve:dsg": "nx run-many --parallel --target=serve --projects=amplication-build-manager,local-data-service-generator-controller",
+    "serve:git": "nx run-many --parallel --target=serve --projects=amplication-git-pull-request-service,local-data-service-generator-controller",
+    "serve:storage": "npx nx serve amplication-storage-gateway",
+    "serve:plugins": "npx nx serve amplication-plugin-api"
```
overbit, line 22: I get the point of facilitating the user understanding but it is not a good practice mix packages scripts; since we are using nx monorepo we should use root commands in nx.json.

overbit:
Test result

`FAILURE` ❌ 

Following the instruction causes the following error since we try to start the same component in different commands:
image

Issue #5411, opened 2023-03-13
As Operator - I want to stop syncing free users accounts to Stripe


GreenMachine01:
**Overview:**
We are currently syncing free and trial subscriptions into Stripe when provisioning subscriptions with Stigg using the shouldSyncFree: true property.

Reasons why this is not recommended:
1. Easier to track and troubleshoot paying customers if you only sync paying users
2. Due to a Stripe limitation, offering discounts on "first purchase" doesn't work when a customer already has a free or trial subscription, since it's considered a purchase

**Requirements:**
1. Change the property 'shouldSyncFree' to 'false'

PR #5504, opened 2023-03-21
disable sync of free customers to stripe

EugeneTseitlin:
Close: #5411 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/billing/billing.service.ts
+++ b/packages/amplication-server/src/core/billing/billing.service.ts
@@ -264,6 +264,7 @@ export class BillingService {
       const stiggClient = await this.getStiggClient();
       await stiggClient.provisionCustomer({
         customerId: workspaceId,
+        shouldSyncFree: false,
         subscriptionParams: {
           planId: plan,
         },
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/billing/billing.service.ts
```diff
@@ -264,6 +264,7 @@ export class BillingService {
       const stiggClient = await this.getStiggClient();
       await stiggClient.provisionCustomer({
         customerId: workspaceId,
+        shouldSyncFree: false,
```
overbit, line 267: Looks good even if we don't have unit test here..

overbit (approved):

packages/amplication-server/src/core/billing/billing.service.ts
overbit, line 267: Missing unit test

overbit (changes requested):

packages/amplication-server/src/core/billing/billing.service.ts
EugeneTseitlin: It's a request to external service. It can't be tested by unit test.

overbit: it can, the amplication business logic is to not sync free account, so we can spy and check that we pass the correct arguments.

overbit (changes requested):

packages/amplication-server/src/core/billing/billing.service.ts
EugeneTseitlin: Currently we don't have any tests for BillingService. Mocking it's dependencies and writing basic tests will take time. I've opened a separate issue form that: https://github.com/amplication/amplication/issues/5539
Let's move on with this minor change and test in manually.

overbit (approved):

Issue #4774, opened 2023-01-01
As User - I want my workspace name and background color to reflect the plan activated on it


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/210178801-b831fe28-bcfe-4851-870f-d21b7c65b8b3.png)

**Design** - https://www.figma.com/file/REUaAG2f5EkgyghHMvACxd/Concept----Paywall?node-id=836%3A18964&t=oJdAJiitbsvucMw2-1

**Requirements:**
1. Plan text color needs to match the color of the plan
2. Workspace icon background color needs to match the color of the plan

### Plan 
Each Subscription type should be associated with corresponding css class
Set color for each of these classes  
Take colors from design that is linked to the issue

Plan is written by @EugeneTseitlin

GreenMachine01:
@EugeneTseitlin Check to do it via entitlement in Stigg

EugeneTseitlin:
@GreenMachine01 What exactly do you want to manage via Stigg? Obtain plan status  and decide about a color in client? Manage color via Stigg?

GreenMachine01:
@yuval-hazaz @EugeneTseitlin Please discuss how to implement this between you.
I don't have a preference to use Stigg for this or not. I think it is overkill to use Stigg for the colors and I would just do the logic on our side based on the plan of the workspace.

PR #5540, opened 2023-03-23
Feat/4774 as user i want my workspace name and background color to reflect the plan activated on it

EugeneTseitlin:
Close: #4774 

## PR Checklist


packages/amplication-client/src/Workspaces/WorkspaceSelector.tsx
```diff
@@ -61,7 +61,12 @@ function WorkspaceSelector() {
               <span className={`${CLASS_NAME}__current__name`}>
                 {currentWorkspace.name}
               </span>
-              <span className={`${CLASS_NAME}__current__plan`}>
+              <span
+                className={classNames(
+                  `${CLASS_NAME}__current__plan`,
+                  currentWorkspace.subscription?.subscriptionPlan?.toLocaleLowerCase()
+                )}
+              >
```
abrl91, line 69: it looks good, but I suggest adding a fallback

abrl91 (approved):

packages/amplication-client/src/Workspaces/WorkspaceSelector.tsx
EugeneTseitlin: what kind of? there is a default color for this text. In case subscription plan doesn't exist, additional class just won't show up.

Issue #5568, opened 2023-03-24
🐛 Bug Report: e2e ci fails to start


overbit:
### What happened?

https://github.com/amplication/amplication/actions/runs/4513146176/jobs/7947692891#step:7:75

### What you expected to happen

Run

### How to reproduce

-

### Are you willing to submit PR?

Yes I am willing to submit a PR!

overbit:
## Test result 
`PASSED` ✅ 
https://github.com/amplication/amplication/actions/runs/4518356022/jobs/7958178163

PR #5569, opened 2023-03-24
chore(e2e): fix issue due to cypress missing in CI cache

overbit:
Close: #5568 

## PR Details

Fix the issue reported, due to CI cache being generated without including cypress dependencies.
The solution is to fix the cypress bin into the cached node_modules folder https://docs.cypress.io/guides/references/advanced-installation#Binary-cache

## PR Checklist


.github/workflows/nx.template.yml
```diff
@@ -57,7 +58,6 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          ref: ${{ inputs.nx-head }}
```
levivannoort, line 60: Why is is this ref removed?

levivannoort (commented):
Other changes make sense but I am unsure about the checkout step - maybe it is redundant?

.github/workflows/nx.template.yml
overbit: It's to use the default configuration. I'll move this in a different PR to test it in isolation

levivannoort (approved):

Issue #5515, opened 2023-03-21
🐛 Bug Report: Override git provider settings doesn't work on a new workspace without prior Git connections


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/226653000-c8351fac-65ae-4d13-ae40-faf5af56377b.png)

**Steps to Reproduce:**
1. Create a new Workspace
2. Create a new project
3. Create a new service
4. go to the service git provider tab

**Expected:**
User can override the project settings and connect to his desired git provider and repo

**Actual:**
The override toggle doesn't change when clicked.

**Severity:**
 3 - Major

Jyotiraditya24:
hey can I take up this issue?

Jyotiraditya24:
Done 🚀

PR #5558, opened 2023-03-23
Fix(client) fix git toggle override

YAbargil:
Close: #5515

## PR Details

Changed the `isToggleDisable` 's condition

## PR Checklist

- [ ] Tests for the changes have been added
* [x] `npm test amplication-client` doesn't throw any error

PR #5572, opened 2023-03-25
fix(client): git toggle override

Jyotiraditya24:
the variable isToggleDisable was set to true , which should be false initially
Close: #5515

this solves the problem of -> The override toggle doesn't change when clicked.
now, User can override the project settings and connect to his desired git provider and repo

### DIFF
--- a/packages/amplication-client/src/Resource/git/ServiceConfigurationGitSettings.tsx
+++ b/packages/amplication-client/src/Resource/git/ServiceConfigurationGitSettings.tsx
@@ -99,7 +99,7 @@ const ServiceConfigurationGitSettings: React.FC<Props> = ({
     [resource.id, trackEvent, updateResourceOverrideStatus]
   );
 
-  const isToggleDisable = currentWorkspace?.gitOrganizations?.length === 0;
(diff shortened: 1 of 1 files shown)


packages/amplication-client/src/Resource/git/ServiceConfigurationGitSettings.tsx
```diff
@@ -99,7 +99,7 @@ const ServiceConfigurationGitSettings: React.FC<Props> = ({
     [resource.id, trackEvent, updateResourceOverrideStatus]
   );
 
-  const isToggleDisable = currentWorkspace?.gitOrganizations?.length === 0;
+  const isToggleDisable = currentWorkspace?.gitOrganizations?.length !== 0;
```
overbit, line 102: Thanks @Jyotiraditya24 for the PR. 
Can you please confirm that the original functionality still works?

If you have a git provider connected already at the project level, can you override it at the resource level?

overbit (changes requested):

overbit:
The issue did already have a PR to fix it. #5558

Issue #5607, opened 2023-03-28
🐛 Bug Report: Generated app-admin-ui: can't create an object that has related  field with "_"  in the name


morhag90:
### What happened?

You can't create an object with a related field with "_"  in the name. 

It happened because the admin-ui component was generated with the wrong source name.  

For example:

Schema:

model Employee {

  created                  DateTime                     @default(now())
  email                     String                       @unique
  id                           String                       @id @default(auto()) @map("_id") @mongo.ObjectId
  last_name            String
  name                    String
  **role_Id                  Role                         @relation(fields: [role_IdId], references: [id])
  role_IdId              String**      
  updated               DateTime                     @updatedAt
}

CreateEmployee component:

  <ReferenceInput **source="role.id"** reference="Role" label="role_id">

It should be:

  <ReferenceInput **source="role_Id.id"** reference="Role" label="role_id">

### What you expected to happen

Create employee successfully. 

### How to reproduce

Create an entity with a field name that contains "_," for example role_id 
Create another entity and add a related field to the field above. 

Try to create a new object to the first entity from the admin-UI. 

### Amplication version

1.4.3

PR #5608, opened 2023-03-28
fix(dsg): change source value for selectInput: Lookup type

morhag90:
Close: #5607

## PR Details

Fix source value for creating input filed type Lookup. 

## PR Checklist

Test case: 

Create an entity with a field name that contains "_," for example role_id
Create another entity and add a related field to the field above.

Try to create a new object to the first entity from the admin-UI.


overbit (approved):
Nice and easy fix. Well done! 💯

tupe12334 (approved):

EugeneTseitlin (approved):

Issue #4519, opened 2022-12-01
🐛 Bug Report:  - Build logs are not presented in the order of execution


GreenMachine01:
**We need estimation before fixing**

![Image](https://user-images.githubusercontent.com/112330500/205108117-56e2dc6a-17f3-4c20-8bef-918614233ce7.png)

**Steps to Reproduce:**
1. Click on build code
2. go to the Build log

**Expected:**
log events are ordered in the order they were executed

**Actual:**
log events are not ordered based on execution time
the timestamp is ordered but the events weren't processed in that order

PR #5617, opened 2023-03-29
feat(build-manager): emit build log events in the same partition to achieve log sorting by emission time  

overbit:
Close: #4519 

## PR Details

This PR introduce the KafkaProducer to `build-manager` to allow the emission of events that will be consumed in the publishing the order. 
Since in Kafka, order can only be guaranteed within a partition, events that we need to be consumed in order, need to be published on the same partition.
This is achieved by passing a key, as well as the values when publishing an event, that consists of the `buildId` in this case, to maintain the order of all the log events for the same build Id

## PR Checklist


tupe12334 (approved):

libs/util/nestjs/kafka/src/producer/KafkaProducer.module.ts
```diff
@@ -0,0 +1,31 @@
...
+  providers: [
+    {
+      provide: KAFKA_SERIALIZER,
+      useClass: KafkaMessageJsonSerializer,
+    },
+    KafkaProducerService,
+  ],
+  exports: [KAFKA_SERIALIZER, KafkaProducerService],
+})
+export class KafkaProducerModule {}
```
EugeneTseitlin, line 31: I think that this Module actually enable Consumer also. I suggest to make it KafkaModule.

EugeneTseitlin (changes requested):

libs/util/nestjs/kafka/src/producer/KafkaProducer.module.ts
overbit: It makes perfect sense to me. Updated

PR #5666, opened 2023-04-03
fix(server): add missing events

yuval-hazaz:
Close: #5668

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/git/git.provider.service.ts
+++ b/packages/amplication-server/src/core/git/git.provider.service.ts
@@ -35,6 +35,11 @@ import { EnumGitProvider } from "./dto/enums/EnumGitProvider";
 const GIT_REPOSITORY_EXIST =
   "Git Repository already connected to an other Resource";
 const INVALID_GIT_REPOSITORY_ID = "Git Repository does not exist";
+import {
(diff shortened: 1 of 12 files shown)


packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -329,7 +335,8 @@ export class GitProviderService {
 
   // installation id flow (GitHub ONLY!)
   async createGitOrganization(
-    args: CreateGitOrganizationArgs
+    args: CreateGitOrganizationArgs,
+    currentUser: User
```
EugeneTseitlin, line 339: There is no need to pass the whole user, only account id is actually used. Also, it won't be necessary to import User type to the service.

EugeneTseitlin (dismissed):

packages/amplication-server/src/core/git/git.provider.service.ts
yuval-hazaz: no gain in passing just the ID - I am always in favor of sending the entire object, instead of adding parameters one by one per use case (and saving the extra import is not a goal)

tupe12334 (approved):

PR #5689, opened 2023-04-04
feat(schema-registry): add schema registry library

overbit:
Close: https://github.com/amplication/amplication/issues/5694

## PR Details

As per BitBucket requirements, we need pass the repository group name for the repository target of a PR.
This PR extends this requirement, adding a new `schema-registry` library that will allow us to share Key and Value of a Kafka event between the services publishing and the ones consuming from the topic in question.
 
## PR Checklist

### DIFF
--- a/libs/schema-registry/README.md
+++ b/libs/schema-registry/README.md
@@ -1,6 +1,8 @@
 # schema-registry
 
-This library was generated with [Nx](https://nx.dev).
+This library acts as a simplified version of a schema registry for amplication internal events.
+
+It contains schemas, that consists of a `Key` and a `Value` class for easy consumption between different services.
 
 ## Building
 


libs/schema-registry/src/lib/create-pr-request/value.ts
```diff
@@ -1,9 +1,24 @@
...
-import { CreateCommit } from "./create-commit.dto";
 
-export class CreatePullRequestArgs {
+class Commit {
+  @IsString()
+  @IsOptional()
+  base?: string | undefined;
+  @IsOptional()
+  @IsString()
+  head?: string | undefined;
```
overbit, line 14: @tupe12334 / @abrl91 Why do we need `base` and `head`? `amplication-server` never produce an event passing these values so I guess it's safe to remove it.

tupe12334: @yuval-hazaz Wanted to leave that in case we want to do a future implementation; I agree with you but it's up to @yuval-hazaz

abrl91 (approved):

Issue #5440, opened 2023-03-15
As a User - I want to push my generated code to Bitbucket


abrl91:
part of: #5261

### getFile
- [x] #5505

`/2.0/repositories/{workspace}/{repo_slug}/src/{branchName}/{path}`

[get file or directory content](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-source/#api-repositories-workspace-repo-slug-src-commit-path-get)

---

### createPullRequestFromFiles
- [x] #5614

The simple PR strategy is not supported on Bitbucket

---

### createCommit
- [x] #5506

`/2.0/repositories/{workspace}/{repo_slug}/commit/{commit}/comments`

[create comment for a commit](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-commits/#api-repositories-workspace-repo-slug-commit-commit-comments-post)

--- 

### getPullRequest + createPullRequest
- [x] #5654

---

### Branch 

- [x] #5526

- getBranch
`/2.0/repositories/{workspace}/{repo_slug}/refs/branches/{name}`

[get a branch](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-refs/#api-repositories-workspace-repo-slug-refs-branches-name-get)

- createBranch
`/2.0/repositories/{workspace}/{repo_slug}/refs/branches`

[create a branch](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-refs/#api-repositories-workspace-repo-slug-refs-branches-post)

---

### getFirstCommitOnBranch
- [x] #5520

---

### getCurrentUserCommitList
- [x] #5547

refactor from using te API to using git client

---

### getCloneUrl + getToken 

- [x] #5527

---

### commentOnPullRequest
- [x] #5543

`POST /2.0/repositories/{workspace}/{repo_slug}/pullrequests/{pull_request_id}/comments`

[create a comment on PR](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-pullrequests/#api-repositories-workspace-repo-slug-pullrequests-pull-request-id-comments-post)

---

### Refactoring to leverage git cli instead of provider APIs
- [x] #5648

--- 

### Add support for repository group for providers that support this feature
- [ ] #5699 

--- 
### ⚠️  TODO ⚠️  

https://github.com/amplication/amplication/issues/5440#issuecomment-1497533085

--- 

### Things to test:
- amplicationignore

abrl91:
@overbit  FYI

[...]

Issue #5699, opened 2023-04-05
Add support for repository group for providers that support this feature


overbit:
### Feature description

We need to persist the Group name for repositories hosted on providers that support repository grouping.

As implemented as part of https://github.com/amplication/amplication/issues/5261, `useGroupingForRepositories`, returned by the `getOrganization` query, allow the consumer to know if or not, a group name needs to be passed when creating a new repository.

PR #5706, opened 2023-04-06
fix(git-utils): getfile logic and parameters 

overbit:
Close: #5440 

## PR Details

- Made groupName and git reference parameters of get file to make easier to understand logic
- default GetFile to collect file from HEAD of default branch if git reference is not passed

## PR Checklist


libs/schema-registry/src/lib/create-pr-request/value.ts
overbit, line 1: ⚠️  based on #5699. Merge that first

abrl91 (approved):

Issue #5707, opened 2023-04-06
Rework the bitbucket authentication flow, after the access token and refresh token are collected


overbit:
Rework the bitbucket authentication flow, after the access token and refresh token are collected. Right now, only the amplication-server is able to read this data, but other services need that too

_Originally posted by @overbit in https://github.com/amplication/amplication/issues/5440#issuecomment-1497533085_
            
Currently, the bitbucket flow won't work due to missing token along the way.

abrl91:
Fixed by https://github.com/amplication/amplication/pull/5708

PR #5708, opened 2023-04-06
feat(pull-request-service): fix auth for BitBucket flow

overbit:
Close: #5707

## PR Details

This PR fix the auth flow (`pull-request-service` needed the token to proceed with any PR related actions). 
It also brings a set of fixes around several issues related to the new BitBucket provider and its repository group feature.

## PR Checklist

### DIFF
--- a/packages/amplication-git-utils/src/types.ts
+++ b/packages/amplication-git-utils/src/types.ts
@@ -36,18 +36,7 @@ export interface GitProvidersConfiguration {
 
 export interface OAuthProviderOrganizationProperties
   extends OAuthData,
-    CurrentUser {
-  links: { avatar: LinksMetadata };
-  username: string;
-  useGroupingForRepositories: boolean;
-  uuid: string;
-  displayName: string;
-  accessToken: string;
-  refreshToken: string;
-  tokenType: string;
-  expiresAt: number;
-  scopes: string[];
-}
+    CurrentUser {}
 
 export const isOAuthProviderOrganizationProperties = (
   properties: unknown


packages/amplication-git-utils/src/types.ts
```diff
@@ -33,6 +33,26 @@ export interface GitProvidersConfiguration {
...
+export type BitBucketProviderOrganizationProperties =
+  OAuthProviderOrganizationProperties;
+
+export interface GitHubProviderOrganizationProperties {
+  installationId: string;
+}
+
 export interface GitProviderArgs {
   provider: EnumGitProvider;
   providerOrganizationProperties: any;
```
mulygottlieb, line 58: Shouldn't we connect both (/all) types of *OrganizationProperties so that we don't use _any_ as the type for _providerOrganizationProperties_ ?

overbit: I agree

packages/amplication-server/src/core/git/git.provider.service.ts
```diff
@@ -379,9 +383,16 @@ export class GitProviderService {
   async createGitOrganization(
     args: CreateGitOrganizationArgs
   ): Promise<GitOrganization> {
+    if (args.data.gitProvider !== EnumGitProvider.Github) {
```
mulygottlieb, line 386: Is this on purpose not allowing BitBucket yet?

overbit: We should refactor this service method since it is prone to mistake/misunderstanding. 
The reason is that we use a different flow for bitbucket (OAuth authorisation grant flow) from github (that is using a custom Github App flow)

packages/amplication-git-utils/src/providers/bitbucket/errors.ts
```diff
@@ -0,0 +1,13 @@
...
+    Object.assign(this, { ...options });
+    Object.setPrototypeOf(this, BitbucketGenericError.prototype);
+  }
+}
+export class BitbucketNotFoundError extends BitbucketGenericError {
+  constructor() {
+    super("404 Not found");
+    Object.setPrototypeOf(this, BitbucketNotFoundError.prototype);
+  }
+}
```
abrl91, line 13: Nice!
we also have this one: https://github.com/amplication/amplication/blob/master/packages/amplication-git-utils/src/utils/custom-error.ts
so, maybe we can delete it and use what you did, just add another class/es for the other errors

packages/amplication-git-utils/src/types.ts
```diff
@@ -33,9 +34,66 @@ export interface GitProvidersConfiguration {
...
+  username: string;
+  useGroupingForRepositories: boolean;
+  uuid: string;
+  displayName: string;
+  accessToken: string;
+  refreshToken: string;
+  tokenType: string;
+  expiresAt: number;
+  scopes: string[];
+}
```
abrl91, line 50: what am I missing?
OAuthData includes these properties
accessToken: string;
  refreshToken: string;
  tokenType: string;
  expiresAt: number;
  scopes: string[];
  
  and CurrentUser includes these properties
  links: { avatar: LinksMetadata };
  username: string;
  useGroupingForRepositories: boolean;
  uuid: string;
  displayName: string;
  
  so what are we extending here?

overbit: done! Good spot!

Issue #5635, opened 2023-03-30
Printing application successfully generated in the build log


overbit:
### Feature description

Currently, we don't print any final message in case of a successful build as shown here.

<img width="1173" alt="image" src="https://user-images.githubusercontent.com/2861984/228856373-3c9789f0-8d69-4b2d-a75b-e2f85020c89e.png">

I'd love to see a final message in case of success stating that the app generation process finished successfully

Aman1919:
I want to work on this

souravjain540:
assigned to you @Aman1919

ArslanYM:
@souravjain540 If there is no movement on this issue , Can I work on this?

PR #5777, opened 2023-04-12
feat(build):Printing application successfully generated in the build log

Aman1919:
Issue : https://github.com/amplication/amplication/issues/5635


overbit (dismissed):
@Aman1919 well done! Thanks for your contribution

Aman1919:
> @Aman1919 well done! Thanks for your contribution

Thanks  : )

Aman1919:
If everthing is all right can u merge it @overbit

packages/data-service-generator/src/create-data-service.ts
```diff
@@ -47,6 +47,8 @@ export async function createDataService(
       durationInMs: endTime - startTime,
     });
 
+    await context.logger.info("App generation process finished successfully");
```
barshimi, line 50: you can use `logger.info` instead of context and you don't need sync method for that

mulygottlieb (approved):

Issue #5789, opened 2023-04-13
Rename `getAccessToken` and `getToken` in the `git-provider.interface`


abrl91:
I the git-utils library we have the git-provider interface. This interface includes `getAccessToken` and `getToken` methods, which are too confusing and too similar.

We need to rename them merge them into one method and get the data we need in each case.

abrl91:
Fixed by https://github.com/amplication/amplication/pull/5790

PR #5790, opened 2023-04-13
Refactor/5789 renaming get access token

abrl91:
Close: #5789

## PR Details
- remove getToken from the interface and use it in the service as a private method
- rename getAccessToken

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/git/git.provider.service.ts
+++ b/packages/amplication-server/src/core/git/git.provider.service.ts
@@ -543,13 +543,13 @@ export class GitProviderService {
 
       const gitClientService = await this.createGitClient(newGitProviderArgs);
       this.logger.info("Token is going to be expired, refreshing...");
-      const newOAuthData = await gitClientService.refreshAccessToken(
+      const newOAuthTokens = await gitClientService.refreshAccessToken(
         providerProperties["refreshToken"]
       );
 
       const newProviderProperties = {
         ...(providerProperties as object),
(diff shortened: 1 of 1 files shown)


packages/amplication-git-utils/src/git-provider.interface.ts
```diff
@@ -30,7 +30,7 @@ export interface GitProvider {
   init(): Promise<void>;
   getGitInstallationUrl(amplicationWorkspaceId: string): Promise<string>;
   getCurrentOAuthUser(accessToken: string): Promise<CurrentUser>;
-  getAccessToken(authorizationCode: string): Promise<OAuthData>;
+  getOAuth2FlowData(authorizationCode: string): Promise<OAuthData>;
```
overbit, line 33: what do you think of the following?
```suggestion
  getOAuthTokens(authorizationCode: string): Promise<OAuthTokens>;
```

packages/amplication-git-utils/src/providers/bitbucket/bitbucket.service.ts
```diff
@@ -526,7 +527,7 @@ export class BitBucketService implements GitProvider {
     );
   }
 
-  async getToken(): Promise<string> {
+  private async getToken(): Promise<string> {
```
overbit, line 530: Is this needed at all? the accessToken should be a property of the service already without refreshing it every time

Issue #5718, opened 2023-04-09
🐛 Bug Report: Auth module icon shows up in a delay in the onboarding wizard


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/230772663-bb841205-34e1-40f2-9abc-f6c97f4b4ef5.png)

**Expected:**
The Auth core module icon is shown immediately when the step is opened

**Actual:**
The Auth core module icon is shown after a few seconds of delay

GreenMachine01:
I noticed other icons are also shown in a delay

abrl91:
It still has a delay. Not only for the Auth Core module


https://github.com/amplication/amplication/assets/39680385/a52e4730-f1ef-493a-aa44-dc7124e1ab7c

abrl91:
screenshot from the screen recording

![Image](https://github.com/amplication/amplication/assets/39680385/20e83511-4b77-423d-90df-a224faafa1b6)



![Image](https://github.com/amplication/amplication/assets/39680385/3dcf0a4c-d133-4c0b-b158-d328fd46e893)

PR #5841, opened 2023-04-20
fix(client): slow image loading

morhag90:
Close: #5718 

## PR Details

Fix slow image loading. 

## PR Checklist

Test case:

Go to service wizard => navigate between steps and see that the images loading is o.k .

### DIFF
--- a/packages/amplication-client/src/Resource/create-resource/wizard-pages/LabelDescriptionSelector.tsx
+++ b/packages/amplication-client/src/Resource/create-resource/wizard-pages/LabelDescriptionSelector.tsx
@@ -25,7 +25,7 @@ export type LabelDescriptionSelectorProps = {
   description: string;
   subDescription?: string;
   icon?: string;
-  image?: string;
+  image?: string | JSX.Element;
   imageSize?: string;
   customClassName?: string;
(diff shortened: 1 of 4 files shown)


packages/amplication-client/src/Resource/create-resource/wizard-pages/LabelDescriptionSelector.tsx
```diff
@@ -65,15 +65,7 @@ export const LabelDescriptionSelector: React.FC<
...
-      {image && (
-        <CreateServiceCircleBadge>
-          <img
-            className={classNames(`${className}__logo`, imageSize)}
-            src={image}
-            alt={""}
-          />
-        </CreateServiceCircleBadge>
-      )}
+      {image && <CreateServiceCircleBadge>{image}</CreateServiceCircleBadge>}
```
barshimi, line 68: if you want to preserve the old behaviour you need to keep it in case image is string

morhag90: We don't need it anymore, I fixed all places using this @barshimi

Issue #5791, opened 2023-04-13
🐛 Bug Report: Unable to build service generated with MongoDB plugin


overbit:
### What happened?

Creating a service with mongodb plugin will generate a server application that fails to build due to malformed file(s) as shown in the app here:
https://github.com/overbit-work/microservices/blob/amplication/apps/service-b/src/health/base/health.service.base.ts

### What you expected to happen

The generated app is buildable and runnable without issues. 

### How to reproduce

1. create a new service
2. select the latest mongo db plugin
3. build the service
4. observe the generated code / try to build and run the generated server.

Alternatively, this is the input.json describing the build
[input.json.txt](https://github.com/amplication/amplication/files/11230315/input.json.txt)

### Amplication version

v1.5.0

mulygottlieb:
This does not reproduce for me at all on production with two separate attempts.

Issue #5837, opened 2023-04-20
Refactor `data-service-generator` misleading types


overbit:
Currently some DSG types shared through `code-gen-types` are not matching their use and they can cause broken executions and unknown results of the generatation.
This issues summarise the possible types that need to be taken care of before any private or community plugin since their refactoring will cause a breaking change.

- [x] `Module[]` : A Map,  or simply an Object since, under the hood, the JavaScript Object is implemented with a hash table. This array implies that more that the same file (file path) can have several different content, and at this time, we will need to make a decision on what which one will be the one to win
   - Impact: HIGH
   - Affects: `data-service-generation`, `code-gen-types`, `pull-request-service`, all current plugins (simple refactoring)
   - #5845 

- [x] Update the docs with the new events methods signatures https://github.com/amplication/docs/pull/290

overbit:
Related to the bug #5791

overbit:
Created new issues for the following tasks

- [ ] #5887
 - Impact: MEDIUM
 - Affect: data-service-generation, code-gen-types, plugins that leverage EventNames.CreateServerDotEnv
 
- [ ] #5888
 - Impact: LOW
 - Affect: data-service-generation, code-gen-types, plugins that leverage EventNames.CreateServerDockerCompose or EventNames.CreateServerDockerComposeDB

- [ ] #5889
 - Impact: LOW
 - Affect: data-service-generation, code-gen-types, plugins that leverage related events

PR #5845, opened 2023-04-21
fix(dsg): [BREAKING CHANGE] duplicated modules 

overbit:
Close: #5791 #5837
## PR Details

In order to fix the duplicated modules in the modules lists, this PR brings a breaking change to the way we interact with file generation by updating the type to a map type
This makes it clear to the developer if a file is going to be overwritten or not and it will allow dsg to always write one file only one time during a single run.

⚠️ This PR contains breaking changes and it will need to go live with https://github.com/amplication/plugins/pull/113 

**As part of this PR, I've published `@amplication/code-gen-types@2.0.0` with `beta` tag. Once released we need to just retag this version with `latest` tag.**

## PR Checklist


packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -129,6 +124,77 @@ export type Module = {
...
+  async replaceModulesCode(fn: (code: string) => string): Promise<void> {
+    for await (const module of Object.values(this.map)) {
+      module.code = fn(module.code);
+      this[module.path] = module;
+    }
+  }
+
+  /**
+   * Returns all modules in the map
+   * @returns An array of modules
```
abrl91, line 191:
```suggestion
   * @returns An array of modules
```

packages/data-service-generator/src/prepare-context.ts
```diff
@@ -26,7 +25,7 @@ export async function prepareContext(
   dSGResourceData: DSGResourceData,
   logger: ILogger,
   pluginInstallationPath?: string
-): Promise<Module[]> {
+): Promise<void> {
```
abrl91, line 28: Why did you change the type?

packages/data-service-generator/src/prepare-context.ts
```diff
@@ -68,7 +67,6 @@ export async function prepareContext(
   );
 
   context.plugins = plugins;
-  return [];
```
abrl91, line 71: why did you remove that?

packages/data-service-generator/src/prepare-context.ts
overbit: To reflect the actual behaviour. It never returns Module[]

packages/data-service-generator/src/prepare-context.ts
overbit: It is never called with the intent of creating a Module array

packages/amplication-code-gen-types/src/code-gen-types.ts
```diff
@@ -129,6 +124,76 @@ export type Module = {
...
+      await this.merge(map);
+    }
+  }
+
+  async set(path: string, module: Module) {
+    if (this.map[path]) {
+      await this.logger.warn(`Module ${path} already exists. Overwriting...`);
+    }
+    this.map[path] = module;
+  }
```
overbit, line 151: Auto review for @overbit. How about changing the signature of set to accept only `Module`?

```suggestion
  async set(module: Module) {
    if (this.map[module.path]) {
      await this.logger.warn(`Module ${module.path} already exists. Overwriting...`);
    }
    this.map[module.path] = module;
  }
```

mulygottlieb:
@overbit @barshimi please complete the review and agree on the changes still needed to merge this. I want this done ASAP.

packages/amplication-code-gen-types/src/code-gen-types.ts
yuval-hazaz: @overbit I Agree. let's use only module

yuval-hazaz (changes requested):

Issue #5798, opened 2023-04-14
🐛 Bug Report: createdAt and updatedAt fields are included in the input DTO when not using the default field name


yuval-hazaz:
### What happened?

When using a custom name for the `createdAt` or `updatedAt` fields, the generated code includes the fields in CreateInput and updateInput DTOs

<img width="723" alt="image" src="https://user-images.githubusercontent.com/43705455/232112051-3b0e5116-7c6f-43e7-80bb-b876334c3348.png">

<img width="1163" alt="image" src="https://user-images.githubusercontent.com/43705455/232112690-b8db2783-fe1e-48da-b8ba-232330ba4424.png">

### What you expected to happen

the fields should not be included even when a custom name is used

Instead of using the field name to filter out the property, we should use the special dataTypes 

https://github.com/amplication/amplication/blob/60c750e64de4a783a680e0cd218b26ece1b306db/packages/data-service-generator/src/utils/field.ts#L8

<img width="776" alt="image" src="https://user-images.githubusercontent.com/43705455/232112969-95c5564d-7a4a-43cb-a387-0bf4104edebb.png">

<img width="776" alt="image" src="https://user-images.githubusercontent.com/43705455/232113015-7fc915ac-86a3-46f5-9e9b-9c7a5d680fd9.png">

### How to reproduce

change the name of the field from createdAt to created 

### Amplication version

1.5.0

morhag90:
Passed QA:

<img width="777" alt="image" src="https://user-images.githubusercontent.com/97830649/234248968-6568fab1-9eec-4b48-b822-cba7e581695a.png">



<img width="712" alt="image" src="https://user-images.githubusercontent.com/97830649/234248891-ad3a70b9-4bdb-4430-8f65-7d0688ef7e64.png">

PR #5848, opened 2023-04-23
fix(dsg): ignore specific DTO field types

tupe12334:
Close: #5798 

## PR Details

This pull request fixes the problem that the ignore mechanism worked on the bases of name and not the bases of type.

To prevent future bugs with that, I changed also one of the fields of the `createAt` to `createIn` in the mock data.

## PR Checklist


mulygottlieb:
Looks very good @tupe12334.
Should we add a new test that proves the change works (that those fields are ignored/removed from CreateInput and Update Input DTOs when used with a custom name)?

mulygottlieb (approved):

Issue #4223, opened 2022-10-21
🐛 Bug Report: Autogenerated @isJson Validator for created Json-fields does not allow valid input in Request body


vincenzodomina:
### What happened?

Autogenerated @isJson Validator from class-validator for created Json-fields does not allow valid objects in request body and throws error:

property must be a string!

Autogenerated DTO entry:

```
  @ApiProperty({
    required: false,
  })
  @IsJson()
  @IsOptional()
  @Field(() => GraphQLJSON, {
    nullable: true,
  })
  reference_urls?: JsonValue;
```

My workaround with correct behaviour: remove the @IsJson() decorator, but i added some other validation to have some custom validation:

```
  @ApiProperty({
    required: false,
  })
  @IsArray()
  // "each" tells class-validator to run the validation on each item of the array
  @IsString({ each: true })
  @IsOptional()
  @Field(() => GraphQLJSON, {
    nullable: true,
  })
  reference_urls?: JsonValue;
```

### What you expected to happen

When adding the following input to a request body, it should not throw an error and pass it to prisma:

```
{
...
  "reference_urls": [
    "https://www.web.com/",
    "https://www.test.com/"
  ],
...
}
```

Since I want to store arrays of strings I have to use Json fields which is great concerning the postgres and prisma related parts, but the autogenerated @isJson validator prevents to work with it

So I think the solution would be to remove the autogenerated   @IsJson() decorator and maybe even add a feature to add more specific validation (if it is array, object or string, or custom DTO, etc..)

### How to reproduce

-Create an Entity with a Json field
- pass an array of string in the Create or Update route request body property

### Amplication version

1.0.2

souravjain540:
@tupe12334 can you help?

yuval-hazaz:
@arielweinberger @tupe12334 please review

tupe12334:
Hi @vincenzodomina I tried to reproduce the problem and was not able to, I can enter an array of strings into JSON field; can you provide more details, please?
What is the error message?

amirping:
having the same issue after the change of the role field in the user entity, sending an array of string like ['admin'], causing a bad request exception.

yuval-hazaz:
@amirping can you share a link to a repo with the generated code so we can reproduce the issue?

vincenzodomina:
> 

Sure, here is the error:
![image](https://user-images.githubusercontent.com/54762917/202130696-c03f27d7-e46c-4397-b27f-8f4136554ed3.png)

[...]

PR #5876, opened 2023-04-26
fix(dsg): replace @IsJson validator with custom validator for Json fields

overbit:
Close: #4223

## PR Details

This PR replaces the default `class-validator` `@IsJSON()` validator with a new custom validator, `@IsJSONValue()` that will be generated alongside the rest of the application code (including unit tests for the new custom validator).
This new validator will be used to validate any `Json` fields. 

The previous use of `@IsJSON()` was erroneous since it will succeed only when validating `string` containing a Json-like set of characters and not objects or arrays that, are valid JSON objects https://datatracker.ietf.org/doc/html/rfc7159#section-3   

This new `@IsJSONValue()` leverage `JSON.stringify` and `validator.js`.
 
## PR Checklist


levivannoort (approved):
LGTM.

PR #5899, opened 2023-05-03
🐛 Bug Report: Service Overview shows bad information

gitstart:
#### DEMO VIDEO

https://www.loom.com/share/02d7b93343ef4f7db7fb9e2de75ede32

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.


abrl91:
@gitstart Thank you. We'll review your PR soon

abrl91:
create branch from fork

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -210,7 +210,7 @@ export class ResourceService {
 
     if (
       isOnBoarding ||
-      (!gitRepositoryToCreate.isOverrideGitRepository &&
+      (!gitRepositoryToCreate?.isOverrideGitRepository &&
```
abrl91, line 213: Can you please explain how this change is related to this PR

packages/amplication-client/src/models.ts
```diff
@@ -592,6 +592,12 @@ export enum EnumAuthProviderType {
   Jwt = 'Jwt'
 }
 
+export enum EnumDbType {
+  mongo = 'mongo',
+  postgres = 'postgres',
+  mysql = 'mysql'
+}
+
```
abrl91, line 600: Did you add it manually?

packages/amplication-client/src/models.ts
```diff
@@ -1891,6 +1897,7 @@ export type ServiceSettings = IBlock & {
   serverSettings: ServerSettings;
   updatedAt: Scalars['DateTime'];
   versionNumber: Scalars['Float'];
+  dbType?: Scalars['String'];
```
abrl91, line 1900: Did you add it manually?

packages/amplication-client/src/Resource/constants.ts
```diff
@@ -139,7 +139,10 @@ export function prepareServiceObject(
           generateRestApi: generateRestApi,
           serverPath: serverDir,
         },
-        authProvider: EnumAuthProviderType.Jwt,
+        authProvider:
+          auth === "core"
+            ? EnumAuthProviderType.Jwt
+            : EnumAuthProviderType.Http,
```
abrl91, line 145: @morhag90 can you look at this please?

packages/amplication-client/src/models.ts
olalekanteeblaze: Yes

packages/amplication-client/src/models.ts
abrl91: Our models are generated automatically. If you want this enum only in the client, please add it where you need in another file.
If you also need it for the server, please have a look at EnumGitProvider for example (in amplication-server).
After adding it to the server, run `npx nx build amplication-server` and you will see the changes in all models

packages/amplication-client/src/models.ts
abrl91: Our models are generated automatically. I noticed that you added it to the DTO on the server. Remove it from here and run `npx nx build amplication-server`. You will see how the models are being changed automatically

packages/amplication-client/src/Resource/constants.ts
morhag90: @abrl91 Looks fine for now (in the future, if we add another option besides the "core," we will need to update this logic)

packages/amplication-client/src/Resource/OverviewTile/OverviewTile.tsx
```diff
@@ -14,6 +14,18 @@ const AuthProviderLabels: { [k in EnumAuthProviderType]: string } = {
...
+  mongo = "mongo",
+  postgres = "postgres",
+  mysql = "mysql",
+}
+
+const DbTypeLabels: { [k in EnumDbType]: string } = {
+  [EnumDbType.mongo]: "MongoDB",
+  [EnumDbType.mysql]: "MySQL",
+  [EnumDbType.postgres]: "PostgresSQL",
+};
```
abrl91, line 27: I would add it to the server in `packages/amplication-server/src/enums/EnumDbType`
```ts
export enum EnumDbType {
  MongoDB = 'MongoDB',
  MySQL = 'MySQL',
  PostgresSQL = 'PostgresSQL'
}

registerEnumType(EnumDbType, {
  name: "EnumDbType",
});
```

Then run `npx nx build amplication-server`

packages/amplication-client/src/Resource/OverviewTile/OverviewTile.tsx
```diff
@@ -25,6 +37,11 @@ const OverviewTile: React.FC<Props> = ({ resourceId }: Props) => {
     },
   });
 
+  const dbType = DbTypeLabels[data?.serviceSettings.dbType];
```
abrl91, line 40: here, you will be able to use the `EnumDbType` from models

abrl91 (changes requested):

packages/amplication-client/src/Resource/OverviewTile/OverviewTile.tsx
yuval-hazaz: by design the server should not be aware of the type of plugin used for the DB. I don't think we need to create this enum. 
I think it makes more sense to use fixed values in the client based on the installed plugin and in the future we will be able to add support for plugin categories to be able to identify the DB plugins more easily.


so my suggestion - get a list of installed plugins (use the hook usePlugin) and find one of the existing DB plugins and show the name of the one with the highest "order"

packages/amplication-server/src/core/serviceSettings/serviceSettings.service.ts
```diff
@@ -163,6 +163,8 @@ export class ServiceSettingsService {
             },
           },
           ...settings,
+          authProvider:
+            serviceSettings.authProvider || EnumAuthProviderType.Jwt,
```
abrl91, line 167: @morhag90 @gitstart 
When I choose: skip authentication, I see: Authentication: HTTP
I shouldn't see anything

image

olalekanteeblaze: @abrl91 authProvider is not nullable, so it's either HTTP or JWT. Should we make it nullable or JWT by default?. Also, there's is no option to select HTTP on the service creation wizard currently

abrl91: I understand, but it doesn't make sense to see the authentication section when you choose not to use it. 
@GreenMachine01 @mulygottlieb

GreenMachine01: @abrl91 you are correct. for a service without Auth the "Authentication" should indicate properly the situation.
I think instead of leaving it empty it will be better to show a text like "Not in Use" or "Disabled"

abrl91:
@olalekanteeblaze @gitstart I still see the authentication section even though I chose not to use authentication

olalekanteeblaze:
> @olalekanteeblaze @gitstart I still see the authentication section even though I chose not to use authentication

@abrl91 authentication is currently set to "Disabled"  when not set as @GreenMachine01 suggested. Should we just remove it in that case instead?

GreenMachine01:
@olalekanteeblaze can I see a screenshot?

olalekanteeblaze:
@GreenMachine01 
Screenshot 2023-06-01 at 12 16 50

GreenMachine01:
@olalekanteeblaze @abrl91 looks good to me.

abrl91 (approved):

Issue #5799, opened 2023-04-14
🐛 Bug Report: Duplicate function name when using a self-relation property with the same name


yuval-hazaz:
### What happened?

The resolver class is generated with a duplicate function name when using a self-relation property with the same name of the entity in plural 

e.g. 
entity: employee
field: employees

the resovler includes two function called employees - one for findMany and another one for the relation 

<img width="770" alt="image" src="https://user-images.githubusercontent.com/43705455/232115103-e5514626-0c4a-417c-b0d7-694d6ffbec0b.png">

<img width="770" alt="image" src="https://user-images.githubusercontent.com/43705455/232115134-93ee645f-c526-48a7-8a4b-77233d26b378.png">

### What you expected to happen

avoid duplicate name

### How to reproduce

entity: employee
field: employees

### Amplication version

1.5.0

overbit:
TEST PASSED 


![Image](https://user-images.githubusercontent.com/2861984/237050832-32e2704f-9de6-48d2-8f6a-5e7ad10de1cf.png)

PR #5910, opened 2023-05-07
fix(dsg):resolve duplicate function name on self relations

morhag90:
Close: #5799 

## PR Details

Resolved duplicated function names on self relations fields. 

## PR Checklist

Test cases:

Create an entity:  "Employee" with the following fields: 
1. employees
2. employee 

Run the server successfully.


packages/data-service-generator/src/server/resource/resolver/to-one.template.ts
```diff
@@ -10,11 +10,15 @@ declare interface SERVICE {
 
 declare const ENTITY_NAME: string;
 declare const RELATED_ENTITY_NAME: string;
```
mulygottlieb, line 12: @morhag90 let's just remove the unused RELATEDE_ENTITY_NAME like you did in the to-many.template.ts

mulygottlieb (approved):

Issue #5852, opened 2023-04-23
As User - I want creating a service without Auth to not create the 'User' entity


GreenMachine01:

mulygottlieb:
QA Passed:


![Image](https://github.com/amplication/amplication/assets/1912933/d4bde1c6-a934-41af-80d9-a4a8c6eb0a76)

PR #5996, opened 2023-05-14
feat(server/client): requireAuthenticationEntity flag to create user entity 

morhag90:
Close: #5852 

## PR Details

Add a flag to create a user entity when creating a new service. 

## PR Checklist

Test case:

Create new services: 

1. with no template and no auth module => no entities. 
2. with no template and with auth module => User entity. 
3. with template and no auth module => entities from the template with no user entity. 
4. with template and with auth module => entities from the template and user entity.

### DIFF
--- a/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
+++ b/packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
@@ -161,25 +161,6 @@ const PLUGIN_LOGO_BASE_URL =
   "https://raw.githubusercontent.com/amplication/plugin-catalog/master/assets/icons/";
 const CREATE_SERVICE_PATTERN = CREATE_SERVICE_STEPS.map((step) => step.index);
 
-const AUTH_PLUGINS = [
-  {
-    displayName: "Auth-core",
-    pluginId: "auth-core",
-    enabled: true,
-    npm: "@amplication/plugin-auth-core",
-    version: "latest",
-    resource: { connect: { id: "" } },
-  },
-  {
-    displayName: "Auth-jwt",
-    pluginId: "auth-jwt",
-    enabled: true,
-    npm: "@amplication/plugin-auth-jwt",
-    version: "latest",
-    resource: { connect: { id: "" } },
-  },
-];
-
 const signupCookie = getCookie("signup");
 
 const CreateServiceWizard: React.FC<Props> = ({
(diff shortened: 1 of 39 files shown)


packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -313,6 +319,8 @@ export class ResourceService {
       throw new ReservedEntityNameError(USER_ENTITY_NAME);
     }
 
+    console.log({ data });
```
mulygottlieb, line 322: Please remove console.log statement

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
```diff
@@ -402,6 +402,7 @@ const CreateServiceWizard: React.FC<Props> = ({
             isOverrideGitRepository: isOverrideGitRepository,
           };
         }
+        const requireAuthenticationEntity = authType === "core";
```
mulygottlieb, line 405: Is this true if and only if the user chooses to create the service with auth? (just checking with you since I don't know the wizard code well)

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
```diff
@@ -18,6 +18,11 @@ export class ResourceCreateInput {
   })
   description!: string;
 
+  @Field(() => Boolean, {
+    nullable: true,
+  })
+  requireAuthenticationEntity?: boolean;
+
```
abrl91, line 25: If this is optional, don't we want a default value? What will happen if we don't send anything?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -275,7 +280,8 @@ export class ResourceService {
       data: { ...USER_RESOURCE_ROLE, resourceId: resource.id },
     });
 
-    await this.entityService.createDefaultEntities(resource.id, user);
+    requireAuthenticationEntity &&
+      (await this.entityService.createDefaultEntities(resource.id, user));
```
abrl91, line 284: What will happen if requireAuthenticationEntity is false or falsy (undefined)?

packages/amplication-client/src/Resource/create-resource/CreateServiceWizard.tsx
morhag90: @mulygottlieb  yes

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @abrl91 add default value = true

abrl91 (approved):

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -257,9 +257,11 @@ export class ResourceService {
   async createService(
     args: CreateOneResourceArgs,
     user: User,
-    wizardType: string = null
+    wizardType: string = null,
```
barshimi, line 260: it's many args, I would think to convert it to object instead of parameters

morhag90: For 2 optional params???? @barshimi

PR #5975, opened 2023-05-12
feat(client): add support for static website deployment

overbit:
Fixes : https://github.com/amplication/amplication-infrastructure-next/issues/50

## PR Details

Replace the injectEnvVariables.js with nx build configurations to build the client application with the correct variables per environment.
The injectEnvVariables related scripts, nx target `docker:build` and Dockerfile will be deleted as soon as we complete the migration to the new static website setup.

## PR Checklist

- [x] Tests for the changes have been added
- [x] `npm test` doesn't throw any error

IMPORTANT: Please review the [CONTRIBUTING.md](https://github.com/amplication/amplication/blob/master/CODE_OF_CONDUCT.md) file for detailed contributing guidelines.

PR #6036, opened 2023-05-17
refactor(client): refactor to static website deployment & workflow changes

levivannoort:
Fixes : https://github.com/amplication/amplication-infrastructure-next/issues/50

Addition to: #5975 

## PR Details

Adds support for building and publishing a static website - in this case for the `client`-component. This also incorporates reworking the affected setup by standardizing the naming convention of `deploy` & `package.

## PR Checklist

### DIFF
--- a/packages/data-service-generator/src/tests/__snapshots__/create-data-service-custom-path.spec.ts.snap
+++ b/packages/data-service-generator/src/tests/__snapshots__/create-data-service-custom-path.spec.ts.snap
@@ -56,7 +56,7 @@ yarn-error.log*
     \\"build\\": \\"react-scripts build\\",
     \\"test\\": \\"react-scripts test\\",
     \\"eject\\": \\"react-scripts eject\\",
-    \\"docker:build\\": \\"docker build .\\"
+    \\"package:container\\": \\"docker build .\\"
   },
   \\"eslintConfig\\": {
     \\"extends\\": [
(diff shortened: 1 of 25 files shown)


packages/data-service-generator/src/tests/__snapshots__/create-data-service-custom-path.spec.ts.snap
```diff
@@ -56,7 +56,7 @@ yarn-error.log*
     \\"build\\": \\"react-scripts build\\",
     \\"test\\": \\"react-scripts test\\",
     \\"eject\\": \\"react-scripts eject\\",
-    \\"docker:build\\": \\"docker build .\\"
+    \\"package:container\\": \\"docker build .\\"
```
overbit, line 59: This change would probably need a change in the README files for the generated apps too.
@abrl91 do we mention docker:build in our docs too?

levivannoort: Ah I think I was too rigoureus with the find and replace, we could leave it that way on the generated side, but it is your call!

overbit: Let's do it for the generated app too since it's a good improvement. Just check the README for now

overbit (approved):

packages/data-service-generator/src/tests/__snapshots__/create-data-service-custom-path.spec.ts.snap
overbit: I've checked the generated README and it is OK

Issue #5941, opened 2023-05-10
🐛 Bug Report: Stigg webhook generate too many errors


overbit:
### What happened?

<img width="1120" alt="image" src="https://github.com/amplication/amplication/assets/2861984/4057c93f-f0e4-4266-b6d0-307a66fe4e0f">

### What you expected to happen

The platform should not fail

### How to reproduce

NA

### Amplication version

v1.5.7

EugeneTseitlin:
Check for existing account and issue a warning when trying to create account more than once.

PR #6039, opened 2023-05-18
check if subscription existence on create and update, issue warning i…

EugeneTseitlin:
…f operation should not occur

Close: #5941

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/subscription/subscription.service.ts
+++ b/packages/amplication-server/src/core/subscription/subscription.service.ts
@@ -260,7 +260,8 @@ export class SubscriptionService {
           await this.update(updateStatusDto.id, updateSubscriptionInput);
         } else {
           this.logger.warn(
-            `Trying to update a non-existing subscription with id: ${updateStatusDto.id}`
+            `Trying to update a non-existing subscription with id: ${updateStatusDto.id}`,
+            updateStatusDto
           );
         }
         break;
--- a/packages/amplication-server/src/core/subscription/subscription.service.ts
+++ b/packages/amplication-server/src/core/subscription/subscription.service.ts
@@ -244,7 +244,8 @@ export class SubscriptionService {
           );
         } else {
           this.logger.warn(
-            `Trying to create an existing subscription with id: ${updateStatusDto.id}`
+            `Trying to create an existing subscription with id: ${updateStatusDto.id}`,
+            updateStatusDto
           );
         }
         break;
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/subscription/subscription.service.ts
```diff
@@ -199,26 +211,41 @@ export class SubscriptionService {
...
-        ) {
-          await this.analyticsService.track({
-            userId: userId,
-            properties: {
-              workspaceId: updateStatusDto.customer.id,
-            },
-            event: EnumEventType.WorkspacePlanUpgradeCompleted,
-          });
+        const subs = await this.getSubscriptionById(updateStatusDto.id);
+        if (subs) {
```
yuval-hazaz, line 238: in case a subscription already exists, I think we should update the details instead of ignoring the event

packages/amplication-server/src/core/subscription/subscription.service.ts
```diff
@@ -199,26 +211,41 @@ export class SubscriptionService {
...
+          await this.create(updateStatusDto.id, createSubscriptionInput);
+          await this.trackUpgradeCompletedEvent(
+            createSubscriptionInput.plan,
+            updateStatusDto.metadata?.userId,
+            updateStatusDto.customer.id
+          );
+        } else {
+          this.logger.warn(
+            `Trying to create an existing subscription with id: ${updateStatusDto.id}`
```
overbit, line 247: Let's add the updateDto to the log that it might give us some more insight about the event.

packages/amplication-server/src/core/subscription/subscription.service.ts
EugeneTseitlin: this line is under "subscription.created" case, which mean that notification about creation was repeated, it's not an update

packages/amplication-server/src/core/subscription/subscription.service.ts
EugeneTseitlin: Done

morhag90 (approved):

packages/amplication-server/src/core/subscription/subscription.service.ts
yuval-hazaz: i just assumed that we if get a notification and the identifier already exist we better update the date then ignore it.
in case this is a duplicate notification - I understand your point. 
but can we guarantee that we will not get this notification for existing users? for example, when a user upgrade from one tier to another, could it be that we get a new notification?

yuval-hazaz (approved):

Issue #6056, opened 2023-05-19
🐛 Bug Report: Generated code is created with wrong args for base services 


overbit:
### What happened?

```sh
ERROR in ./packages/amplication-plugin-api/src/plugin/base/plugin.service.base.ts:21:37
TS2345: Argument of type 'SelectSubset<T, PluginFindManyArgs>' is not assignable to parameter of type 'Subset<PluginCountArgs, PluginCountArgs>'.
  Types of property 'select' are incompatible.
    Type 'T["select"]' is not assignable to type 'true | PluginCountAggregateInputType'.
      Type 'PluginSelect' is not assignable to type 'true | PluginCountAggregateInputType'.
        Type 'PluginSelect' is not assignable to type 'PluginCountAggregateInputType'.
          Type 'T["select"]' is not assignable to type 'PluginCountAggregateInputType'.
            Type 'PluginSelect' is not assignable to type 'PluginCountAggregateInputType'.
              Types of property 'createdAt' are incompatible.
                Type 'boolean' is not assignable to type 'true'.
    19 |     args: Prisma.SelectSubset<T, Prisma.PluginFindManyArgs>
    20 |   ): Promise<number> {
  > 21 |     return this.prisma.plugin.count(args);
       |                                     ^^^^
    22 |   }
    23 |
    24 |   async findMany<T extends Prisma.PluginFindManyArgs>(

ERROR in ./packages/amplication-plugin-api/src/plugin/base/plugin.service.base.ts:32:5
TS2322: Type 'HasReject<RejectOnNotFound | RejectPerOperation, T["rejectOnNotFound"] extends RejectOnNotFound ? T["rejectOnNotFound"] : undefined, "findUnique", "Plugin"> extends 1 ? Prisma__PluginClient<...> : Prisma__PluginClient<...>' is not assignable to type 'Plugin'.
  Type 'Prisma__PluginClient<PluginGetPayload<T>, never> | Prisma__PluginClient<PluginGetPayload<T>, null>' is not assignable to type 'Plugin'.
    Type 'Prisma__PluginClient<PluginGetPayload<T>, never>' is missing the following properties from type 'Plugin': createdAt, description, github, icon, and 7 more.
    30 |     args: Prisma.SelectSubset<T, Prisma.PluginFindUniqueArgs>
    31 |   ): Promise<Plugin | null> {
  > 32 |     return this.prisma.plugin.findUnique(args);
[...]

Issue #5877, opened 2023-04-26
[Security] Update `class-validator` to v0.14 on generated app


overbit:
The amplication generated application should be built with the latest version of `class-validator` to fix the vulnerability described here:  
https://github.com/amplication/amplication/security/dependabot/657
https://security.snyk.io/vuln/SNYK-JS-CLASSVALIDATOR-1730566

## Tech notes

Version 0.14 comes with a breaking change related to `forbidUnknownValues` option default value that has been swapped.
This will cause any generated Dto that is missing the validation metadata (validation decorators) to automatically fail.

⚠️ setting `forbidUnknownValues: false` **is not an acceptable solution** but instead we need to update the dto generation with the correct validator decorators like in the example below showing the current vs the future setup
<img width="1064" alt="image" src="https://user-images.githubusercontent.com/2861984/234594854-42eca5bb-c812-409f-8d24-f519a97454da.png">

ognjenjevremovic:
Hey @overbit, may you assign the issue to me please?  
I would like to give this one a go.

The scope of the issue should be related to `data-service-generator` only?

overbit:
Yes, you are correct, the scope should be just related to the `data-service-generator`. 
Assigned! Good luck

PR #6082, opened 2023-05-22
fix(data-service-generator): wrong args for generated `count` action

ognjenjevremovic:
Close: #6056 

## PR Details

Change the generated code for `ServiceBase` class, to match the valid types for the `count` operation for argument type.  
Update `jest` testing-snapshots to match the code change.

### Summary
🆕🔄🛠️

This pull request updates the data service generator to use the new `COUNT_ARGS` type for the `count` method of the service classes, following the changes in the Prisma client. It also modifies the template files `create-service.ts` and `service.base.template.ts` to reflect this change.

> _`COUNT_ARGS` added_
> _Template mapping changes_
> _Fall of `FIND_MANY`_

### Walkthrough
*  Add a new type identifier `COUNT_ARGS` to the template mapping object for generating service classes ([link](https://github.com/amplication/amplication/pull/6082/files?diff=unified&w=0#diff-9af0fb5e24fec67389a8d46017a442d19725496874d46493feb14029f48ce657R304))
*  Update the base service class to use the `COUNT_ARGS` type for the `count` method and the generic type parameter `T` ([link](https://github.com/amplication/amplication/pull/6082/files?diff=unified&w=0#diff-456f752b8503d5632ca12faeb990b0a26b6389f6edcf71fec1f585cfab3c152bL15-R16))

## PR Checklist


ognjenjevremovic:
Hey @yuval-hazaz I think you might be right.  
Testing the changes locally, from the generated code output, I can see the following in the `user.resolver.base.ts` file:

```ts
  @graphql.Query(() => MetaQueryPayload)
  @nestAccessControl.UseRoles({
    resource: "User",
    action: "read",
    possession: "any",
  })
  async _usersMeta(
    @graphql.Args() args: UserFindManyArgs
  ): Promise<MetaQueryPayload> {
    const results = await this.service.count({
      ...args,
      skip: undefined,
      take: undefined,
    });
    return {
      count: results,
    };
  }
```

I want to draw the attention to the type of `args` passed to the `this.service.count` method call.  
`count` method (on the `UserServiceBase` class) is called with the `args` of type `UserFindManyArgs` in the resolver, where the contract from within `user.service.base.ts` on the generated `UserServiceBase` class accepts the args of type `UserCountArgs` as following:

```ts
  async count<T extends Prisma.UserCountArgs>(
    args: Prisma.SelectSubset<T, Prisma.UserCountArgs>
  ): Promise<number> {
    return this.prisma.user.count(args);
  }
```

Were you referring to this?  
It seems that the `resolver` should be updated as well then?

overbit:
@ognjenjevremovic you are correct. I'd suggest to update the `controller` and `resolver` with a mapping that will translate the input args into the needed args for the `count` method of the `service` class.

ognjenjevremovic:
Hey @overbit I apologise for my slight delay in response.  
I missed your previous comment (it must have flied over my email unnoticed).

I will update the `contoller` and `resolver` templates accordingly. It might take me a day or two, as per busy working week, but I'll try do it asap 🙂.

ognjenjevremovic:
Hey @overbit I've updated the `resolver` argument data mapping to match the expected type for the `count` method on the service class instance.  
The changed required introducing a new `code-gen-type` (or at least, that was my understanding of it), for specifying the `countArgs` argument.

I’ve created a new graphql dto for handling the mapping of the incoming data to `COUNT_ARGS` in the `META_QUERY` handler.  
The snapshots have been updated and the new set of test(s) have been supplied for the `create-count-args` utility, for generating the data mapper.

Please do let me know if everything is in order 🙂.

I was wondering if exposing the `count` handler on the controller would be the next step, since the resolver has the `_metaQuery` for getting the count of the documents matching the query? Would that inline with the goals of the project?  
If so, may I open an issue for extending the controller with the `count` handler and submitting the PR (as I was experimenting with extending the controller while working on the resolver part) 🚀 .

Many thanks!

mulygottlieb:
Thanks a lot @ognjenjevremovic !
We will review this soon.

Your suggestion for the controller is a very good one! but this specific task is reserved for other purposes at the moment so please do not submit such a PR right now 😊

mulygottlieb:
BTW @ognjenjevremovic  if you're looking for an interesting challenge, would you like to look at #5877?

ognjenjevremovic:
Hey @mulygottlieb thank you for your prompt feedback 🚀 .

Sure, I'll take a closer look at #5877 🙂.  
I'd quickly just go through the failing tests first and update them accordingly. As I see some tests are failing after the latest changes (even though the snapshots were updated in the last commits).

overbit (approved):
Nice work @ognjenjevremovic 💯

overbit:
@ognjenjevremovic can you double check the failing test so we can merge? Thanks

overbit (approved):
💯 !!!

Issue #6149, opened 2023-05-31
🐛 Bug Report:  Fix non OS-agnostic directory handling in GitCli


tiagovb:
### What happened?

In the GitCli class, specifically within the method handling file writing to the local filesystem (line 111), there is an issue with how we're determining the parent directory of a file path. The current approach involves manually splitting the file path string, removing the last segment, and joining it back together. The current approach may lead to unexpected behavior if the file path is not in the expected format or if the code is run on different operating systems, as it does not account for OS-specific path conventions.

### What you expected to happen

The retrieval of the parent directory of a file path should be done in a manner that is OS-agnostic, efficient, and reliable. This can be achieved using built-in Node.js functions designed to handle file paths.

### How to reproduce

To reproduce the issue, Amplication must be installed on a Windows machine with GitHub connected. Furthermore, all necessary settings should be in place for creating a service. Most importantly, the amplication-git-pull-request-service must be running with the correct GitHub environment configuration.

- Install the Amplication server and CLI.
- Follow the necessary steps to create a new service, or use an existing one.
- Initiate a build (or rebuild) for the service.
- Observe the Build log failures:
```
Push changes to Github 178s
2023-05-31T01:21:46.395Z (Info) Starting to push changes to Github
2023-05-31T01:24:33.077Z (Info) Push changes to Github failed
2023-05-31T01:24:33.089Z (Info) ENOENT: no such file or directory, mkdir ''"
```

### Amplication version

1.7.0

### Environment

 Node version: v18.3.0
 NPM version: 8.11.0 
 Docker version: Docker version 23.0.5, build bc4487a
 Cloud Provider: N/A
 OS Details (Windows): 
 BuildNumber  Caption                                OSArchitecture  Version
 22623              Microsoft Windows 11 Pro  64-bit                 10.0.22623

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #6150, opened 2023-05-31
fix(git-cli): ensure OS-agnostic file parent directory handling

tiagovb:
Close: #6149

This commit addresses an issue found within the amplication-git-utils package, specifically in the git-cli.ts file. 

Previously, the handling of the file parent directory (`fileParentDir`) was not OS-agnostic, causing issues particularly when running the software on Windows machines. This was manifesting as an ENOENT error, indicating that a required directory could not be created or found.

The changes in this commit include utilizing the `dirname` function from Node.js built-in 'path' module for extracting the parent directory from a file path, which provides a more consistent OS-agnostic solution, instead of spliting string with "/" .


yuval-hazaz:
@tiagovb thanks for your contribution to Amplication 💜
The team will review the PR soon

Issue #6029, opened 2023-05-17
🐛 Bug Report: UI/UX - Middle Window in the main screen Lacks Padding on the Right Side.


alexbass86:
### What happened?

**Issue Summary:**
The app's main screen displays a middle window that needs to be appropriately padded on the right side. 

**Issue Details:**
When accessing the app's main screen, users encounter a middle window not adequately padded on the right side. 
As a result, the window's contents need to be positioned closer to the edge, causing an imbalance in the layout and reducing the overall aesthetic appeal.

![image](https://github.com/amplication/amplication/assets/96179897/f277d7a2-98d3-4901-a982-4b2800bf19a5)

### What you expected to happen

All sides should have a padding of 16px:
padding: 16px 16px 16px 16px;

### How to reproduce

In main project

### Amplication version

1.6.0

### Are you willing to submit PR?

Yes I am willing to submit a PR!

jaysoni-ash42:
@alexbass86 i wan to work on this issue? but   when the scroll bar is shown it has an extra width of 16px so the total would be 32 pixel  without it would be 16px by default .  would that change resolve that issue

yuval-hazaz:
@jaysoni-ash42 thanks for your help
The padding was looking fine in previous versions, so I would investigate what was changed recently and why and consider reverting the change

GreenMachine01:
@mulygottlieb please decide who can take it.
@jaysoni-ash42 you want to take it?

jaysoni-ash42:
@GreenMachine01  currently working on a different issue. plz assign it to somebody else thanks.

mulygottlieb:
@GreenMachine01 let's open it for the community and wait for a community member that wants to take it. @souravjain540 FYI.

abrl91:
passed QA ✅ 

https://github.com/amplication/amplication/assets/39680385/416257d2-f86c-47a6-aea3-cd1c369e29d9

PR #6165, opened 2023-06-02
fix(client): main content padding right

abrl91:
Close: #6029 

### Summary
🖥️🖨️🎨

This pull request improves the appearance and usability of the page content and scrollbar elements in the client package. It adjusts the padding of the `page content` element and updates the scrollbar style in `mixin.scss` to match the design system.

> _`page content` padded_
> _scrollbar style harmonized_
> _winter web design_

### Walkthrough
* Remove right padding of page content element to align with sidebar and header ([link](https://github.com/amplication/amplication/pull/6165/files?diff=unified&w=0#diff-e6fa601f11c49c9c943c08a0cb5257e3312266477ab55e2c43ac7f2abbfda3daL9-R10))
* Restore right padding of page content element for print media query to prevent overflow ([link](https://github.com/amplication/amplication/pull/6165/files?diff=unified&w=0#diff-e6fa601f11c49c9c943c08a0cb5257e3312266477ab55e2c43ac7f2abbfda3daR34))
* Increase opacity of scrollbar thumb to improve visibility ([link](https://github.com/amplication/amplication/pull/6165/files?diff=unified&w=0#diff-32a6894d90022b03852964ffb2a469cfec9880bdb60788a78e9c09f693cba5e0L245-R245))
* Remove hover effect of scrollbar thumb to simplify style ([link](https://github.com/amplication/amplication/pull/6165/files?diff=unified&w=0#diff-32a6894d90022b03852964ffb2a469cfec9880bdb60788a78e9c09f693cba5e0L256-L261))

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Layout/PageContent.scss
+++ b/packages/amplication-client/src/Layout/PageContent.scss
@@ -7,6 +7,7 @@ $tabs-width: 300px;
   flex: 1;
   gap: var(--large-spacing);
   padding: var(--large-spacing);
+  padding-right: 0;
   background: var(--page-content-background);
   box-sizing: border-box;
   overflow: hidden;
@@ -29,7 +30,8 @@ $tabs-width: 300px;
   }
 
   &__main {
-    @include scrollbars(1px, var(--black60), transparent);
+    @include scrollbars(16px, var(--black60), transparent);
+    padding-right: var(--large-spacing);
     flex: 1;
     overflow: auto;
     position: relative;
--- a/packages/amplication-client/src/Layout/PageContent.scss
+++ b/packages/amplication-client/src/Layout/PageContent.scss
@@ -29,7 +29,7 @@ $tabs-width: 300px;
   }
 
   &__main {
-    @include scrollbars(16px, var(--black60), transparent);
+    @include scrollbars(1px, var(--black60), transparent);
     flex: 1;
     overflow: auto;
     position: relative;
--- a/packages/amplication-client/src/Layout/PageContent.scss
+++ b/packages/amplication-client/src/Layout/PageContent.scss
@@ -6,7 +6,7 @@ $tabs-width: 300px;
   display: flex;
   flex: 1;
   gap: var(--large-spacing);
-  padding: var(--large-spacing) 0 var(--large-spacing) var(--large-spacing);
+  padding: var(--large-spacing);
   background: var(--page-content-background);
   box-sizing: border-box;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


abrl91:
@alexbass86 

https://github.com/amplication/amplication/assets/39680385/0444208b-01ed-4836-9e1d-2d0ec0f84d16

yuval-hazaz (commented):
@abrl91 in the video, I don't see the scrollbar when it should be visible

Also, please have a look at this PR - https://github.com/amplication/amplication/commit/74cc977c0c27a19783f2e787e688a54d2cc922df

I see why removing the extra padding settings will revert the change - but:
1. what about the original issue the previous PR tries to fix? Is it resolved? Should we reopen? is it a matter of a specific browser? Specific page?
2. why did you decide to change the scrollbar settings? in the original PR they didn't change that, so I assume the previous state of the scrollbar worked correctly before the problem started with the previous PR

abrl91:
> @abrl91 in the video, I don't see the scrollbar when it should be visible
> 
> Also, please have a look at this PR - [74cc977](https://github.com/amplication/amplication/commit/74cc977c0c27a19783f2e787e688a54d2cc922df)
> 
> I see why removing the extra padding settings will revert the change - but:
> 
> 1. what about the original issue the previous PR tries to fix? Is it resolved? Should we reopen? is it a matter of a specific browser? Specific page?
> 2. why did you decide to change the scrollbar settings? in the original PR they didn't change that, so I assume the previous state of the scrollbar worked correctly before the problem started with the previous PR

1. I don't see the original issue, now we have 16px , like in the design
2. when we open the console, the screen's height gets smaller and the scroller appears but its width was (another) 16px and then the padding-right was doubled, that's why I reduced the width. we can still sroll

yuval-hazaz:
> * I don't see the original issue, now we have 16px , like in the design
> * when we open the console, the screen's height gets smaller and the scroller appears but its width was (another) 16px and then the padding-right was doubled, that's why I reduced the width. we can still sroll

1. I added a link to the original PR that caused the issue.
2. reducing the width of the scroll is not the right solution - unless we decide we want this behaviour across the board, and then we need to implement it in all pages (but I don't think we want to do that). 

we suppose to have 16px padding for the panel, and the scroll should not take extra space and should be positioned in the same space of the padding. If this is not the behavior, we should look for the reason why it is not that way - since it was working like this before. 
have a look at this page for example on how it should looks like 
image

yuval-hazaz (approved):

Issue #6230, opened 2023-06-13
🐛 Bug Report: Service settings are not saved


mulygottlieb:
### What happened?

When you make changes to the service settings in one of these categories:
![image](https://github.com/amplication/amplication/assets/1912933/461e11c4-1a03-41f9-a1a4-56bd368e1ef0)

They are not saved:
https://amplic-my.sharepoint.com/personal/mor_amplication_com/_layouts/15/stream.aspx?id=%2Fpersonal%2Fmor%5Famplication%5Fcom%2FDocuments%2FMicrosoft%20Teams%20Chat%20Files%2FScreen%20Recording%202023%2D06%2D13%20at%2018%2E10%2E10%2Emov&referrer=Teams%2ETEAMS%2DELECTRON&referrerScenario=p2p%5Fns%2Dbim&ga=1

### What you expected to happen

Service settings should save when changing tabs

### How to reproduce

create a service in Amplication, go to the service settings, make changes e.g. to base directories, navigate away, return to those settings - observe your changes were not saved.

### Amplication version

1.7.1

AkhilJ321:
Hey, I wanted to try working on this issue, but I am new to the codebase, I regenerated the error, but am not able to locate the code for the specific. It would be very significant if you could help me with that.

AkhilJ321:
Hey, I am interested in working on this issue,could you please assign it to me?

GreenMachine01:
@souravjain540 FYI

souravjain540:
@AkhilJ321 assigned :) you can get a limited edition amplication shirt for this issue. let me know if you need any help!

AkhilJ321:
Thanks for assigning me. I have been trying to solve the issue since it is posted, but as this is first time  I have been trying to contribute to this project and the project structure is new to me. I found the component where the code is written for this page, but i am not able to find from where it is actually rendering. So it will be really great if you could help me with finding the required files for this bug from which it is actually rendering.

souravjain540:
@mulygottlieb can you help @AkhilJ321 ? maybe you guys can talk in discord support :)

AkhilJ321:
I have solved the issue.I will be posting the PR soon after reading contributing guidelines.

souravjain540:
@AkhilJ321 cool. let me know :)

AkhilJ321:
[...]

PR #6310, opened 2023-06-19
#6230  Fixed the service settings not saved issue.

AkhilJ321:
Close: #6230

## PR Details

Fixed the problem of service settings not getting updated on tab change.
This problem was caused during real-time changes, data was not fetched using the Apollo client, so it was showing previously fetched data. I added the refetch functionality. It is working great now.

## PR Checklist


barshimi:
@AkhilJ321  please move useQuery serviceSettings: models.ServiceSettings; to useResource and expose it to context. then you can use it on all wizard pages and refetch accordingly.

barshimi (changes requested):
@AkhilJ321 look at the comment I wrote at the main conversation page

AkhilJ321:
Thanks for the review.I will try to incorporate them as soon as possible.

AkhilJ321:
@barshimi ,I created the useResource customHook from where now i am performing the query step, but for the expose it to the context part, I am not understanding what to do exactly. I made some changes in appcontextinterface to expose it but then it is disturbing a lot of files .Can you please provide some help with this.

AkhilJ321:
@barshimi @yuval-hazaz .I added some changes according to the change request. Please review them, I could not understand the context part properly, so I missed that. It will be really helpful if you could give more insights if other changes are still needed.

morhag90:
Hi @AkhilJ321, 
Thank you for your contributions :) 

You need to move the query to a file named: resourcesQueries.ts, then delete the new file you created named: useResources and use this query in the existing file under workspaces/hooks/useResources. In this file, you should expose the refetch and the data and use them in your component.

AkhilJ321:
Hi @morhag90  ,I added the queries in the useResource hook file but when I am using this in generatesettingform component ,now I have to pass four arguments ,I am getting confused about what to pass in that , and there is already useQuery defined,so now if I am adding another query,refetch and data are getting clashed.

AkhilJ321:
Hi @morhag90 ,I have added all the recommended changes now.Let me know if something more is required.

AkhilJ321:
@yuval-hazaz I made the mentioned changes several days ago. It would be great if someone could review them.

PR #6354, opened 2023-06-22
fix(client): hide demo repo banner if a git organziation was already connected

mulygottlieb:
Close: #[issue-number]

## PR Details

copilot:all

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
+++ b/packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
@@ -368,18 +368,23 @@ const WorkspaceHeader: React.FC<{}> = () => {
         </div>
       </div>
 
-      {currentProject?.useDemoRepo && (
-        <div className={`${CLASS_NAME}__highlight`}>
-          Notice: You're currently using a preview repository for your generated
-          code. For a full personalized experience, please&nbsp;
-          <Link
-            title={"Go to project settings"}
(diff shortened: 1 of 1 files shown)


packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -368,18 +368,23 @@ const WorkspaceHeader: React.FC<{}> = () => {
...
-          >
-            connect to your own repository
-          </Link>
-        </div>
-      )}
+      {currentProject?.useDemoRepo &&
+        (currentProject?.resources?.length === 0 ||
+          currentProject?.resources.find(
+            (resource) =>
+              resource.gitRepository?.name !== currentProject?.demoRepoName
```
yuval-hazaz, line 375: instead of changing the condition, I think the fix should be about updating the cache on the client so the value of `currentProject?.useDemoRepo` is up to date

yuval-hazaz (changes requested):

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
mulygottlieb: You are right. I tested without the changes, and it's the same result. I missed the fact that the problem only exists in the interim after connecting an org but **before** choosing a repo. Will try to figure out how to update the cache.

Issue #6121, opened 2023-05-26
🐛 Bug Report: services with entities with JSON field cannot run successfully on monorepo


overbit:
### What happened?

<img width="833" alt="image" src="https://github.com/amplication/amplication/assets/2861984/90bf4bca-a213-4cc0-b529-65c95546d855">

### What you expected to happen

Creating a service with an entity with a JSON field will produce a code that depends on `type-fest`.
Since this dependency should only be a `devDependency` it will cause the application to not start successfully.

 

### How to reproduce

Create a service with an entity with a JSON field

### Amplication version

v.1.6.5

Rahulcheerio:
Hii @overbit , want to contribute to this bug , can you please explain more how to create the service .

overbit:
@Rahulcheerio sorry for the late reply. This issue needed more internal discussions and we decided to not open it to the community.
Please have a look to any issue tagged as `open to community` https://github.com/amplication/amplication/labels/open%20to%20community

PR #6357, opened 2023-06-22
fix(dsg): add `type-fest` as dependency 

overbit:
Close: #6121 

## PR Details

This solution represent a temporary solution until we refactor the DSG to understand difference between `imports` and `imports types`. 
Types, i.e. types from `type-fest` like `JsonValue`, should be imported as `import type {} from ""` to avoid needing the dependency at runtime. 
This pr configure `type-fest` as dependency  instead of devDependency to unblock any user that run the service installing only the production dependencies
 

## PR Checklist


abrl91 (approved):

tupe12334:
@overbit Don't you want to open an issue of undoing that change when you have a fix to the import type problem?

PR #6378, opened 2023-06-26
fix routes to handle wrong paths [Fixes #4424]

### LINKED ISSUE #4424: 🐛 Bug Report: No default route
### What happened?

When changing the path in the url to something that doesn't exist, the application will load a blank page.

I would expect that when you try to visit something that doesn't exist it would route to the home page or a custom error page.

### How to reproduce

When you visit the following url you run into this issue: "application url/example-path-that-doesnt-exist"

https://app.amplication.com/example-subpath-that-doesnt-exist

If the url contains a sub-path the routing does use the sub path as the default route: "application-url/subpath-uuid/example-subpath-that-doesnt-exist"

### Amplication version

v1.1.0

### Environment

_No response_

_No response_

anshgoyalevil:
Close: #4424 

## PR Details

This PR resolves the issue where the routes where not being redirected to the 404 page on modifying them to something which doeesn't exist in react router.

Amplication uses regex patterns to handle the patterns so the minimum length of all those patterns is supposed to be more than 20 according to ```([A-Za-z0-9-]{20,})```

Now, all the unhandled routes would be directed to ```NotFoundPage```

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/routes/appRoutes.tsx
+++ b/packages/amplication-client/src/routes/appRoutes.tsx
@@ -1,6 +1,5 @@
 import { ComponentType, lazy, LazyExoticComponent } from "react";
 import resourceRoutes from "./resourceRoutes";
-import NotFoundPage from "../404/NotFoundPage";
 
 export interface RouteDef {
   path: string;
--- a/packages/amplication-client/src/routes/appRoutes.tsx
+++ b/packages/amplication-client/src/routes/appRoutes.tsx
@@ -183,7 +183,7 @@ export const Routes: RouteDef[] = [
       // Fallback route for wrong paths
   {
     path: "*",
-    Component: NotFoundPage,
+    Component: lazy(() => import("../404/NotFoundPage")),
     exactPath: false,
     isAnalytics: false,
   },
--- a/packages/amplication-client/src/routes/appRoutes.tsx
+++ b/packages/amplication-client/src/routes/appRoutes.tsx
@@ -1,5 +1,6 @@
 import { ComponentType, lazy, LazyExoticComponent } from "react";
 import resourceRoutes from "./resourceRoutes";
+import NotFoundPage from "../404/NotFoundPage";
 
 export interface RouteDef {
   path: string;
@@ -178,5 +179,12 @@ export const Routes: RouteDef[] = [
     routeTrackType: "signup",
     exactPath: true,
     isAnalytics: true,
+  },
+      // Fallback route for wrong paths
+  {
+    path: "*",
+    Component: NotFoundPage,
+    exactPath: false,
+    isAnalytics: false,
   },
 ];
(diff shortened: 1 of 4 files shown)


packages/amplication-client/src/routes/appRoutes.tsx
```diff
@@ -178,5 +179,12 @@ export const Routes: RouteDef[] = [
     routeTrackType: "signup",
     exactPath: true,
     isAnalytics: true,
+  },
+      // Fallback route for wrong paths
+  {
+    path: "*",
+    Component: NotFoundPage,
```
yuval-hazaz, line 186: There are TS errors when using this code since the Component property expects to get a type of  LazyExoticComponent<ComponentType<any>>;

instead use this code (also remove the import at the begging of the page)

same for all the files below

```suggestion
    Component: lazy(() => import("../404/NotFoundPage")),
```

yuval-hazaz (changes requested):

anshgoyalevil:
@yuval-hazaz Done with the changes. Thanks for pointing them out. 🚀

yuval-hazaz:
@anshgoyalevil 
Thanks for the update. 
I ran some local tests and saw this behavior
please fix this issue and test the solution on other routes
image

yuval-hazaz (changes requested):
see my comment and screenshot

anshgoyalevil:
@yuval-hazaz Can you please share the documentation link for testing the changes locally? I checked the Amplication website and found the instructions for setting up dev environment for just source code generated through amplication.

After following the instructions on the README.md file, I was able to run the server and client, but couldn't login to the localhost client's login screen.

yuval-hazaz:
> @yuval-hazaz Can you please share the documentation link for testing the changes locally? I checked the Amplication website and found the instructions for setting up dev environment for just source code generated through amplication.
> 
> After following the instructions on the README.md file, I was able to run the server and client, but couldn't login to the localhost client's login screen.

take a look at the README at the root of the project
https://github.com/amplication/amplication

Follow the instructions under "Development"
image

anshgoyalevil:
@yuval-hazaz 
Was able to setup in my local machine.

I tried finding the root cause of this, and it couldn't be solved the way we are handling routes.

Actually, we are parsing the routes using ```regex``` and using those values to get the data using graphql. Now the problem here is that we must allow all routes, no matter their pattern.

For example,
If we enter ```http://localhost:3001/cljdyt01o0002u0mc5vd910uh/cljdyt02d000cu0mcc0cpzxn8\abcdefghijklmnopqrst```

Here, ```abcdefghijklmnopqrst``` is being treated as the ```resourceId``` and thus is being used to create graphql connection. On receiving the 404 (or any other server-side error), we would need to render the 404 react component.

overbit:
As per https://github.com/amplication/amplication/pull/6378#issuecomment-1609086405 we are closing this issue as we are not going to fix the behaviour now.

Issue #6379, opened 2023-06-26
🐛 Bug Report: 


kush002:
### Issue: Project side bar compressed in empty state which leads to button overflow

![issue](https://github.com/amplication/amplication/assets/16662383/018fe807-0c7f-4845-8ed1-0a73a40a8f95)

## Resolution: When no project in workspace, project side bar should maintain its width like this

![Resolution](https://github.com/amplication/amplication/assets/16662383/eef40018-c501-4777-a776-92716d4a5814)

### When workspace is not empty project sidebar looks like this... 

![withProject](https://github.com/amplication/amplication/assets/16662383/f6b45843-c38d-484b-828d-cd2863bea3d5)

**when user delete all the project it looks like this...**

![without_project](https://github.com/amplication/amplication/assets/16662383/a6904c0d-abee-44b1-9a3d-fd5ea8b71153)

**Leads to **overflow** of create new workspace button**

### Amplication version

v1.7.6

### Are you willing to submit PR?

Yes I am willing to submit a PR!

kush002:
I already resolved it submitting PR. Check and if you want any correction let me know

PR #6380, opened 2023-06-26
fix(client): project sidebar button overflow

kush002:
Close: #6379 

## PR Details

Project Sidebar Button Overflow When Empty


package.json
```diff
@@ -275,7 +276,7 @@
     "npm-run-all": "^4.1.5",
     "nx": "15.4.5",
     "path-browserify": "^1.0.1",
-    "prisma": "^4.15.0",
+    "prisma": "^4.16.1",
```
yuval-hazaz: Please revert these changes from this PR since they are not part if the scope

yuval-hazaz:
thanks @kush002 while I was reviewing your code I noticed several other differences in the project list when it is empty compared to its full state.

I will approve and accept your PR since it is actually fixing the issue as requested! thanks for that!

In addition, I will also fetch your branch and make a few more changes 

I will duplicate the content from packages/amplication-client/src/Project/ProjectPage.tsx  into packages/amplication-client/src/Project/ProjectEmptyState.tsx

yuval-hazaz (approved):

Issue #6355, opened 2023-06-22
Improve notification about creating user entity (avoid pop-up over pop-up)


GreenMachine01:
![Image](https://user-images.githubusercontent.com/112330500/247983004-c6e384f9-d42c-4178-a4fc-544d63b20953.png)

**Overview:**
We now support for the user to add back the 'User' entity if it was deleted. the current flow involves showing a pop-up over another pop-up. This is not a good UI/UX solution.

**Requirements:**
1. Before showing the "Restore 'User' Entity?" notification close the "Create entity" pop-up
2. Clicking X on the "Restore 'User' Entity?" notification should close the pop-up and return the user to where he was (not return the user to the "Create entity" pop-up)

kush002:
Please assign me this @morhag90

kush002:
I have submitted PR. Please look into it and let me know if need any correction.

barshimi:
on hold. needs to get @yuval-hazaz  review

PR #6387, opened 2023-06-26
Fix/6355 fix pop up over pop up

kush002:
Close: #6355 

## PR Details

**Before PR Submission**: when user click on Add entity button. Create Entity popup appears and after that when user enter  'User' as a entity name new popup will appear over "create entity popup"

popupold

**After PR Submission**: when user click on Add entity button. Create Entity popup appears:
1. Before showing the "Restore 'User' Entity?" notification, It closes the "Create entity" pop-up
2. Clicking X on the "Restore 'User' Entity?" notification should close the pop-up and return the user to where he was (not return the user to the "Create entity" pop-up)

popupnew


morhag90:
create branch from fork

mulygottlieb:
hello @kush002 
I apologize but we decided that from a product perspective, we desire to handle this in a different way and not just manipulate the popups (closing and opening). So until we decide how exactly we want to handle this, I am closing this PR.
Thanks a lot for your contribution! And apologies for needing to close the PR after the time you invested in it!

Issue #6333, opened 2023-06-21
As a User - I want to see the most recent state in the UI after importing a Prisma schema


abrl91:
Related issue: #4290 
Related PR: #6164 

After importing a Prisma schema, the newly created entities are not visible on: 
- the pending changes - fixed by #6367 
-  in the related entity dropdown - fixed by - #6398 

We need to re-fetch in these places.

mulygottlieb:
This only works after a page refresh, so I am re-opening this one (not a blocker for release)

PR #6398, opened 2023-06-27
Feat(client): no-fetch on entity list options relation field

abrl91:
Close: #6333 

## PR Details

copilot:all

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Components/EntitySelectField.tsx
+++ b/packages/amplication-client/src/Components/EntitySelectField.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React, { useEffect } from "react";
 import { gql, useQuery } from "@apollo/client";
 import { SelectField, SelectFieldProps } from "@amplication/ui/design-system";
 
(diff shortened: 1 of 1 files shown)


packages/amplication-client/src/Components/EntitySelectField.tsx
```diff
@@ -11,27 +11,37 @@ type TEntities = {
...
 };
 
 const EntitySelectField = ({ resourceId, ...props }: Props) => {
+  const [relatedEntityListOptions, setRelatedEntityListOptions] =
+    React.useState<RelatedEntityOption[]>([]);
   const { data: entityList } = useQuery<TEntities>(GET_ENTITIES, {
     variables: {
       resourceId: resourceId,
     },
+    fetchPolicy: "no-cache",
```
overbit, line 30: Is this on purpose? We shouldn't bypass cache apart for specific scenarios such as queries that return one time tokens or similar.

overbit (changes requested):

packages/amplication-client/src/Components/EntitySelectField.tsx
abrl91: Yes. This is on purpose. Without it, it just doesn't work

packages/amplication-client/src/Components/EntitySelectField.tsx
```diff
@@ -11,27 +11,37 @@ type TEntities = {
...
-      : [];
+  useEffect(() => {
+    if (!entityList) return;
+
+    const relatedEntityOptions = entityList.entities.map((entity) => ({
+      value: entity.id,
+      label: entity.displayName,
+    }));
+
+    setRelatedEntityListOptions(relatedEntityOptions);
```
yuval-hazaz, line 41: why are we using useEffect + useState, instead of the useMemo that was used before?

packages/amplication-client/src/Components/EntitySelectField.tsx
yuval-hazaz, line 30: instead of no-cache we should use refetchQueries option on the mutation in the import page (packages/amplication-client/src/Entity/ImportPrismaSchema/EntitiesImport.tsx)

yuval-hazaz (changes requested):

packages/amplication-client/src/Components/EntitySelectField.tsx
barshimi: you are using useMemo in a wrong way. useMemo should memoized the result of a logic function. react component has a life cycle which controlled by state and state listeners (useEffect). this is the right way to do it, in sense of React tools and more important the render cycle .

packages/amplication-client/src/Components/EntitySelectField.tsx
barshimi: refetchQueries is an option to refetch different query under specific mutation... why do you want to make two requests of the same query when you can do only one. other than that, we are managing our global/local state under React and not under Apollo. so, no-cache is important here as the component render it should get "fresh" data from server.

barshimi: this is probably due to same  requests that has been done before this component got rendered. this is bad performance and practice. this is the reason she need to cancel previous cache

overbit: can we force the cache for this instead of never caching?

barshimi: you don't manage cache in two places. you do once source of truth

PR #6457, opened 2023-07-04
fix(server): segment tracking crashes when no session ID exists (e.g. request from Stigg or Kafka rather than from the client)

mulygottlieb:
Close: https://github.com/amplication/private-issues/issues/16
Close: https://github.com/amplication/private-issues/issues/10

### Summary
🐛🔧🛠️

Fixed a server crash bug caused by missing analytics middleware. Added optional chaining to `req.analyticsSessionId` in `segmentAnalytics.service.ts`.

> _`req` may be null_
> _Optional chaining saves us_
> _Winter bug is fixed_

### Walkthrough
*  Prevent server crash when analytics middleware is not initialized by adding optional chaining operator to `req.analyticsSessionId` ([link](https://github.com/amplication/amplication/pull/6457/files?diff=unified&w=0#diff-18685efac1e573e80e4297b774d6d17dae8ccbbcde9db01faa86aa7740526745L87-R87))

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/services/segmentAnalytics/segmentAnalytics.service.ts
+++ b/packages/amplication-server/src/services/segmentAnalytics/segmentAnalytics.service.ts
@@ -84,7 +84,7 @@ export class SegmentAnalyticsService {
     if (!this.analytics) return;
 
     const req = RequestContext.currentContext.req;
-    const analyticsSessionId = req.analyticsSessionId;
+    const analyticsSessionId = req?.analyticsSessionId;
 
     this.analytics.track({
       ...data,


barshimi (approved):

packages/amplication-server/src/services/segmentAnalytics/segmentAnalytics.service.ts
overbit, line 1: All good. It would make sense to change the following too 
```
analyticsSessionId: analyticsSessionId
```

to 
```
analyticsSessionId
```

overbit (approved):

packages/amplication-server/src/services/segmentAnalytics/segmentAnalytics.service.ts
mulygottlieb: agreed but per prior agreements, we do not change other people's code if there is no bug or strong reason to do so, and since this is not my code nor a bug, I am leaving this as-is for now.

PR #6461, opened 2023-07-04
fix(client): refresh created repo ui

morhag90:
Close: #[issue-number]

## PR Details

Fix refresh selected repo after a new repo creation. 

copilot:all

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
+++ b/packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
@@ -82,6 +82,7 @@ const SyncWithGithubPage: React.FC = () => {
             resource={data.resource}
             onDone={handleOnDone}
             gitRepositorySelectedCb={gitRepositorySelectedCb}
+            gitRepositoryCreatedCb={handleOnDone}
           />
         )}
         {!isProjectConfiguration && data?.resource && (
@@ -89,6 +90,7 @@ const SyncWithGithubPage: React.FC = () => {
             resource={data.resource}
             onDone={handleOnDone}
             gitRepositorySelectedCb={gitRepositorySelectedCb}
+            gitRepositoryCreatedCb={handleOnDone}
           />
         )}
         <Snackbar open={Boolean(error)} message={errorMessage} />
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
```diff
@@ -82,13 +82,15 @@ const SyncWithGithubPage: React.FC = () => {
             resource={data.resource}
             onDone={handleOnDone}
             gitRepositorySelectedCb={gitRepositorySelectedCb}
+            gitRepositoryCreatedCb={handleOnDone}
```
abrl91, line 85: why do we need two props to hold the same value?

abrl91: `onDone={handleOnDone}` and `gitRepositoryCreatedCb={handleOnDone}`

packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
```diff
@@ -82,13 +82,15 @@ const SyncWithGithubPage: React.FC = () => {
...
           />
         )}
         {!isProjectConfiguration && data?.resource && (
           <ServiceConfigurationGitSettings
             resource={data.resource}
             onDone={handleOnDone}
             gitRepositorySelectedCb={gitRepositorySelectedCb}
+            gitRepositoryCreatedCb={handleOnDone}
```
abrl91, line 93: same here

packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
morhag90: To trigger the refetch method in two different cases

packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
morhag90: Same answer :)

mulygottlieb (approved):

Issue #6296, opened 2023-06-18
As PO - I want a message in the build log when the app generation finished successfully


mulygottlieb:
### Feature description

Today, if you generate code without being connected to a git provider, you will not get the `Successfully pushed changes to Github` message, and your last message on the build log would be `Formatting code...`

We want, when the code generation finished (and the green checkmark is marked next to 
![image](https://github.com/amplication/amplication/assets/1912933/0b065989-45de-482a-9da1-32ab2cc8c464)

To also add a line to the build log saying:
`Successfully Generated Application`

### Use case

Clearer build log when no git provider is used

kirtanchandak:
Can I work on this issue?

ashutosh887:
@souravjain540 let me work on this

jeffmchang:
## Proposal

### Please re-state the problem that we are trying to solve in this issue.

- The current issue is that the panel for the 'Generate Code' is not properly displaying the correct status:

![image](https://github.com/amplication/amplication/assets/109321774/c5faaffb-bc7a-4d2c-b9e0-a7105a2c3d81)


### What is the root cause of that problem?

- The root cause of the problem seems to be coming from BuildSteps.tsx on line 27:

![image](https://github.com/amplication/amplication/assets/109321774/8c30a09e-6a76-4a1f-89dd-dece6cc3de17)


- The `data.build.action.steps.find((step) => step.name === GENERATE_STEP_NAME)` is calling for a `step.name` that is `GENERATE_APPLICATION`

![image](https://github.com/amplication/amplication/assets/109321774/8b6e87d8-5ae6-4877-9618-42f649f4aa20)


- But upon successful completion, the object returns`ADD_TO_QUEUE`:

![image](https://github.com/amplication/amplication/assets/109321774/20c1354f-6448-4010-bc32-61f9869036b9)



### What changes do you think we should make in order to solve the problem?

- Changing `GENERATE_STEP_NAME` to `ADD_TO_QUEUE` leads to the expected result:

![image](https://github.com/amplication/amplication/assets/109321774/e1964c5c-7fd1-4d9d-bbca-2a27f782dd0b)
![image](https://github.com/amplication/amplication/assets/109321774/eb14816a-30fe-4856-b31f-b11de54a7e25)

professorabhay:
Hey @souravjain540, If that issue is still not resolved then please assign it to me. I would like to contribute!

souravjain540:
@jeffmchang do you want to work on this issue?

[...]

PR #6540, opened 2023-07-19
issue #6296, checkmark now shows up on successful build

jeffmchang:
## PR Details
This PR only fixes the checkmark, but I will be working on the system returning a build log message in the next PR

### Please re-state the problem that we are trying to solve in this issue.

- The current issue is that the panel for the 'Generate Code' is not properly displaying the correct status:

image

### What is the root cause of that problem?

- The root cause of the problem seems to be coming from BuildSteps.tsx on line 27:

image

- The `data.build.action.steps.find((step) => step.name === GENERATE_STEP_NAME)` is calling for a `step.name` that is `GENERATE_APPLICATION`

image

- But upon successful completion, the object returns`ADD_TO_QUEUE`:

image

### What changes do you think we should make in order to solve the problem?

- Changing `GENERATE_STEP_NAME` to `ADD_TO_QUEUE` leads to the expected result:

image
image

### DIFF
--- a/packages/amplication-client/src/VersionControl/ActionLog.tsx
+++ b/packages/amplication-client/src/VersionControl/ActionLog.tsx
@@ -121,7 +121,7 @@ const ActionLog = ({ action, title, versionNumber }: Props) => {
         )}
       </div>
       <div className={`${CLASS_NAME}__body`}>
-        {logData.map((stepData) => (
+        {logData.map((stepData, index) => (
           <div className={`${CLASS_NAME}__step`} key={stepData.id}>
             <div className={`${CLASS_NAME}__step__row`}>
               <span
@@ -136,6 +136,7 @@ const ActionLog = ({ action, title, versionNumber }: Props) => {
               <span className={`${CLASS_NAME}__step__message`}>
                 {stepData.message}
               </span>
+
               <span className={`${CLASS_NAME}__step__duration`}>
                 {stepData.duration}
               </span>
--- a/packages/amplication-client/src/VersionControl/BuildSteps.tsx
+++ b/packages/amplication-client/src/VersionControl/BuildSteps.tsx
@@ -24,7 +24,7 @@ export const EMPTY_STEP: models.ActionStep = {
   message: "",
 };
 
-export const GENERATE_STEP_NAME = "GENERATE_APPLICATION";
+export const GENERATE_STEP_NAME = "ADD_TO_QUEUE";
 export const BUILD_DOCKER_IMAGE_STEP_NAME = "BUILD_DOCKER";
 export const DEPLOY_STEP_NAME = "DEPLOY_RESOURCE";
 export const PUSH_TO_GIT_STEP_NAME = (gitProvider: models.EnumGitProvider) =>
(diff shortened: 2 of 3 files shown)


packages/amplication-client/src/VersionControl/BuildSteps.tsx
```diff
@@ -47,13 +47,20 @@ const BuildSteps = ({ build }: Props) => {
     if (!data.build.action?.steps?.length) {
       return EMPTY_STEP;
     }
+    console.log(
+      "TESTING DATA -----------------",
+      data.build.action.steps[0].status || EMPTY_STEP
+    );
+    console.log("DATA STEPS -----------------", data.build.action.steps);
```
yuval-hazaz, line 54: remove this code

packages/amplication-client/src/VersionControl/BuildSteps.tsx
```diff
@@ -47,13 +47,20 @@ const BuildSteps = ({ build }: Props) => {
...
+
     return (
       data.build.action.steps.find(
         (step) => step.name === GENERATE_STEP_NAME
       ) || EMPTY_STEP
     );
   }, [data.build.action]);
 
+  console.log("generateCode STATUS-------------", stepGenerateCode);
```
yuval-hazaz, line 63: remove this code

packages/amplication-client/src/VersionControl/ActionLog.tsx
```diff
@@ -121,7 +121,7 @@ const ActionLog = ({ action, title, versionNumber }: Props) => {
         )}
       </div>
       <div className={`${CLASS_NAME}__body`}>
-        {logData.map((stepData) => (
+        {logData.map((stepData, index) => (
```
yuval-hazaz, line 124: why do we need this change?

packages/amplication-client/src/VersionControl/BuildSteps.tsx
```diff
@@ -24,7 +24,7 @@ export const EMPTY_STEP: models.ActionStep = {
   message: "",
 };
 
-export const GENERATE_STEP_NAME = "GENERATE_APPLICATION";
+export const GENERATE_STEP_NAME = "ADD_TO_QUEUE";
```
yuval-hazaz, line 27: I am not sure this change is expected, but since I don't see a link to an issue I am not sure what id the goal.

yuval-hazaz (changes requested):
Thanks @jeffmchang  for the PR
few comments:
1. In case this PR is fixing an existing issue, please add a reference to an issue number. 
2. Your PR includes changes in package.json and package-lock.json which seems not related to the actual fix. please exclude these changes from the PR. 
3. Your pr includes several unnecessary changes like "console.log" command and other changes. please do not include any changes that are not part of the fix

overbit:
@jeffmchang did you have any time to address @yuval-hazaz comments?

mulygottlieb:
Thanks for your contribution @jeffmchang !
Unfortunately, since there is no response to the CR comments on this PR, I will close it for now due to inactivity. If you do want to keep working on it and address the comments, feel free to open a new PR and address the comments there. Sorry for the inconvenience.

PR #6555, opened 2023-07-22
feat(git): create amplication branch from *latest* amplication commit on base branch

mulygottlieb:
Close: #6647

## PR Details

The concepts here are:

1. We create the amplication branch from the latest amplication build in the source branch (e.g. main) rather than the first commit on the source branch
2. As a result of (1), we removed the cherry-picking process altogether (no longer needed)
3. We now use two diffs and patch with them - one for the potential existing changes that already exist on the amplication branch (custom code on the amplication branch), and one for the diff between the latest previous amplication build, and the new one - so that we only patch the differences rather than whole files (which helps a lot not to run over changes in the source branch)

### Summary
🌲🧑‍💻🔀

Improve branch creation logic in `git-client.service.ts`. Use the latest user commit as the branch point and ignore bot commits.

> _`Branch` from the base, escape the bot's control_
> _Find the latest `commit` that reflects your soul_
> _Filter out the noise, the `parameter` is your key_
> _Create a new reality, unleash your code's fury_

### Walkthrough
*  Add a parameter to control whether to include bot commits in the log result ([link](https://github.com/amplication/amplication/pull/6555/files?diff=unified&w=0#diff-7030ed1d7460a9345852066fcdbb545162d9aa94cf278b82f9d0e5eebe1a9600L351-R359))
*  Get the latest user commit hash on the base branch instead of the first commit hash ([link](https://github.com/amplication/amplication/pull/6555/files?diff=unified&w=0#diff-7030ed1d7460a9345852066fcdbb545162d9aa94cf278b82f9d0e5eebe1a9600L441-R455))
*  Use the latest user commit hash to create the new branch and comment out the unnecessary checkout of the base branch ([link](https://github.com/amplication/amplication/pull/6555/files?diff=unified&w=0#diff-7030ed1d7460a9345852066fcdbb545162d9aa94cf278b82f9d0e5eebe1a9600L451-R467))

## PR Checklist


libs/util/git/src/git-client.service.ts
```diff
@@ -282,29 +281,41 @@ export class GitClientService {
       baseBranch,
     });
 
-    const { diff } = await this.preCommitProcess({
+    // calculate diff on everything *already existing* on the amplication branch since last amplication build
+    const preCommitDiff = await this.preCommitProcess({
```
overbit, line 285: Since the precommit method is now on charge of 2 operations, would it make more sense to rename it in something more specific?

libs/util/git/src/git-client.service.ts
```diff
@@ -282,29 +281,41 @@ export class GitClientService {
...
+    if (preCommitDiff.diff) {
+      await this.applyPostCommit(
+        preCommitDiff.diff,
+        owner,
+        repositoryName,
+        gitCli
+      );
+    }
+    // on top of the already existing changes, apply the diff between the previous amplication build and the new one
+    if (postCommitDiff.diff) {
```
overbit, line 312: How can this not wipe the conflict resolution chosen committed by the user before the new build?

libs/util/git/src/git-client.service.spec.ts
overbit, line 1: Let's remove that unused import are removed.

libs/util/git/src/git-client.service.ts
mulygottlieb: I couldn't reproduce an example of how this would happen, if you have a scenario in mind, please test it with this branch and see if you can create the problem you are describing in reality.

mulygottlieb:
@overbit @morhag90 I added tests, and now the CI passes. Feel free to do some testing on the branch, and if everything looks good, you can go ahead and merge it to `next`.

libs/util/git/src/git-client.service.ts
overbit: We will test it properly on sandbox with a simple repo

overbit (approved):

Issue #6557, opened 2023-07-25
🐛 Bug Report: The relations in the order template service are set in an incorrect way


abrl91:
### What happened?

The relation between order and product should be one-to-many

![image](https://github.com/amplication/amplication/assets/39680385/05dbb55e-8e89-4f04-ba1b-f4ced28fd4a3)

same for customer and orders

![image](https://github.com/amplication/amplication/assets/39680385/b7b31859-5406-4467-86b0-2b0519fc1d1c)

### What you expected to happen

The relations should be one-to-many 

### How to reproduce

Create a service from a template

### Amplication version

1.8.0

mulygottlieb:
This seems to have fixed the relation between customers and orders, but the relation between orders and products is still 1:1


![Image](https://github.com/amplication/amplication/assets/1912933/a44153ee-91ec-4531-b30c-120ca440b887)


Re-opening

abrl91:
this is exactly as it was before, I have an old service in prod

12 months ago
![image](https://github.com/amplication/amplication/assets/39680385/3d6b362e-e8c7-4d89-b1be-425d1ea5fe83)

![image](https://github.com/amplication/amplication/assets/39680385/fdc1309b-9acc-4ec7-969f-0145120d4c6b)

![image](https://github.com/amplication/amplication/assets/39680385/90c2df2d-2f2b-4525-aaeb-7e7f7859ce9a)

PR #6558, opened 2023-07-25
fix(server): order template service relations

abrl91:
Close: #6557 

### Summary
🔄🔗🗃️

This pull request adds support for inverse relationship fields in the entity service. It modifies the `entity.service.ts` file to create and update the inverse fields with the appropriate settings.

> _Sing, O Muse, of the skillful coder who devised_
> _A way to link the fields of entities with grace_
> _And set the `createFieldArgs` to match the case_
> _Of each inverse relationship that he revised._

### Walkthrough
*  Set the inverse relationship field's cardinality to the opposite of the original field's cardinality ([link](https://github.com/amplication/amplication/pull/6558/files?diff=unified&w=0#diff-9a1aca769e1c8ddba4436f7b850d36912e8fa38e946d5c40d68631be40167e59R1994-R1996))
*  Remove the default value of `false` for the `allowMultipleSelection` property of the `createFieldArgs.properties` object to avoid overriding the previous change ([link](https://github.com/amplication/amplication/pull/6558/files?diff=unified&w=0#diff-9a1aca769e1c8ddba4436f7b850d36912e8fa38e946d5c40d68631be40167e59L2407-R2410))

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -1981,19 +1981,22 @@ export class EntityService {
     // In case created data type is Lookup define related field names according
     // to the entity
     if (data.dataType === EnumDataType.Lookup) {
-      const { allowMultipleSelection } =
+      const { allowMultipleSelection: entityFieldAllowMultipleSelection } =
         data.properties as unknown as types.Lookup;
 
       createFieldArgs.relatedFieldName = camelCase(
-        !allowMultipleSelection ? entity.pluralDisplayName : entity.name
+        !entityFieldAllowMultipleSelection
+          ? entity.pluralDisplayName
+          : entity.name
       );
 
-      createFieldArgs.relatedFieldDisplayName = !allowMultipleSelection
-        ? entity.pluralDisplayName
-        : entity.displayName;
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -1991,6 +1991,9 @@ export class EntityService {
       createFieldArgs.relatedFieldDisplayName = !allowMultipleSelection
         ? entity.pluralDisplayName
         : entity.displayName;
+
+      createFieldArgs.relatedFieldAllowMultipleSelection =
+        !allowMultipleSelection;
```
overbit, line 1996: The name of the property seems to describe the opposite. Is it meant to be *NOT AllowMultipleSelection*?

abrl91: Yes
This is how it was before we changed it to use the args, so I just changed the args to have this value
image

overbit (approved):

packages/amplication-server/src/core/entity/entity.service.ts
overbit: Fixed the naming for better understanding

overbit (approved):

Issue #5428, opened 2023-03-15
As User- I want to easily view and edit the plugin settings in the code view window


tupe12334:
**Overview:**
As a part of our ongoing efforts to improve the user experience within our application, we need to add a feature that automatically formats JSON data in a readable manner after the user clicks on the "Save" button.

**Current:**
![image](https://github.com/amplication/amplication/assets/112330500/bbb486b2-d5c5-4f33-92c6-24b9f1c1aae0)

**Requested:**
![image](https://github.com/amplication/amplication/assets/112330500/4dfa0840-a02e-4286-aed8-482e70c13a9d)

**Acceptance Criteria:**

1. After entering JSON data in the code viewer and clicking 'Save', the system automatically formats the JSON text to be more readable.
2. Any default settings will be shown in the prettified format when the user opens the settings of a plugin.

GreenMachine01:
@tupe12334 
can you add an example of current versus expected?

mulygottlieb:
@GreenMachine01, if you or @alexbass86 add a visual example of the required changes, we can open it for the community. But we need a clear example or explanation of the desired result to give it to someone.

riju951:
Can you assign me this issue ?

souravjain540:
@riju951 i assigned another issue to you. let someone other work on this :)

iriteshmishra:
Assign me this issue

souravjain540:
Assigned. Let me know if you need any help. @Saurav-Pant

Saurav-Pant:
sure Thanks for this, when I  start working on this I will take help

Saurav-Pant:
Can you please assist me in finding this file in the codebase? @souravjain540

souravjain540:
@Saurav-Pant when can I expect a PR?

Saurav-Pant:
I apologize for the delay. As a newcomer to Amplication, both as a user and a contributor, I am having difficulty understanding where I should make changes on the website. I kindly request your guidance in identifying the appropriate location for these changes.

Pratyakshshiv:
can i work along side with @Saurav-Pant  on this @souravjain540 ?

souravjain540:
@Saurav-Pant open a support ticket in discord, if you are okay I can assign it to @Pratyakshshiv as well.

Saurav-Pant:
> @Saurav-Pant open a support ticket in discord, if you are okay I can assign it to @Pratyakshshiv as well.

[...]

PR #6667, opened 2023-08-12
fix:implement code formatting on plugin setting window

Tjsingh01996:
Closes: #5428 

implemented Json formatting in plugin settings


libs/util/json/src/lib/util-json.ts
```diff
@@ -5,3 +5,14 @@ export const isValidJSON = (str: string) => {
...
+export const JsonFormatting = (json: string) => {
+  if (Object.prototype.toString.call(json) === "[object Object]") {
+    json = JSON.stringify(json);
+  }
+  const isValid = isValidJSON(json);
+  if (isValid) {
+    return JSON.stringify(JSON.parse(json!), null, "\t");
+  }
+  return json;
+};
```
overbit, line 18: @Tjsingh01996 avoid using json parse and stringify multiple times since it is an heavy load operation. 

```suggestion
export  const JsonFormatting = (json: any) => {
  if (Object.prototype.toString.call(json) === "[object Object]") {
    json = JSON.stringify(json);
  }
  const isValid = isValidJSON(json);
  if (isValid) {
    return JSON.stringify(data, null, 2);
  }
  return json;
};
```

@Tjsingh01996 can you add unit tests for this new method?

overbit (changes requested):

libs/util/json/src/lib/util-json.ts
Tjsingh01996: @overbit 
const jsonParse = (str: string) => {
  try {
    let data = JSON.parse(str)
    return data && data;
  } catch (e) {
    return false;
  }
}

export const JsonFormatting = (json: any) => {
  let parsedJson;
  if (Object.prototype.toString.call(json) === "[object Object]") {
    json = JSON.stringify(json);
  }
  parsedJson = jsonParse(json);
  if (parsedJson) {
    return JSON.stringify(parsedJson, null, 2);
  }
  return json;
};

overbit: Sorry @Tjsingh01996 I don't understand your comment.  IsValidJson and jsonParse use the same JSON.parse. How is this different from the first implementation?

Tjsingh01996: can you give me suggestion how i can do this 
isValid was validating the json and return. boolean 
but in jsonParse function. i am returning  parsed data if it validate 
in this way we can use json.parse single time

Tjsingh01996: export const isValidJSON = (str: string) => {
  try {
    return JSON.parse(str) && !!str;
  } catch (e) {
    return false;
  }
};

Tjsingh01996: @overbit 
export const JsonFormatting = (data: any) => {
  if (typeof data !== "string") {
    return JSON.stringify(data, null, 2);
  }
  try {
    let object = JSON.parse(data);
    return JSON.stringify(object, null, 2);
  } catch (e) {
    return data
  }
};

Tjsingh01996: can we do like this ?
this code also support the string if we pass in the fuction

overbit (approved):

overbit:
@Tjsingh01996 I've realised that we need to format the editor on change too. Can you do this before we merge? Thank you.

packages/amplication-client/src/Plugins/InstalledPluginSettings.tsx
```diff
@@ -186,7 +186,9 @@ const InstalledPluginSettings: React.FC<Props> = ({
           </div>
           <HorizontalRule />
           <CodeEditor
-            defaultValue={pluginInstallation?.PluginInstallation.settings}
+            defaultValue={JsonFormatting(
+              pluginInstallation?.PluginInstallation.settings
+            )}
             resetKey={resetKey}
             onChange={onEditorChange}
```
overbit, line 193: Let's format on change too.

overbit (changes requested):

packages/amplication-client/src/Plugins/InstalledPluginSettings.tsx
Tjsingh01996: sure

Tjsingh01996: @overbit i have commit change can you review now

overbit (approved):

Issue #6676, opened 2023-08-14
DB Schema Import - id field issues


abrl91:
1. in the imported Prisma Schema, there are some cases when the model doesn't have an id field - we need to add an id field 
2. in the imported Prisma Schema, there are some cases when the id field (the primary key) is also the FK - we should not support it and throw an error

PR #6677, opened 2023-08-14
Feat/handling issues with id field

abrl91:
Close: #6676

### Summary
🧪🔧🚫

This pull request improves the prisma schema parser service to handle id fields consistently and robustly. It ensures that every model has a `String` id field, and rejects schemas with id fields that are foreign keys. It also updates the test cases accordingly.

> _Prisma schema parse_
> _`id` field must be unique_
> _Autumn leaves fall fast_

### Walkthrough
*  Add `prepareIdField` operation to ensure every model has an id field compatible with Amplication ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-101d4943af6514d1fdfefcb8ee7be3d2c6e0f28cb7621a0d41bc39a06d0f7bd3R743-R761))
  - Throw an error if the id field is also a foreign key field of a relation ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-101d4943af6514d1fdfefcb8ee7be3d2c6e0f28cb7621a0d41bc39a06d0f7bd3R767-R772))
  - Add a test case for this operation ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-15585c73ba81eb439d9e2cfcec6fb58e527c258cb4f9c064eb971f971d890b3aL666-R793))
* Remove code that adds an id field in `prepareModelIdAttribute` operation ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-101d4943af6514d1fdfefcb8ee7be3d2c6e0f28cb7621a0d41bc39a06d0f7bd3L610-R610))
  - Update the comment to reflect the new logic ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-101d4943af6514d1fdfefcb8ee7be3d2c6e0f28cb7621a0d41bc39a06d0f7bd3L577-R577))
  - Remove a test case that is no longer relevant ([link](https://github.com/amplication/amplication/pull/6677/files?diff=unified&w=0#diff-15585c73ba81eb439d9e2cfcec6fb58e527c258cb4f9c064eb971f971d890b3aL181-L292))

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.spec.ts
+++ b/packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.spec.ts
@@ -565,7 +565,7 @@ describe("prismaSchemaParser", () => {
           
           model Admin {
             createdAt  DateTime @default(now())
-            username   String   @unique @db.VarChar(256)
+            username   String   @db.VarChar(256)
             roles      Json?
           }`;
           const existingEntities: ExistingEntitySelect[] = [];
@@ -603,7 +603,7 @@ describe("prismaSchemaParser", () => {
                   displayName: "Username",
                   dataType: EnumDataType.SingleLineText,
                   required: true,
-                  unique: true,
+                  unique: false,
                   searchable: false,
                   description: "",
                   properties: {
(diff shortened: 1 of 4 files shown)


packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.spec.ts
```diff
@@ -663,8 +551,539 @@ describe("prismaSchemaParser", () => {
...
+                  },
+                  customAttributes: "",
+                },
+              ],
+            },
+          ];
+          expect(result).toEqual(expectedEntitiesWithFields);
+        });
+
+        it("should add a field with the @id attribute if the model doesn't have an id filed but it has a field with @unique attribute that is not named id", async () => {
```
abrl91, line 892: @yuval-hazaz @overbit 
Should we rename the field to `id` and add the `@id` instead?

I did it like that becaue we are adding the id field to every model that doesn't have it. 
But we can also say that we add the `@id` to a field named id, so why not to rename these kinds of fields to id and add the `@id`

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.ts
```diff
@@ -811,6 +825,25 @@ export class PrismaSchemaParserService {
...
+          field.attributes?.some((attr) => attr.name === ID_ATTRIBUTE_NAME) ??
+          false
+      );
+
+      // add the id field if it doesn't exist. The type is the default type for id field in Amplication - String
+      if (!hasIdField) {
+        builder
+          .model(model.name)
+          .field(ID_FIELD_NAME, "String")
+          .attribute(ID_ATTRIBUTE_NAME);
```
overbit, line 840: Can we add a test to cover this scenario?
I'm not sure if this logic would work in case of multiple unique attributes
```ts
datasource db {
            provider = "postgresql"
            url      = env("DB_URL")
          }
          
          generator client {
            provider = "prisma-client-js"
          }
          
          model Example {
            id    Int @id @default(0)
            value Int    @default(123)    
          }
  
          model Test {
            test_id    String @unique @default("mock_id")
            value2     Int  @unique
          }
```

overbit (changes requested):

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.ts
abrl91: Just saying, this is not a new logic, I just moved it

abrl91: [008158c](https://github.com/amplication/amplication/pull/6677/commits/008158ce97605f72b8bfb295be30232083078d19)

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.spec.ts
overbit: Resolved offline :)

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.ts
```diff
@@ -749,12 +742,88 @@ export class PrismaSchemaParserService {
...
+            .attribute(ID_ATTRIBUTE_NAME);
+
+          builder
+            .model(model.name)
+            .field(uniqueFieldAsIdField.name)
+            .then<Field>((field) => {
+              field.name = ID_FIELD_NAME;
+            });
+
+          this.logger.debug("builder", { builder });
```
overbit, line 796: Is this needed?

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.ts
```diff
@@ -749,12 +742,88 @@ export class PrismaSchemaParserService {
...
+          `id field was added to model "${model.name}"`,
+          EnumActionLogLevel.Warning
+        );
+      }
+
+      const uniqueFieldAsIdField = modelFields.find(
+        (field) =>
+          isValidIdFieldType(field.fieldType as string) &&
+          field.attributes?.some((attr) => attr.name === UNIQUE_ATTRIBUTE_NAME)
+      );
```
overbit, line 778: Can this boolean and the one at line 751 be merged into one? I guess we can just use this in both the if statements logic

packages/amplication-server/src/core/prismaSchemaParser/prismaSchemaParser.service.ts
```diff
@@ -749,12 +742,88 @@ export class PrismaSchemaParserService {
...
+
+      // if we have more than one unique field that can be used as id field, we pick the first one and convert it to id field
+      if (!hasIdField && uniqueFieldAsIdField) {
+        if (uniqueFieldAsIdField.name !== ID_FIELD_NAME) {
+          builder
+            .model(model.name)
+            .field(uniqueFieldAsIdField.name)
+            .attribute("map", [`"${uniqueFieldAsIdField.name}"`])
+            .attribute(ID_ATTRIBUTE_NAME);
```
overbit, line 787: In case of situations like this:
```ts
model Test {
  message String @unique
  title   String @unique
  id      String @unique 
}
```

We should add `@id` to the id field and not the message but here the result might be not what we expect

overbit (changes requested):

overbit (approved):

PR #6792, opened 2023-08-30
fix(github service): change no commit error message

morhag90:
Close: #31

## PR Details

Change the error message when there is no new commits

### Summary
🔄🚫📝

Changed the `createPullRequest` method in the `GitProvider` interface and the `GithubService` class to handle the case when no pull request can be created due to no code changes or identical branches. Updated the `git-provider.interface.ts` and `github.service.ts` files accordingly.

> _`createPullRequest`_
> _Returns null if no changes_
> _Autumn leaves falling_

### Walkthrough
* Change the return type of `createPullRequest` method in `GitProvider` interface to allow `null` value ([link](https://github.com/amplication/amplication/pull/6792/files?diff=unified&w=0#diff-ff8aa5b0325c531705eb125cbdd735571a8001b8ca941f9ee0665e627c7fe3c6L54-R54))
* Implement the new interface and handle the error when GitHub API returns 422 status code in `createPullRequest` method in `GithubService` class ([link](https://github.com/amplication/amplication/pull/6792/files?diff=unified&w=0#diff-7825d1b6180f809d60624f1acd8e9ee0f7acfcab132ec45ce7480e4252b1120bL445-R463))

## PR Checklist

Expected result: 

image

### DIFF
--- a/libs/util/git/src/git-provider.interface.ts
+++ b/libs/util/git/src/git-provider.interface.ts
@@ -51,7 +51,7 @@ export interface GitProvider {
   ) => Promise<PullRequest | null>;
   createPullRequest: (
     createPullRequestArgs: GitProviderCreatePullRequestArgs
-  ) => Promise<PullRequest>;
+  ) => Promise<PullRequest | null>;
   getBranch: (args: GetBranchArgs) => Promise<Branch | null>;
   createBranch: (args: CreateBranchArgs) => Promise<Branch>;
   getCloneUrl: (args: CloneUrlArgs) => Promise<string>;
--- a/libs/util/git/src/providers/github/github.service.ts
+++ b/libs/util/git/src/providers/github/github.service.ts
@@ -442,16 +442,25 @@ export class GithubService implements GitProvider {
     pullRequestTitle,
     pullRequestBody,
     baseBranchName,
-  }: GitProviderCreatePullRequestArgs): Promise<PullRequest> {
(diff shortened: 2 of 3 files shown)


libs/util/git/src/git-provider.interface.ts
```diff
@@ -51,7 +51,7 @@ export interface GitProvider {
   ) => Promise<PullRequest | null>;
   createPullRequest: (
     createPullRequestArgs: GitProviderCreatePullRequestArgs
-  ) => Promise<PullRequest>;
+  ) => Promise<PullRequest | null>;
```
abrl91, line 54: Please update all providers with the new returned type of this function

libs/util/git/src/providers/github/github.service.ts
```diff
@@ -442,16 +442,25 @@ export class GithubService implements GitProvider {
...
+        title: pullRequestTitle,
+        body: pullRequestBody,
+        head: branchName,
+        base: baseBranchName,
+      });
+      return { url: pullRequest.html_url, number: pullRequest.number };
+    } catch (error) {
+      if (error.status === 422) {
+        throw new Error(
+          `Hey there! Looks like your code hasn't changed since the last build. We skipped creating a new pull request to keep things tidy.`
```
abrl91, line 459: Like!

abrl91 (approved):

Issue #6771, opened 2023-08-29
Bug Report: After setting a topic of a message broker within a service to receive or send you cant set other topics that belong to the message broker


GreenMachine01:

Sheersendu:
@GreenMachine01 I would like to work on this issue can it be kindly assigned to me please?

GreenMachine01:
@Sheersendu assigned to you. We are interested in releasing the fix for this one as soon as possible. If you are unable to solve it soon please let me know.

Sheersendu:
HI @GreenMachine01 , I followed your issue replication steps but I didn't face the bug : I'm able to set the second topic just after setting the first topic (only once). Is there any other way to replicate this issue?

GreenMachine01:
@Sheersendu Ill try to recreate it on my side and update you.

Sheersendu:
@GreenMachine01  yup i tried once again I got the issue, checking why the reason why.

GreenMachine01:
@Sheersendu It just happened to me on a new service.
select the first topic to be 'send'

Sheersendu:
Even sometimes the first topic is even not getting set I see, looking into it

> @Sheersendu It just happened to me on a new service. select the first topic to be 'send'

Sheersendu:
Hey @GreenMachine01 , I have found a working solution but I'm stuck in one place where we need to update the memo used in TopicsList.tsx as the pattern value is changing. But its not possible to update memo, tried to use  a combination of useState along with useMemo still it didn't work, mysteriously some keys stored in memo is getting deleted. Open for further discussion or any possible solution if known. I've have spend 5-6 hours on this and really want to solve this.

GreenMachine01:
@overbit or @abrl91 can you assist?

Sheersendu:
@GreenMachine01 I've created PR, could you please review it?

GreenMachine01:
@Sheersendu thank you! the team will check it out.

Sheersendu:
Hey @GreenMachine01, can you please let me know the status of this issue. Its under review I can see, but it also says changes requested, kindly let me know what should be my follow up steps. Thanks 😄

GreenMachine01:
@Sheersendu thanks for checking in. I will ping the team to see where it is standing.

overbit:
Issue move to ready for QA

morhag90:
Doesn't pass QA.
Now have an issue with the first topic. 
The screen recording is attached. 

[...]

PR #6794, opened 2023-08-31
fix(client):  Message Broker component selection issue

Sheersendu:
Close: #6771 

## PR Details

1. Updated SelectedPatternType with newly selected pattern as by default it was not getting updated instantly.


packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
overbit, line 1: @Sheersendu  try to see if there is an easier option. 
i.e. messagePatternsByTopicId useMemo depends on `messagePatterns`, so it might be enough to update that object

overbit (changes requested):

packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
Sheersendu: Hi @overbit , I agree with you that messagePatternsByTopicId useMemo depends on `messagePatterns`, hence any changes to `messagePatterns` should automatically recompute changes for messagePatternsByTopicId. But I would like to draw your attention towards this [section](https://github.com/amplication/amplication/pull/6794/files#:~:text=%7Btopic%7D-,selectedPatternType%3D%7B,%7D,-onMessagePatternTypeChange%3D%7B)  which is responsible for rendering the updated pattern type, which on debugging I have found out that the value we pass is always `EnumMessagePatternConnectionOptions.None` although `messagePatterns` was being updated, hence not displaying the updated result on the fly. Hence my approach was to recompute selected pattern type before passing at as prop for rendering. 

I'm open for further discussion and improvements😄

overbit: Good point. I'd need some help from our wizard @barshimi
What do you think about this change?

Sheersendu: Hi @overbit, do we have any updates? Kindly let me knows required changes if any 😄

GreenMachine01:
@barshimi @overbit What's the status with this one?

packages/amplication-client/src/ServiceConnections/topics/TopicsList.tsx
overbit: As we got now answer, I'll approve this PR

overbit (approved):

Issue #6871, opened 2023-09-07
As a Maintainer - I want to create amplication release with only changes relevant to the type of release


overbit:
### Feature description

When generating Release notes, I'd like to the PRs that are affecting:
- all applications / libraries apart from `data-service-generator` when creating generic releases
- only  `data-service-generator` and the libraries that  `data-service-generator` depends on, when creating DSG Releases

### Tech notes

Have a look to https://github.com/terraform-aws-modules/terraform-aws-iam/blob/da36c73d8b899040ae37f7df5fc59f3e65ef8af5/.github/workflows/pr-title.yml for PR title standardization

overbit:
#6720

PR #6984, opened 2023-09-22
chore: add pr labeler workflow that will automatically label PRs based on affected projects in the monorepo

overbit:
Part of: #6871

### Summary
🏷️🚀🛠️

Add a workflow file for the Pull Request Labeler action. The workflow labels pull requests based on the affected apps and libs by the changes and the Nx project structure.

> _Pull requests get labels_
> _`Nx affected` guides the action_
> _Autumn leaves fall fast_

### Walkthrough
*  Add a workflow file for the Pull Request Labeler action ([link](https://github.com/amplication/amplication/pull/6984/files?diff=unified&w=0#diff-eadf2c6de99f6632a36af544d29eee58f6d868620bb798c56284cf43b0ee8c60L1-R80))

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,0 +1,82 @@
+name: Pull Request Labeler
+# This workflows works in conjunction with the release.production.yml workflow to automate release notes based on labels.
+
+on:
+  pull_request:
+
+env:
+  NX_CLOUD_ACCESS_TOKEN: ${{ secrets.NX_CLOUD_ACCESS_TOKEN }}
+
+jobs:
+  nx:
+    name: Nx
+    uses: ./.github/workflows/nx.template.yml
+    with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,93 @@
+name: Pull Request Labeler
+# This workflows works in conjunction with the release.production.yml workflow to automate release notes based on labels.
+
+on:
+  pull_request:
```
levivannoort, line 5: doesn't this require additional configuration like `on close`, `on synchronize` et cetera?

overbit: > By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened

They should be enough

mulygottlieb (approved):

levivannoort (approved):

Issue #7151, opened 2023-10-04
🐛 Bug Report: GitHub icon used when not connected to a git provider instead of the generic Git icon


GreenMachine01:
------------------------------------------------------------------------------------------------------------------------------------
**Important Note for Developers:**

Please ensure that any fixes related to this issue are made against the `Next` branch. If you have any questions or need further clarification, please reach out.

Thank you for your attention to this matter.

------------------------------------------------------------------------------------------------------------------------------------

**Steps to Reproduce:**

1. Create a new project and make sure it is not connected to a Git provider
2. Create a new service and inherit the Git provider from the project settings
3. Go to the Project page to view all your services

**Actual:**
The new service shows it is not connected to a Git provider but it uses the GitHub icon

![Image](https://github.com/amplication/amplication/assets/112330500/530bf92e-792e-425f-8e04-2d4beaecfb16)

**Expected:**
The new service shows it is not connected to a Git provider and it uses the generic Git icon

![Image](https://github.com/amplication/amplication/assets/112330500/8b84b9aa-6a62-4168-80a9-9a8982033da7)

GreenMachine01:
@souravjain540 - FYI

xonx4l:
I made a pr please check .Thanks.

trishanu-init:
@GreenMachine01 can you assign me this issue

GreenMachine01:
@souravjain540 - FYI

ankushbhardwaj408:
Hi @souravjain540  I would like to work on this.

yuval-hazaz:
I think the GitHub Icon make more sense to most users and we should leave it

GreenMachine01:
@xonx4l I assigned to you as you requested first and provided a PR. the team will check it.

xonx4l:
THANKS @GreenMachine01 .

mulygottlieb:
> I think the GitHub Icon make more sense to most users and we should leave it

Closing this ticket based on this comment

PR #7163, opened 2023-10-05
fix(client): GitHub icon used when not connected to a git provider instead of the generic Git icon.

xonx4l:
issue-:#7151


overbit (approved):

mulygottlieb:
Closing this PR since:

1. There was no response to our comment on this PR
2. The original ticket should be canceled as per @yuval-hazaz's comment on it that we should not make this change anyway

Issue #7127, opened 2023-10-03
🐛 Bug Report: Entity permissions page is anchor to the footer


overbit:
### What happened?

https://jam.dev/c/47e3a372-b066-43fa-8588-6b50a983e0df

### What you expected to happen

Permissions should be shown from top to bottom

### How to reproduce

Increase the browser height when on that page

AyushDhimanIITR:
Can I work on this issue?

vedant-z:
@overbit Please assign me this issue.

overbit:
Assigned to @AyushDhimanIITR as per priority. Please let us know if you cannot work on it.

vedant-z:
@overbit @AyushDhimanIITR is is already going to be assigned to one issue. Please let me work on this.

overbit:
Then reassigning this to you.

PR #7232, opened 2023-10-12
Update EntityPermissionAction.scss

vedant-z:
Issue #7127


vedant-z:
@overbit If this one is correct then could you please add Hacktoberfest label

abrl91 (changes requested):
@vedant-z I implemented your change and the outcome is not what we want. The gap between the title and the items is too big, and it seems like the panel is half-opened

Screenshot 2023-10-17 at 10 16 52

mulygottlieb:
Closing this PR as there is no response to our comment on it

Issue #7275, opened 2023-10-17
🐛 Bug Report: Empty state image Misalignment on Commits Page


alexbass86:
### What happened?

On the "commits" page, the image associated with the empty state (no elements) is not correctly centered. 
It should be centered in the middle, but currently, it is aligned to the top.

### What you expected to happen

The image should be vertically centered in the middle of the page when there are no elements.

### How to reproduce

1. Navigate to the "commits" page.
2. Ensure there are no elements displayed.
3. Observe the misaligned image.

**Bad**
![image](https://github.com/amplication/amplication/assets/96179897/b3db804c-93ab-4463-9773-f02c5457031e)

**Good**
![Entities - List](https://github.com/amplication/amplication/assets/96179897/8340338c-b9cb-42e2-9924-9dc0aef918d8)

**Figma**
https://www.figma.com/file/MQ0awwJRdlFYlHhj25ByEI/Community-Task-2023-24?type=design&node-id=2318%3A637&mode=design&t=S0ND8zrjH9XYAQcn-1

### Amplication version

1.10.5

GreenMachine01:
@alexbass86 please add expected example.

alexbass86:
> @alexbass86 please add expected example.

done

sjcode99:
Hey @GreenMachine01 , would like to work on this issue. assign this to me please.

GreenMachine01:
@sjcode99 Assigned to you. Good luck!

Prashant04627:
Can I work on this issue?

GreenMachine01:
@Prashant04627 the item was already assigned. if it becomes available Ill check with you.

sjcode99:
Hey @GreenMachine01, I can see the PR is still not merged. Is there anything that requires my attention?

GreenMachine01:
@mulygottlieb - FYI

PR #7284, opened 2023-10-18
fix: empty state image misalignment on commits page

sjcode99:
Close: #7275 

## PR Details

This PR fixes the empty state image misalignment issue on the commits page.

### DIFF
--- a/packages/amplication-client/src/Components/EmptyState.scss
+++ b/packages/amplication-client/src/Components/EmptyState.scss
@@ -9,6 +9,7 @@ $svg-img-height: 200px;
   flex-direction: column;
   justify-content: center;
   align-items: center;
+  height: 100%;
 
   p {
     @include regular-12;


packages/amplication-client/src/Components/EmptyState.scss
```diff
@@ -9,6 +9,7 @@ $svg-img-height: 200px;
   flex-direction: column;
   justify-content: center;
   align-items: center;
+  height: 100%;
```
yuval-hazaz, line 12: @sjcode99, thanks for your PR.
Changing the style of the component may impact the appearance in all places where the component is being used. 
Did you test it on other pages besides the commits page? 
Did you check how the alignment is achieved in other pages (without the addition of height:100%)? I see that the component had flex:1 - I assume this is the property that makes it happen in other pages

yuval-hazaz (changes requested):

mulygottlieb:
Closing this PR since there is no response to our comment above

PR #7328, opened 2023-10-26
refactor(build-manager): dsg performance - use tar to copy the generated files

abrl91:
Close: #7327

### Summary
🛠️🗜️📦

Refactor build-runner service to improve file handling and artifact size. Use `fsPromises` for file operations and `tar-stream` for compression and decompression of the job directory.

> _Oh, we're the build-runner crew and we know what to do_
> _We use `fsPromises` to read and write_
> _We compress and decompress with the `tar-stream` finesse_
> _And we ship the artifacts in the night_

### Walkthrough
*  Replace `fs` with `fsPromises` for consistency and to avoid mixing callback and promise-based APIs ([link](https://github.com/amplication/amplication/pull/7328/files?diff=unified&w=0#diff-a2d46b2df4132a6474b67758980206d93609e2deb576bab82bd20c13553f470bL28-R34), [link](https://github.com/amplication/amplication/pull/7328/files?diff=unified&w=0#diff-a2d46b2df4132a6474b67758980206d93609e2deb576bab82bd20c13553f470bL37-R41))
*  Add imports for `fs`, `AmplicationLogger`, and `tar` modules, and inject logger service in constructor ([link](https://github.com/amplication/amplication/pull/7328/files?diff=unified&w=0#diff-a2d46b2df4132a6474b67758980206d93609e2deb576bab82bd20c13553f470bL4-R18))
*  Add `compressPath` variable to store the path of the directory where the tar file will be created ([link](https://github.com/amplication/amplication/pull/7328/files?diff=unified&w=0#diff-a2d46b2df4132a6474b67758980206d93609e2deb576bab82bd20c13553f470bR63-R67))
*  Replace copying job directory to artifact directory with creating and extracting a tar file from the job directory, using `tar-stream` module and helper methods `packDirectory`, `tarDirectory`, and `untarDirectory` ([link](https://github.com/amplication/amplication/pull/7328/files?diff=unified&w=0#diff-a2d46b2df4132a6474b67758980206d93609e2deb576bab82bd20c13553f470bL65-R137))

## PR Checklist


packages/amplication-build-manager/src/build-runner/build-runner.service.ts
```diff
@@ -56,12 +60,79 @@ export class BuildRunnerService {
...
+    this.logger.debug(`Compressing ${jobPath} to ${tarFile}`);
+    await this.tarDirectory(jobPath, tarFile);
+    this.logger.debug(`Created tar file ${tarFile}`);
+
+    await this.untarDirectory(tarFile, artifactPath);
+    this.logger.debug(`Extracted tar file ${tarFile} to ${artifactPath}`);
+  }
+
+  async packDirectory(pack, source, relativePath = "") {
+    const files = fs.readdirSync(source);
```
overbit, line 85: Try to avoid using sync `fs` operation that will affect the rest of the runtime

packages/amplication-build-manager/src/build-runner/build-runner.service.ts
```diff
@@ -56,12 +60,79 @@ export class BuildRunnerService {
...
+    for (const file of files) {
+      const filePath = join(source, file);
+      const relFilePath = join(relativePath, file);
+      const stat = fs.statSync(filePath);
```
overbit, line 89: Try to avoid using sync fs operation that will affect the rest of the runtime

packages/amplication-build-manager/src/build-runner/build-runner.service.ts
```diff
@@ -56,12 +60,79 @@ export class BuildRunnerService {
...
+      if (stat.isFile()) {
+        pack.entry({ name: relFilePath }, fs.readFileSync(filePath));
```
overbit, line 91: Try to avoid using sync fs operation that will affect the rest of the runtime

overbit (changes requested):

overbit:
@mulygottlieb not sure we want to merge this. What do you think?

abrl91:
> @mulygottlieb not sure we want to merge this. What do you think?

I really think we can close it and wait for the dsg job split. Then decide if we want/need another performance improvement, but this will not affect the performance at all

mulygottlieb:
Okay @abrl91 @overbit @yuval-hazaz  - After some discussions between us on this matter of the tar/zip and the lack of performance improvement we see, I am closing this PR for now :(

I think the correct solution we should consider is to let the DSG write the modules directly into the target folder in EFS (instead of copying it over there), while making sure that the mount/FS permissions that the DSG gets are strictly limited to that folder so it cannot go outside and "see" other build artifacts. This will, of course, require some DevOps work and some spike, and we can decide if/when we want to tackle this effort. For now, I'm just closing this PR as it didn't prove any improvement on the one hand and might introduce new issues on the other hand.

Issue #6905, opened 2023-09-13
As User - I want to use gRPC to communicate between my services


GreenMachine01:
We should start with a manual (custom code) POC - create two simple services with Amplication and see how to create a gRPC endpoint in one that is called from the other. We need to understand what it takes on both sides and what should be shared (e.g. the protobuf or another definition of the input/output of the endpoint, etc.). After we have this POC let's present it and discuss how we envision implementation this inside Amplication:
- Create another endpoint option (REST / GQL / gRPC)?
- Add automatic connection between services in the same project like with Kafka?

Useful reading:
- https://docs.nestjs.com/microservices/grpc
- https://grpc.io/docs/what-is-grpc/introduction/

**Documentation** - https://github.com/amplication/docs/issues/404

mulygottlieb:
Passed QA:


![Image](https://github.com/amplication/amplication/assets/1912933/2906b81b-6118-4065-9783-93d94282e73a)

PR #7329, opened 2023-10-26
fix(dsg): update logic changed in #7321 to be back-compatible

overbit:
Part of https://github.com/amplication/amplication/issues/6905

## PR Details

Refactor code changed in #7321 to be back-compatible and avoid breaking existing plugins
## PR Checklist


morhag90 (approved):

libs/util/code-gen-utils/src/lib/files.ts
```diff
@@ -4,21 +4,23 @@ import { memoize } from "lodash";
...
-  if (path.endsWith(".yml") || path.endsWith(".yaml")) {
-    return format(code, { parser: "yaml" });
-  }
-  if (path.endsWith(".md")) {
-    return format(code, { parser: "markdown" });
-  }
-  if (path.endsWith(".graphql")) {
-    return format(code, { parser: "graphql" });
+export const formatCode = (code: string, path?: string): string => {
+  if (path) {
```
mulygottlieb, line 8: This is partially backward compatible since if no path was given, it will not format at all, compared to the original code.
Perhaps we can , if no path was specified, try to format the code as "typescript" and catch any exception and return the give code as-is on exception? Since most modules are ts/tsx, this will ensure formatting on at least most files. WDYT?

mulygottlieb (approved):

PR #7332, opened 2023-10-26
test(dsg): fix flacky e2e tests

overbit:
Close: #7334

## PR Details

This PR consists of 3 changes:

1. This PR introduce a dynamic wait time before starting e2e tests  based on the actual state of the docker-compose stack instead of a static timeout. 
  With this change, the test run time will also be reduce as it won't need to always wait for 30 seconds before running the actual tests. 🚀
2. Update docker-compose to v2

3. ⚠️ During the investigation, I've noticed that another reason for these e2e tests to fail was the following:

 image
 In order to fix this, in the most simple way, this PR adds a restart policy to the docker-compose stack `server` service to avoid the situation were server failed to start, due to race condition with the migration service, and stayed dead.
With the new policy, if `server` starts and fails due to migrations not being completed, it will automatically be restarted by the docker engine.

### DIFF
--- a/packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
+++ b/packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
@@ -1,7 +1,7 @@
 import * as os from "os";
 import * as fs from "fs";
 import * as path from "path";
-import * as compose from "docker-compose";
+import { v2 as compose } from "docker-compose";
 import getPort from "get-port";
 import sleep from "sleep-promise";
 import fetch from "cross-fetch";
@@ -170,8 +170,6 @@ describe("Data Service Generator", () => {
             // // See: https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/
             PORT: String(port),
             DB_PORT: String(dbPort),
-            COMPOSE_DOCKER_CLI_BUILD: "1",
-            DOCKER_BUILDKIT: "1",
           },
         };
         logger.debug("dockerComposeOptions", { dockerComposeDir });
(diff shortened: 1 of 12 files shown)


packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
```diff
@@ -193,8 +192,54 @@ describe("Data Service Generator", () => {
             logger.error(err.message, err);
           });
 
+        // Wait for the server to be ready and for database migration before running tests.
+        // TODO replace with a better docker-compose solution that waits for the migraton to be completd before starting the server
+        logger.info("Waiting for db migration to be ...");
```
mulygottlieb, line 197: "to be ? ... " :) Did you mean "to be completed..." ?

packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
```diff
@@ -193,8 +192,54 @@ describe("Data Service Generator", () => {
...
+          );
+          if (migrateContainer.state.indexOf("Exited (0)") !== -1) {
+            migrationCompleted = true;
+            logger.info("migration completed!");
+            break;
+          }
+          await sleep(1000);
+        } while (
+          !migrationCompleted ||
+          startTime + SERVER_START_TIMEOUT < Date.now()
```
mulygottlieb, line 218: Do you want to use the same timeout (SERVER_START_TIMEOUT) for both the migration and the server start?

mulygottlieb (approved):

packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
overbit: We can make them different but I don't seen any reasons for it at the moment

packages/data-service-generator/tests/e2e/data-service-generator.e2e.spec.ts
overbit: fixed

Issue #7337, opened 2023-10-26
🐛 Bug Report: Local DSG run with the latest published code-gen-types package


yuval-hazaz:
### What happened?

When running the local-dsg service, it should run with the local unpublished changes of the code-gen-types to allow testing local changes without publishing. 

### What you expected to happen

Run local-dsg with unpublished changes 

### How to reproduce

make changes on code-gen-types and run DSG locally 

### Amplication version

1.10.5

PR #7338, opened 2023-10-26
fix(DSG): Run local DSG with unpublished code-gen-types

yuval-hazaz:
Close: #7337 

## PR Details

copilot:all

## PR Checklist

### DIFF
--- a/packages/data-service-generator/Dockerfile
+++ b/packages/data-service-generator/Dockerfile
@@ -14,6 +14,10 @@ RUN npm i --omit-dev
 # Copy unpublished logging library
 COPY ./dist/libs/util/logging node_modules/@amplication/util/logging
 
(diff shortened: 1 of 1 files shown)


overbit:
> @overbit @mulygottlieb please take a look at this change. For the best of my knowledge, this DockerFile is only used locally and the change does not have impact on the production deployment, but I would need you to confirm that before we merge this change

@yuval-hazaz the Dockerfile is used to generate any container image, local and/or staging/production.
This approach was a possible one to avoid publishing the libraries before testing them.
We need to investigate if the latest nx will allow us to fully leverage it to automatically copy all the deps in build time (it would be ok when bundling but dsg is not)

yuval-hazaz:
@overbit following our conversation, please close this PR in case you want to proceed in a different route

Issue #7364, opened 2023-11-01
As a User - I want to be able to start the generated service in HTTPS and/or HTTP mode


levivannoort:
### Feature description

As a User I would like to be able to start my generated service in HTTPS mode as well as / or in HTTP mode. Currently the generated service only supports HTTP.

HTTPS:
```typescript
const httpsOptions = {
  key: fs.readFileSync('./secrets/private-key.pem'),
  cert: fs.readFileSync('./secrets/public-certificate.pem'),
};

const app = await NestFactory.create(AppModule, {
  httpsOptions,
});

await app.listen(3000);
```

HTTP & HTTPS:
```typescript
const httpsOptions = {
  key: fs.readFileSync('./secrets/private-key.pem'),
  cert: fs.readFileSync('./secrets/public-certificate.pem'),
};

const server = express();
const app = await NestFactory.create(
  AppModule,
  new ExpressAdapter(server),
);
await app.init();

const httpServer = http.createServer(server).listen(3000);
const httpsServer = https.createServer(httpsOptions, server).listen(443);
```

Documentation: https://docs.nestjs.com/faq/multiple-servers

---

Suggestion implementation: 
> "I would like to see variable(s) in the env file to instantiate one or the other.  we would probably only instantiate the https version."

### Use case

Currently the generated service support starting the service in HTTP mode, which is suitable for many usecases, but sometimes it is desired to have the TLS termination be at the application level as well as or instead of at the loadbalancer level.

xonx4l:
Is this looks good to you @levivannoort . if yes then I can work further on this otherwise you can direct me towards the right side .

overbit:
@PazYanoverr as we decided to implement this feature as a plugin, we will need an icon in order to release it. Can you help with that?

overbit:
## Test result: ✅

PR #7369, opened 2023-11-01
ADDED:  A generated service in HTTPS and/or HTTP mode.

xonx4l:
issue-:#7364

Added a generated service in HTTPS mode.


yuval-hazaz:
@overbit @mulygottlieb please have a look

packages/amplication-server/src/main.ts
```diff
@@ -52,7 +54,16 @@ async function bootstrap() {
...
-  await app.listen(process.env.PORT || 3000);
+  const appMode = process.env.APP_MODE || "http";
+  if (appMode === "https") {
+    const httpsServer = https.createServer(httpsOptions, app.getHttpServer());
+    httpsServer.listen(443);
+    console.log("HTTPS server is running on port 443");
+  } else {
+    const port = process.env.PORT || 3000;
+    app.listen(port);
+    console.log(`HTTP server is running on port ${port}`);
```
overbit, line 65: Please use Nestjs logger instead of console.log.

More details here https://docs.nestjs.com/techniques/logger#dependency-injection

packages/amplication-server/src/main.ts
```diff
@@ -52,7 +54,16 @@ async function bootstrap() {
     app.enableCors();
   }
 
-  await app.listen(process.env.PORT || 3000);
+  const appMode = process.env.APP_MODE || "http";
+  if (appMode === "https") {
+    const httpsServer = https.createServer(httpsOptions, app.getHttpServer());
+    httpsServer.listen(443);
+    console.log("HTTPS server is running on port 443");
```
overbit, line 61: Please use Nestjs logger instead of console.log.

More details here https://docs.nestjs.com/techniques/logger#dependency-injection

packages/amplication-server/src/main.ts
```diff
@@ -52,7 +54,16 @@ async function bootstrap() {
     app.enableCors();
   }
 
-  await app.listen(process.env.PORT || 3000);
+  const appMode = process.env.APP_MODE || "http";
```
overbit, line 57: Please add the new expected environment variable to the .env file.

packages/amplication-server/src/main.ts
```diff
@@ -52,7 +54,16 @@ async function bootstrap() {
...
+  if (appMode === "https") {
+    const httpsServer = https.createServer(httpsOptions, app.getHttpServer());
+    httpsServer.listen(443);
```
overbit, line 60: Users might still need to customise the port and as there is a variable for it `process.env.PORT`, the code should use that value, and if missing, default to 443

overbit (changes requested):

overbit:
@xonx4l please maintain the pull request description in line with the template amplication offers

overbit:
@xonx4l please maintain the pull request description in line with the template amplication offers


@yuval-hazaz @mulygottlieb should we offer this as a plugin instead so the default generated code is slimmer? Many users do TLS termination at higher levels like in load balancer or ingress

xonx4l:
@overbit  made some changes as per your suggestions . Thanks .

overbit:
@xonx4l after a discussion internally to the team we decided that having a new plugin that will change the default TLS termination to be at the service level.
Would you be happy to work on such a plugin? The generated code will match what you have here.

xonx4l:
@overbit  yeah  man I will love to build one .

mulygottlieb:
@overbit @xonx4l  - I am closing this PR based on the fact we want to add this as a plugin.
@xonx4l  you are welcome to continue the work as a plugin as per the above. Just please note that the changes you suggested here seem to relate to our services while the requested changes are for the **generated code**

Issue #7433, opened 2023-11-16
As a Dev - I want to clean up areas in the `data-service-generator` package where we use the `context` of the DSG in an inconsistent way


abrl91:
There are files that we don't extract all the properties from the context and use `context.` . For example:
```
 const { entities, roles, clientDirectories } = context;

 await context.logger.info("Creating admin..."); <--- context.logger
 await context.logger.info(`Admin path: ${clientDirectories.baseDirectory}`);
```

xonx4l:
Put an pr @abrl91 if its looks right to you you can go forward with merging or if any difficulties we can discuss.

xonx4l:
Thanks

abrl91:
Internal change. Logs still log as expected. Passed QA ✅ 


![Image](https://github.com/amplication/amplication/assets/39680385/30817534-b86a-4e18-8555-ab4f275fac37)

PR #7445, opened 2023-11-17
fix(dsg): cleaned dsg where the context of dsg was in an inconsistent way

xonx4l:
Close: #7433

## PR Details

Cleaned data-service-generator where the context of  data-service-generator was in an inconsistent way. where I extracted logger property to remove inconsistency.


packages/data-service-generator/src/admin/create-admin.ts
```diff
@@ -39,10 +39,10 @@ export function createAdminModules(): Promise<ModuleMap> {
 
 async function createAdminModulesInternal(): Promise<ModuleMap> {
   const context = DsgContext.getInstance;
```
overbit, line 41: Thanks @xonx4l for the contribution.
Anyway, there are more places in this and other files where we need a bit of cleanup.
My suggestion is to remove this line here (and the other files) and start fixing the code on the go.
```suggestion
```

overbit (changes requested):

yuval-hazaz (approved):

Issue #7451, opened 2023-11-17
Add support for showing prerelease plugins only on specific environments


overbit:
### Feature description

In order to simplify testing of new plugin versions we will need a way to publish prerelease that would only being available in non production environments

To achieve this, we will need to:

- [x] for Production and Staging:  only store plugin version matching the regex 
      `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)`

- [x] for every other environment:  only store plugin version matching the regex 
      `^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

Part 2 of #7441 

### Are you willing to submit PR?

Yes I am willing to submit a PR!

abrl91:
In sandbox:



![Image](https://github.com/amplication/amplication/assets/39680385/5408a371-7335-495c-a04a-eae75767c760)

PR #7452, opened 2023-11-17
feat(plugin-api): add support for hiding prerelease versions

overbit:
Close: #7451 

## PR Details

Add support for storing only stable plugin versions or also prerelease ones based on flag and a manual SQL migration to clean up the data.

The decision to update the service at the ingest time is to avoid decreasing performance of all requests for plugin versions.

⚠️  The SQL data migration must be run manually 
 

## PR Checklist


packages/amplication-plugin-api/src/pluginVersion/npm-plugin-version.service.ts
```diff
@@ -3,13 +3,35 @@ import { Plugin, PluginVersion } from "../../prisma/generated-prisma-client";
...
+  private readonly versionFilterRegex: RegExp;
+
   constructor(
     @Inject(AmplicationLogger) readonly logger: AmplicationLogger,
-    private readonly npmService: NpmService
-  ) {}
+    private readonly npmService: NpmService,
+    private readonly configService: ConfigService
+  ) {
+    const pluginStableVersionRegex =
```
mulygottlieb, line 17: Can you add a comment to describe an example of what each regex will match?

mulygottlieb: I expected a regex to match (*).beta(*)

abrl91 (approved):

packages/amplication-plugin-api/src/pluginVersion/npm-plugin-version.service.ts
abrl91: it can be anything as long as it is a valid semver version, not specifically "beta"

Issue #1313, opened 2021-04-09
Server: User-Role renaming


timhuebener:
**Describe the bug**
When you rename the user-role to something else the DB seed does not work. 

**To Reproduce**
1. Go to Roles
2. Click on User-Role
3. Rename to name: admin, display-name: admin
4. Application won't run

**Expected behavior**
The best option would be to automatically update the role of the initial admin user.
Otherwise at least show a warning.

yuval-hazaz:
thanks @timhuebener

comixtrip:
Hi, i'm willing to take a look at this issue. Pls confirm that there are no other people already work on that so I can start right away.

yuval-hazaz:
@comixtrip i am assigning you. Good luck. We would love to support on discord if needed

itsnoa04:
@comixtrip are you still working on it? if not @yuval-hazaz, I'd love to take a look at it.

xonx4l:
can you reassign to me @yuval-hazaz .

yuval-hazaz:
closing this issue since it is outdated and the code has changed significantly since the ticket was created.
in case this issue remain, we should open a new ticket with details to reproduce

PR #7482, opened 2023-11-23
FIX: added automated updates to User-Role renaming .

xonx4l:
issue-:#1313

Close: #1313

## PR Details

 Added automated updates to User-Role renaming .


yuval-hazaz (changes requested):
@xonx4l you made changes on data-service-genereator-catalog-admin which is not releated to the issue in context.
I will close this PR for now. IF you are interested in working on this, let's discuss it first on the ticket so you have the right guidance before you restart working

Issue #7493, opened 2023-11-23
The "Password" field in "User" entity should be NOT unique by default and disabled


PazYanoverr:
As password entity field type should never be unique across multiple users for security reasons, any field entity of type password should be not unique by default and disabled so users won't make mistakes.

abrl91:
So, this issue happens in other fields, and I have found the cause. 


https://github.com/amplication/amplication/assets/39680385/abf1a697-2b95-45c7-99fa-c247612d905a

mulygottlieb:
Passed QA:


![Image](https://github.com/amplication/amplication/assets/1912933/9d2d62f5-00a1-49de-afed-c61ac1197ace)

PR #7498, opened 2023-11-24
Fix:The "Username" and "Password" fields in "User" entity cannot be changed to be a unique field 

xonx4l:
Close: #7493

## PR Details

 Both USER_NAME_FIELD and USER_PASSWORD_FIELD have the unique property set to true. This change indicates that the values in these fields must be unique across all records in the "User" entity.


morhag90:
Hi @xonx4l , Thank you for your contribution :) 
This is not the right place to fix this issue.
The problem is that the client doesn't call the server when it's a User entity.
Please let me know if you want to give it another try.

morhag90 (changes requested):
Commented in the issue

PR #7535, opened 2023-12-01
Feat/amplication/private-issues/issues/71

overbit:
Close: amplication/private-issues/issues/71

## PR Details

copilot:all

## PR Checklist

### DIFF
--- a/libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
+++ b/libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
@@ -1,69 +1,85 @@
-.amp-button-progress {
-  transition-property: border-color, background-color, color;
-  transition-duration: 0.2s;
-  transition-timing-function: ease-in-out;
+.amp-button {
+  &.amp-button-progress {
+    transition-property: border-color, background-color, color;
+    transition-duration: 0.2s;
+    transition-timing-function: ease-in-out;
 
-  &.amp-button-progress--left {
-    background-color: transparent;
-    border-radius: var(--default-button-radius) 0 0 var(--default-button-radius) !important;
-  }
+    &.amp-button-progress--left {
+      background-color: transparent;
+      border-radius: var(--default-button-radius) 0 0
+        var(--default-button-radius) !important;
+    }
 
-  &.amp-button-progress--right {
-    color: var(--gray-full);
-    background-color: transparent;
-    border-radius: 0 var(--default-button-radius) var(--default-button-radius) 0 !important;
-  }
+    &.amp-button-progress--right {
+      color: var(--gray-full);
+      background-color: transparent;
+      border-radius: 0 var(--default-button-radius) var(--default-button-radius)
+        0 !important;
+    }
 
-  &.amp-button-progress--default {
-    color: var(--gray-base);
+    .amp-button-progress-pie {
+      &--primary {
+        color: var(--theme-turquoise);
+      }
+      &--warning {
+        color: var(--theme-orange);
+      }
+      &--danger {
+        color: var(--theme-red);
+      }
+    }
 
-    &--left {
-      border-color: var(--theme-turquoise);
-      &:hover,
-      &:active,
-      &:focus {
-        color: var(--gray-base);
+    &.amp-button-progress--primary {
+      color: var(--gray-base);
+
+      &--left {
         border-color: var(--theme-turquoise);
-        cursor: default;
+        &:hover,
+        &:active,
+        &:focus {
+          color: var(--gray-base);
+          border-color: var(--theme-turquoise);
+          cursor: default;
+        }
+      }
+      &--right {
+        border-color: var(--theme-turquoise);
+        background-color: var(--theme-turquoise);
       }
     }
-    &--right {
-      border-color: var(--theme-turquoise);
-      background-color: var(--theme-turquoise);
-    }
-  }
-  &.amp-button-progress--warning {
-    &--left {
-      border-color: var(--theme-orange);
-
-      &:hover,
-      &:active,
-      &:focus {
-        color: var(--gray-base);
+    &.amp-button-progress--warning {
+      &--left {
         border-color: var(--theme-orange);
-        cursor: default;
-      }
-    }
-    &--right {
-      border-color: var(--theme-orange);
-      background-color: var(--theme-orange);
-    }
-  }
-  &.amp-button-progress--danger {
-    &--left {
-      border-color: var(--theme-red);
 
-      &:hover,
-      &:active,
-      &:focus {
-        color: var(--gray-base);
-        border-color: var(--theme-red);
-        cursor: default;
+        &:hover,
+        &:active,
+        &:focus {
+          color: var(--gray-base);
+          border-color: var(--theme-orange);
+          cursor: default;
+        }
+      }
+      &--right {
+        border-color: var(--theme-orange);
+        background-color: var(--theme-orange);
       }
     }
-    &--right {
-      border-color: var(--theme-red);
-      background-color: var(--theme-red);
+    &.amp-button-progress--danger {
+      &--left {
+        border-color: var(--theme-red);
+
+        &:hover,
+        &:active,
+        &:focus {
+          color: var(--gray-base);
+          border-color: var(--theme-red);
+          cursor: default;
+        }
+      }
+      &--right {
+        border-color: var(--theme-red);
+        background-color: var(--theme-red);
+      }
     }
   }
 }
--- a/libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
+++ b/libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
@@ -1,85 +1,84 @@
-.amp-button {
-  &.amp-button-progress {
-    transition-property: border-color, background-color, color;
-    transition-duration: 0.2s;
-    transition-timing-function: ease-in-out;
+.amp-button-progress {
+  transition-property: border-color, background-color, color;
+  transition-duration: 0.2s;
+  transition-timing-function: ease-in-out;
 
-    &.amp-button-progress--left {
-      background-color: transparent;
-      border-radius: var(--default-button-radius) 0 0
-        var(--default-button-radius) !important;
+  &.amp-button-progress--left {
+    background-color: transparent;
+    border-radius: var(--default-button-radius) 0 0 var(--default-button-radius) !important;
+  }
+
+  &.amp-button-progress--right {
+    color: var(--gray-full);
+    background-color: transparent;
+    border-radius: 0 var(--default-button-radius) var(--default-button-radius) 0 !important;
+  }
+
+  .amp-button-progress-pie {
+    &--primary {
+      color: var(--theme-turquoise);
     }
-
-    &.amp-button-progress--right {
-      color: var(--gray-full);
-      background-color: transparent;
-      border-radius: 0 var(--default-button-radius) var(--default-button-radius)
-        0 !important;
+    &--warning {
+      color: var(--theme-orange);
     }
-
-    .amp-button-progress-pie {
-      &--primary {
-        color: var(--theme-turquoise);
-      }
-      &--warning {
-        color: var(--theme-orange);
-      }
-      &--danger {
-        color: var(--theme-red);
-      }
+    &--danger {
+      color: var(--theme-red);
     }
+  }
 
-    &.amp-button-progress--primary {
-      color: var(--gray-base);
+  &.amp-button-progress--primary {
+    color: var(--gray-base);
 
-      &--left {
+    &--left {
+      border-color: var(--theme-turquoise);
+
+      &:hover,
+      &:active,
+      &:focus {
+        color: var(--gray-base);
         border-color: var(--theme-turquoise);
-        &:hover,
-        &:active,
-        &:focus {
-          color: var(--gray-base);
-          border-color: var(--theme-turquoise);
-          cursor: default;
-        }
+        cursor: default;
       }
-      &--right {
+    }
+    &--right {
+      &:not(&:hover, &:active, &:focus) {
         border-color: var(--theme-turquoise);
         background-color: var(--theme-turquoise);
       }
     }
-    &.amp-button-progress--warning {
-      &--left {
-        border-color: var(--theme-orange);
+  }
+  &.amp-button-progress--warning {
+    &--left {
+      border-color: var(--theme-orange);
 
-        &:hover,
-        &:active,
-        &:focus {
-          color: var(--gray-base);
-          border-color: var(--theme-orange);
-          cursor: default;
-        }
-      }
-      &--right {
+      &:hover,
+      &:active,
+      &:focus {
+        color: var(--gray-base);
         border-color: var(--theme-orange);
-        background-color: var(--theme-orange);
+        cursor: default;
       }
     }
-    &.amp-button-progress--danger {
-      &--left {
-        border-color: var(--theme-red);
+    &--right {
+      border-color: var(--theme-orange);
+      background-color: var(--theme-orange);
+    }
+  }
+  &.amp-button-progress--danger {
+    &--left {
+      border-color: var(--theme-red);
 
-        &:hover,
-        &:active,
-        &:focus {
-          color: var(--gray-base);
-          border-color: var(--theme-red);
-          cursor: default;
-        }
-      }
-      &--right {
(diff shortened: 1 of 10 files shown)


libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const {
+    yellowColorThreshold,
+    redColorThreshold,
+    onClick,
+    leftValue,
+    className,
+    children,
+    rest,
+  } = props;
+  let { progress } = props;
```
barshimi, line 80: why don't you take it from props ? under the props on line 70

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  if (progress < 0) {
```
barshimi, line 82: does progress can be negative number ? you can define the type and set default value
you can validate the prop by using ButtonProgress.propsType & ButtonProgress. defaultProps.
this will allow you to set default and pass valid number like:
progress: PropTypes.oneOf([...(new Array(10))].map((_, i) => i + 1))

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+    progress = 0;
+  }
+  if (progress > 100) {
+    progress = 100;
+  }
+
+  let buttonProgressStyle: EnumButtonProgressStyle;
```
barshimi, line 89: this logic should be store in a useMemo function so it wont do the calculation each render. also, it will be easy use ternary and default to prevent if/else soup

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+      <Button
+        className={classNames(
+          "amp-button-progress",
+          className,
+          `amp-button-progress--left`,
+          `amp-button-progress--${buttonProgressStyle}--left`
+        )}
+        buttonStyle={EnumButtonStyle.Outline}
+        onClick={onClick}
+        {...rest}
```
barshimi, line 114: this is could run over all of your classes and the rest of your props

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+      >
+        <FlexItem
+          direction={EnumFlexDirection.Row}
+          itemsAlign={EnumItemsAlign.Center}
+          gap={EnumGapSize.Small}
+        >
+          <Pie
+            value={progress}
+            size="14px"
```
barshimi, line 123: if this is a fix size it should be default to the component rather passed by the parent

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+      </Button>
+      <Button
+        className={classNames(
+          "amp-button-progress",
+          className,
+          "amp-button-progress--right",
+          `amp-button-progress--${buttonProgressStyle}--right`
+        )}
+        onClick={onClick}
+        {...rest}
```
barshimi, line 139: also here

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+        sx={{
+          color: "inherit",
+          position: "absolute",
+        }}
+      />
+    </Box>
+  );
+};
+
+export const ButtonProgress = (props: Props) => {
```
barshimi, line 70: is this a design system component or just a regular one. it seems to me that it is not generic enough to be DS component

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
 
   const [versionAlert, setVersionAlert] = useState(false);
 
+  const [upgradeButtonData, setUpgradeButtonData] = useState<{
```
barshimi, line 93: all of this code should be a hook and not part of the header component

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
...
+    trialDaysLeft?: number;
+    trialLeftProgress?: number;
+    showUpgradeTrialButton: boolean;
+    showUpgradeDefaultButton: boolean;
+  }>({ showUpgradeTrialButton: false, showUpgradeDefaultButton: true });
+
+  const getCurrentSubscription = useCallback(async () => {
+    if (currentWorkspace) {
+      await stigg.setCustomerId(currentWorkspace.id);
```
barshimi, line 102: dont we have this set on our stigg hook ?
if not it should happen on any signin and not on specific check. also it should be part of the context/stigg hook

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
...
+      const [subscription] = await stigg.getActiveSubscriptions();
```
barshimi, line 103: also this

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
...
   const [versionAlert, setVersionAlert] = useState(false);
 
+  const [upgradeButtonData, setUpgradeButtonData] = useState<{
+    trialDaysLeft?: number;
+    trialLeftProgress?: number;
+    showUpgradeTrialButton: boolean;
+    showUpgradeDefaultButton: boolean;
+  }>({ showUpgradeTrialButton: false, showUpgradeDefaultButton: true });
+
+  const getCurrentSubscription = useCallback(async () => {
```
barshimi, line 100: all of this code should be part of stigg hook and pass as a context function/value

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
...
+      const [subscription] = await stigg.getActiveSubscriptions();
+
+      if (!subscription) {
+        console.error("No active subscription found");
+        return;
+      }
+
+      const DAYS_TO_SHOW_VERSION_ALERT_SINCE_END_OF_TRIAL = 14;
+
+      if (
```
barshimi, line 112: please try to use hash map instead of if/else soup

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -82,6 +90,71 @@ const WorkspaceHeader: React.FC = () => {
...
+          trialDaysLeft,
+          trialLeftProgress,
+          showUpgradeTrialButton: true,
+          showUpgradeDefaultButton: false,
+        });
+      }
+    }
+  }, [currentWorkspace]);
+
+  useEffect(() => {
```
barshimi, line 154: this will run every rerender of the header. this is the reason I told you to pass it to the stigg hook and pass it inside the context. by the way, you wont need the state for that because it will happen once

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -187,13 +260,27 @@ const WorkspaceHeader: React.FC = () => {
...
           <div className={`${CLASS_NAME}__links`}>
-            <Button
-              className={`${CLASS_NAME}__upgrade__btn`}
-              buttonStyle={EnumButtonStyle.Outline}
-              onClick={handleUpgradeClick}
-            >
-              Upgrade
-            </Button>
+            {upgradeButtonData.showUpgradeTrialButton && (
+              <ButtonProgress
```
barshimi, line 264: because this button is being used once it is not DS component

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
```diff
@@ -187,13 +260,27 @@ const WorkspaceHeader: React.FC = () => {
...
+                onClick={handleUpgradeClick}
+                progress={upgradeButtonData.trialLeftProgress}
+                leftValue={`${upgradeButtonData.trialDaysLeft} days free trial left`}
+                yellowColorThreshold={50}
+                redColorThreshold={0}
+              >
+                Upgrade
+              </ButtonProgress>
+            )}
+            {upgradeButtonData.showUpgradeDefaultButton && (
```
barshimi, line 275: instead of having this condition I would add this to the upgrade button, as the controller of the button renderer

packages/amplication-client/src/util/BillingPlan.ts
```diff
@@ -0,0 +1,6 @@
+export enum BillingPlan {
```
barshimi, line 1: I think we had it somewhere else, dont we ?

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
```diff
@@ -0,0 +1,84 @@
+.amp-button-progress {
+  transition-property: border-color, background-color, color;
+  transition-duration: 0.2s;
+  transition-timing-function: ease-in-out;
+
+  &.amp-button-progress--left {
+    background-color: transparent;
+    border-radius: var(--default-button-radius) 0 0 var(--default-button-radius) !important;
```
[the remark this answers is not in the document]
overbit: due to the fact that amp-button class was using !important. Removed from there and from here.

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  &.amp-button-progress--right {
+    color: var(--gray-full);
+    background-color: transparent;
+    border-radius: 0 var(--default-button-radius) var(--default-button-radius) 0 !important;
```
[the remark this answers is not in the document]
overbit: due to the fact that amp-button class was using !important. Removed from there and from here.

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
```diff
@@ -0,0 +1,84 @@
...
+  &.amp-button-progress--primary {
+    color: var(--gray-base);
+
+    &--left {
+      border-color: var(--theme-turquoise);
+
+      &:hover,
+      &:active,
+      &:focus {
+        color: var(--gray-base);
```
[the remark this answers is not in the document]
overbit: updated

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.scss
```diff
@@ -0,0 +1,84 @@
...
+        border-color: var(--theme-turquoise);
```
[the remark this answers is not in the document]
overbit: updated

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
overbit: Any reason to prefer an hash map here? Try to understand the benefit as it will be less performant to use hash map for these 3 if cases.
_The hashmap has to do more commands to get you there. Fetch hashcode, calculate distance/position, fetch the array entry, and run an equals operation._

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+import { EnumFormStyle } from "../Form/Form";
+
+export interface Props
+  extends Pick<PrimerButtonProps, "className" | "children" | "onClick"> {
+  /** Progress percentage 0-100 */
+  progress: number;
+  yellowColorThreshold: number;
+  redColorThreshold: number;
+  leftValue: string;
+  rest?: any;
```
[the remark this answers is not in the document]
overbit: removing it as not used

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
overbit: addressed

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
overbit: addressed

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+}
+
+enum EnumButtonProgressStyle {
+  Primary = "primary",
+  Warning = "warning",
+  Danger = "danger",
+}
+
+const Pie = (props: CircularProgressProps) => {
```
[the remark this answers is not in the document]
overbit: addressed

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+  return (
+    <Box
```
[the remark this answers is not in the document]
overbit: Following the MUI guidelines here: https://mui.com/material-ui/react-progress/#customization

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
overbit: It can be reused in different places as per figma design

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
overbit: done

libs/ui/design-system/src/lib/components/ButtonProgress/ButtonProgress.tsx
```diff
@@ -0,0 +1,145 @@
...
+      className={props.className}
+      sx={{
+        position: "relative",
+        display: "flex",
+      }}
+    >
+      <CircularProgress
```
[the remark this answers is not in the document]
overbit: done

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
overbit: What is the benefit in doing so?
Is a pattern that we use or is for performance gain? I'm asking as the current implementation makes at most 1 request to stigg apis even on multiple re-render of the component.

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
overbit: As users will be signed in for days, this check must happen every day they load or move around in the system

packages/amplication-client/src/Workspaces/WorkspaceHeader/WorkspaceHeader.tsx
overbit: Moved to an hook

barshimi (approved):

Issue #7550, opened 2023-12-05
🐛 Bug Report: builds using DSG versions `<v2.2.0` fail 100% of the time


overbit:
New DSG versions that support build split work, but old DSG versions that don't support build split fail without any message

<img width="1671" alt="image" src="https://github.com/amplication/amplication/assets/2861984/d1d0fe92-08d1-4681-8136-11845624bf99">

_Originally posted by @overbit in https://github.com/amplication/amplication/issues/7494#issuecomment-1840670806_

overbit:
<img width="1510" alt="image" src="https://github.com/amplication/amplication/assets/2861984/7b08916e-27ae-4731-a6a9-16f33ad0a501">

levivannoort:
QA:

![image](https://github.com/amplication/amplication/assets/73097785/061fb06a-6f79-425b-b296-7edfe3ec3a7a)

PR #7553, opened 2023-12-06
fix(build-manager): support for old `data-servivce-generator` versions that don't support build split

overbit:
Close: #7550

## PR Details

This pull request fix the issue reported in #7550. The cause being the logic that handle the status reporting only working for builds that were split in multiple jobs. 

## PR Checklist

### DIFF
--- a/packages/amplication-build-manager/src/build-job-handler/build-job-handler.service.ts
+++ b/packages/amplication-build-manager/src/build-job-handler/build-job-handler.service.ts
@@ -10,19 +10,37 @@ import {
 } from "../types";
 import { RedisService } from "../redis/redis.service";
 import { AmplicationLogger } from "@amplication/util/nestjs/logging";
+import { ConfigService } from "@nestjs/config";
+import { Env } from "../env";
+import { CodeGeneratorService } from "../code-generator/code-generator-catalog.service";
 
 type ResourceTuple = [JobBuildId<BuildId>, DSGResourceData];
 @Injectable()
 export class BuildJobsHandlerService {
+  private readonly minDsgVersionToSplitBuild: string;
+
   constructor(
     private readonly redisService: RedisService,
-    private readonly logger: AmplicationLogger
-  ) {}
+    private readonly logger: AmplicationLogger,
+    private readonly configService: ConfigService<Env, true>,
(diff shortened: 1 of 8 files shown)


abrl91 (dismissed):

abrl91 (approved):

packages/amplication-build-manager/src/build-job-handler/build-job-handler.service.ts
```diff
@@ -33,23 +51,29 @@ export class BuildJobsHandlerService {
...
-
-    if (generateAdminUI) {
-      const adminUiDSGResourceData: DSGResourceData =
-        cloneDeep(dsgResourceData);
-      adminUiDSGResourceData.resourceInfo.settings.serverSettings.generateServer =
-        false;
-      const jobBuildId: JobBuildId<BuildId> = `${buildId}-${EnumDomainName.AdminUI}`;
-      await this.setJobStatus(jobBuildId, EnumJobStatus.InProgress);
-      jobs.push([jobBuildId, adminUiDSGResourceData]);
+    if (shouldSplitBuild) {
```
barshimi, line 54: this if/else is redundant. we should treat the build as an array of builds. so it can push one job or split jobs. which means instead of if/else it create a buildId array and run it accordingly

barshimi: I also think the tests should be changed accordingly

overbit: That is indeed the logic you are mentioning as we need to add either one job per domain (server or admin-ui) or a unique one.
Am I missing something @barshimi ?

barshimi: instead of having if/else soup that can be grow in future because of parallel builds. I suggest that you'll create an array of buildIds that will trigger the job itself. so you'll be able to create one job with/without admin or `n` jobs

overbit: Resolved offline. Bottomline, when add a new split job in the logic, we will review and update this.

PR #7579, opened 2023-12-11
feat(server): add logic to prevent changes when entitlements are not in place

abrl91:
Part of: https://github.com/amplication/private-issues/issues/76

## PR Details

Added the server-side logic of blocking creation of: project, resource, entity, field, role after the trial ends

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -263,27 +263,35 @@ export class EntityService {
     workspaceId: string,
     resourceId: string,
     serviceFeature: EnumServiceFeature
-  ) {
-    const canCreate = await this.billingService.isUnderLimitation(
+  ): Promise<boolean> {
+    const feature = await this.billingService.getMeteredEntitlement(
       workspaceId,
-      BillingFeature.Services,
-      resourceId,
-      (usageLimit) =>
-        this.prisma.resource.findMany({
-          where: {
+      BillingFeature.Services
+    );
+
+    if (!feature.usageLimit) {
+      return false;
+    }
+
+    const featuredEntityOrderedByCreationDate =
+      await this.prisma.resource.findMany({
+        where: {
+          deletedAt: null,
+          archived: { not: true },
+          resourceType: EnumResourceType.Service,
+          project: {
+            workspaceId,
             deletedAt: null,
-            archived: { not: true },
-            resourceType: EnumResourceType.Service,
-            project: {
-              workspaceId,
-              deletedAt: null,
-            },
           },
-          orderBy: {
-            createdAt: "asc",
-          },
-          skip: usageLimit,
-        })
+        },
+        orderBy: {
+          createdAt: "asc",
+        },
+        skip: feature.usageLimit,
+      });
+
+    const canCreate = featuredEntityOrderedByCreationDate.some(
+      (featureEntity) => featureEntity.id === resourceId
     );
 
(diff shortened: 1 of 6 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -252,14 +259,63 @@ export class EntityService {
...
+        },
+        skip: feature.usageLimit,
+      });
+
+    const canCreate = featuredEntityOrderedByCreationDate.some(
+      (featureEntity) => featureEntity.id === resourceId
+    );
+
+    if (!canCreate) {
+      throw new AmplicationError(
```
overbit, line 298: Let's use the new `BillingLimitationError` instead of the generic AmplicationError.

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -252,14 +259,63 @@ export class EntityService {
     });
   }
 
+  async checkServiceFeatureUnderLimitation(
```
overbit, line 262: We could try to keep name consistency with the other validation methods like: `validateSubscriptionPlanLimitationsForServiceFeature` 

Ref: https://github.com/amplication/amplication/blob/master/packages/amplication-server/src/core/billing/billing.service.ts#L279C9-L279C56

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -65,6 +65,17 @@ export class ProjectService {
     args: ProjectCreateArgs,
     userId: string
   ): Promise<Project> {
+    const projectEntitlement = await this.billingService.getMeteredEntitlement(
+      args.data.workspace.connect.id,
+      BillingFeature.Projects
+    );
+
+    if (projectEntitlement && !projectEntitlement.hasAccess) {
+      throw new Error(
```
overbit, line 74: Same here, we should keep using the BillingLimitationError for any exception related to limitation errors

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -128,6 +129,17 @@ export class ResourceService {
       );
     }
 
+    const serviceEntitlement = await this.billingService.getMeteredEntitlement(
+      user.workspace.id,
+      BillingFeature.Services
+    );
+
+    if (serviceEntitlement && !serviceEntitlement.hasAccess) {
+      throw new AmplicationError(
```
overbit, line 138: As per the others limitations, let's use the `BillingLimitationError`

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -118,6 +118,7 @@ export class ResourceService {
    * This function should be called from one of the other "Create[ResourceType] functions like CreateService, CreateMessageBroker etc."
    */
   private async createResource(
+    user: User,
```
overbit, line 121: Are you sure that adding this new positional parameter won't break other calls?
Check any other reference of this method to be sure

overbit, line 121: Are you sure that adding this new positional parameter won't break other calls?
Check any other reference of this method to be sure

overbit (changes requested):

packages/amplication-server/src/core/resource/resource.service.ts
abrl91: there are 2 references and I added the user in the right place

PR #7601, opened 2023-12-12
fix(client): pro plan upgrade button logic

overbit:
Close: https://github.com/amplication/private-issues/issues/71

### Summary
🗑️🔄🐛

This pull request refactors and fixes some components and services related to the subscription and billing features. It uses the subscription plan id to determine the upgrade button data, fixes the profile form dialog visibility, and removes the metered entitlement check for the services above entities per service limit feature. These changes support the new trial period feature for the enterprise plan and simplify the billing logic.

> _Sing, O Muse, of the code that the skilled developers wrought_
> _To enhance and refine the subscription plans they sought_
> _They removed the metered check from the `BillingService` class_
> _And freed the users from the errors that would harass_

### Walkthrough
*  Simplify the logic of the `useUpgradeButtonData` hook to use the subscription plan id instead of the status and pricing type ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-97660417b6d1257e4d2e739f831e3b132ba73693ab2478bb20a9df4c41e1e0acL38-R38))
*  Add a case to the `useUpgradeButtonData` hook to handle the enterprise plan with a trial period and show the upgrade trial button ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-97660417b6d1257e4d2e739f831e3b132ba73693ab2478bb20a9df4c41e1e0acL65-R59))
*  Add a case to the `useUpgradeButtonData` hook to handle the enterprise plan without a trial period and hide the upgrade buttons ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-97660417b6d1257e4d2e739f831e3b132ba73693ab2478bb20a9df4c41e1e0acR77-R91))
*  Fix a bug that prevented the profile form dialog from showing up in the `WorkspaceHeader` component by removing the condition on the upgrade button data state ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-7fdb758d72b986ee99ae091fcd9f9b1312ecd063170e9dcbd9d99d5f47812d24L167-R206))
*  Avoid flickering and errors in the `WorkspaceHeader` component by adding the condition on the upgrade button data state to the rendering of the upgrade trial button and the upgrade default button ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-7fdb758d72b986ee99ae091fcd9f9b1312ecd063170e9dcbd9d99d5f47812d24L218-R219), [link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-7fdb758d72b986ee99ae091fcd9f9b1312ecd063170e9dcbd9d99d5f47812d24L227-R349))
*  Disable the billing limitation for the services above entities per service limit feature in the `BillingService` class by removing the code that checks the metered entitlement ([link](https://github.com/amplication/amplication/pull/7601/files?diff=unified&w=0#diff-f9aed5323cc2e69600bb73c5a362dbae00a82615452516017582cb2825fbe7d1L367-L378))

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Workspaces/hooks/useUpgradeButtonData.ts
+++ b/packages/amplication-client/src/Workspaces/hooks/useUpgradeButtonData.ts
@@ -35,16 +35,7 @@ export const useUpgradeButtonData = (
       await stigg.setCustomerId(currentWorkspace.id);
       const [subscription] = await stigg.getActiveSubscriptions();
 
-      if (
-        subscription.status === SubscriptionStatus.Active &&
-        subscription.plan.pricingType !== PricingType.Free
-      ) {
-        setUpgradeButtonData({
-          showUpgradeTrialButton: false,
-          showUpgradeDefaultButton: false,
-          isCompleted: true,
-        });
-      } else if (subscription.plan.id === BillingPlan.Free) {
+      if (subscription.plan.id === BillingPlan.Free) {
         const daysSinceStartOfPlan = Math.abs(
           (Date.now() - new Date(subscription.startDate).getTime()) / ONE_DAY
         );
@@ -62,7 +53,10 @@ export const useUpgradeButtonData = (
           showUpgradeDefaultButton: false,
           isCompleted: true,
         });
-      } else {
+      } else if (
+        subscription.plan.id === BillingPlan.Enterprise &&
+        subscription.trialEndDate
+      ) {
         const trialDaysLeft = Math.round(
           Math.abs((subscription.trialEndDate.getTime() - Date.now()) / ONE_DAY)
         );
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/Workspaces/hooks/useUpgradeButtonData.ts
```diff
@@ -80,6 +74,21 @@ export const useUpgradeButtonData = (
...
+        setUpgradeButtonData({
+          showUpgradeTrialButton: false,
+          showUpgradeDefaultButton: false,
+          isCompleted: true,
+        });
+      } else {
+        setUpgradeButtonData({
+          showUpgradeTrialButton: false,
+          showUpgradeDefaultButton: true,
+          isCompleted: true,
```
abrl91, line 90: Maybe we can add `isCompleted: true` to the initial state and remove the else block

abrl91 (approved):

packages/amplication-client/src/Workspaces/hooks/useUpgradeButtonData.ts
overbit: But we need it to be `isCompleted` false otherwise the initial render won't have the data. am I missing something?

PR #7658, opened 2023-12-19
feat(server): licensed on projects and resources

abrl91:
Close: https://github.com/amplication/private-issues/issues/88

## PR Details

copilot:all

## PR Checklist


packages/amplication-server/src/core/workspace/workspace.service.ts
```diff
@@ -798,6 +798,55 @@ export class WorkspaceService {
     }
   }
 
+  async bulkUpdateWorkspaceProjectsAndResourcesLicensed(
```
overbit, line 801: it would be great to have an idea of the performance of this bulkUpdate. Probably testing this in sandbox and checking the tracing data

packages/amplication-server/src/core/workspace/workspace.service.ts
```diff
@@ -798,6 +798,55 @@ export class WorkspaceService {
...
+    useUserLastActive: boolean
+  ): Promise<boolean> {
+    try {
+      const date = new Date();
+      const userLastActiveQuery = useUserLastActive
+        ? {
+            some: {
+              lastActive: {
+                gte: new Date(date.setDate(date.getDate() - 30)),
```
overbit, line 810: Would it make sense to make this (`30`) parameter customisable as environment variable?

packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -221,4 +221,13 @@ export class WorkspaceResolver {
   ): Promise<Coupon> {
     return this.workspaceService.redeemCoupon(user, args);
   }
+
+  @Mutation(() => Boolean)
+  async bulkUpdateWorkspaceProjectsAndResourcesLicensed(
```
overbit, line 226: This mutation must be protected by some kind of administrator authentication.

overbit (changes requested):

packages/amplication-server/src/core/workspace/workspace.service.ts
abrl91: https://github.com/amplication/amplication-cluster-configuration/pull/157

packages/amplication-server/src/core/workspace/workspace.service.ts
abrl91: we will test it

packages/amplication-server/src/core/workspace/workspace.resolver.ts
abrl91: I implemented `GqlCronGuard` => [commit](https://github.com/amplication/amplication/pull/7658/commits/39f344595681a241d68394eb5d3b6b869060b7e8)

I'm unsure if it's more secure than what we did with the custom actions. But I can definitely say that it's more consistent with what we have and usually use, which is Graphql with guards.

packages/amplication-server/src/guards/gql-cron.guard.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  constructor(private readonly configService: ConfigService) {
+    this.cronSecretKey = this.configService.get<string>(Env.CRON_SECRET_KEY);
+  }
+
+  canActivate(context: ExecutionContext): boolean {
+    const request = context.switchToHttp().getRequest();
+    const secretKey = request.headers[HEADER_KEY];
+
+    if (secretKey !== this.cronSecretKey) {
```
overbit, line 24: 👀

overbit (approved):

overbit (approved):

Issue #7582, opened 2023-12-12
As a user - I want to be able to copy/move entities between my services - part 1 - server-side support


mulygottlieb:

overbit:
## Test result: ❌ 

This resolver is changing, but in the current state the GraphQL input is missing some expected data.

overbit:
@mulygottlieb as per SOC2, we need the issue to contain the actual requirements without defining them on Notion unless we want to move to Notion for the project management.

mulygottlieb:
Passed QA:


https://github.com/amplication/amplication/assets/1912933/c4500d9e-70d7-450d-a42c-f00ee5e084bc

PR #7672, opened 2023-12-20
feat(server): add ability to move services across resources

morhag90:
Close: #7582 

## PR Details

Add a new function to copy/delete entities from resource X to resource Y
## PR Checklist


packages/amplication-server/src/core/entity/entity.service.ts
overbit, line 1: Let's improve the PR title as it will be visible in the release notes.
Looking at the description something like `feat(server): add ability to move services across projects` might be better

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -373,6 +379,192 @@ export class ResourceService {
...
+    if (!resource) throw new AmplicationError("Target Resource doesn't exist");
+
+    entitiesToCopy.map((entity) => {
+      if (resource.entities.find((x) => x.name === entity.name)) {
+        throw new AmplicationError(
+          `Entity ${entity.name} is already exist in resourceId: ${resource.id}, process abort`
+        );
+      }
+    });
+
```
overbit, line 409: Should we check also if an entity is related to another entity not part of the copy?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -373,6 +379,192 @@ export class ResourceService {
...
+    // 1.create resource entities
+
```
overbit, line 411: question: if all the entities to copy are not related to other entities, would it be enough to just update the entity resourceId with the new one?

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @overbit  It happened in the field creations: 
image

In this case, we create a non-relation field based on the relatedEntity id type.

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @overbit, No, Because we always need to create a new field unrelated to the existing ones. 
This is for a case where the user wants to keep the source resource entities.

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -599,6 +604,36 @@ export class EntityService {
     return true;
   }
 
+  async updateFieldDataTypeIdByRelatedEntity(
```
overbit, line 607: It looks generally good but I'd suggest adding the tests for this new feature

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -609,6 +614,36 @@ export class EntityService {
...
+        },
+      },
+    });
+
+    const idProp = relatedIdField.properties as unknown as {
+      [key: string]: JsonValue;
+    };
+
+    const idType = idProp["idType"];
+    if (idType === "CUID" || idType === "UUID") {
```
barshimi, line 635: just to upgrade:
instead of if, we can do:
const idTypeMap = {
  CUID: EnumDataType.SingleLineText,
  UUID:  EnumDataType.SingleLineText,
  AUTO_INCREMENT: EnumDataType.WholeNumber,
  AUTO_INCREMENT_BIG_INT: EnumDataType.WholeNumber
}
then get: field.dataType = idTypeMap[idType]

packages/amplication-server/src/core/resource/resource.service.spec.ts
```diff
@@ -187,6 +188,19 @@ const EXAMPLE_PROJECT: Project = {
   demoRepoName: undefined,
   licensed: true,
 };
+const EXAMPLE_TARGET_RESOURCE_ID = "exampleTargetResourceId";
+const EXAMPLE_TARGET_RESOURCE: Resource = {
+  id: EXAMPLE_TARGET_RESOURCE_ID,
+  resourceType: EnumResourceType.Service,
+  createdAt: new Date(),
```
barshimi, line 195: you can make new Date() once instead of twice

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -609,6 +614,36 @@ export class EntityService {
...
+          entityId: relatedEntityId,
+        },
+      },
+    });
+
+    const idProp = relatedIdField.properties as unknown as {
+      [key: string]: JsonValue;
+    };
+
+    const idType = idProp["idType"];
```
abrl91, line 634:
```suggestion
    const idType: types.Id["idType" ]= idProp["idType"];
```

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -846,6 +881,133 @@ export class EntityService {
     });
   }
 
+  async deleteEntityFromSourceResource(
```
abrl91, line 884: The name is confusing. Can we change it to:
```suggestion
  async deleteEntityFromSource(
```

packages/amplication-server/src/core/entity/entity.service.ts
morhag90: @abrl91  Did you mean something like that: 
image

barshimi (approved):

PR #7690, opened 2023-12-25
refactor(server): emit update project and service licensed field on promotional entitlement

abrl91:
Part of: https://github.com/amplication/private-issues/issues/90

## PR Details

- emit update project and resource licensed field on promotional entitlement
- refactor logic of blocked features on service, such as add entity, add entity field, add role - checking first if the project is not licensed because if the project is not licensed, the resurce (service) should not be licensed
- update entity tests

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.spec.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.spec.ts
@@ -673,7 +673,7 @@ describe("EntityService", () => {
         billingServiceIsBillingEnabledMock.mockReturnValue(false);
       });
 
-      it("should not throw an error even when the project or the service is not licensed", async () => {
+      it("should not throw billing limitation error even when the project or the service is not licensed", async () => {
         const resource = {
           ...EXAMPLE_RESOURCE,
           licensed: false,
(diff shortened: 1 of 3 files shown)


packages/amplication-server/src/core/entity/entity.service.spec.ts
```diff
@@ -646,6 +664,54 @@ describe("EntityService", () => {
...
+      it("should not throw an error when project or service is under license", async () => {
+        await expect(
+          service.checkServiceLicense(EXAMPLE_RESOURCE) // in the example resource the project and the service are licensed
+        ).resolves.not.toThrow(BillingLimitationError);
+      });
+
+      it("should throw an error when project or service is not under license", async () => {
+        const resource = {
+          ...EXAMPLE_RESOURCE,
+          licensed: false,
```
mulygottlieb, line 702: We can add a separate test for each of the licenses (one with project.licensed=F and resource.licensed=T and one test with the opposite). Not mandatory, for your consideration. I will approve the PR anyway and you can add this if you want.

mulygottlieb (approved):
Approved. See one comment on the test for your consideration.

packages/amplication-server/src/core/entity/entity.service.spec.ts
abrl91: if project.licensed=false and resource.licensed=true => we probably did something wrong, but I can add this test because it's a valid scenario based on the function.

PR #7709, opened 2023-12-29
Feat(server+client): create preview account type with auto-generated workspace project and service

abrl91:
Close: https://github.com/amplication/private-issues/issues/108

## PR Details

- added two new fields on account model: `previewAccountType` and `previewAccountEmail`
- create "signup" flow via URL that contains an email provided from an input and preview account type: `http://localhost:3001/signup/preview-account?email=amit+test@amplication.com&previewAccountType=BreakingTheMonolith`
- in the end of this "signup" flow the user is redirected to its newly created workspace/project/service
- default plugins: auth-core, auth-jwt, postgres
- user auth entity

## PR Checklist


packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -423,6 +527,72 @@ export class AuthService {
...
+        include: WORKSPACE_INCLUDE,
+      },
+      account.id
+    )) as unknown as Workspace & { users: AuthUser[] };
+    const [user] = workspace.users as AuthUser[];
+
+    const project = await this.projectService.createPreviewProject(
+      {
+        data: {
+          name: account.previewAccountType,
```
mulygottlieb, line 546: Do we want to use the same project name for all preview workspaces? (perhaps it's in the spec and I don't remember)

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -423,6 +527,72 @@ export class AuthService {
...
+              generateGraphQL: true,
+              generateRestApi: true,
+              serverPath: "server",
+            },
+          },
+        },
+      },
+      user,
+      pluginsToInstall: [],
+      requireAuthenticationEntity: true,
```
mulygottlieb, line 585: What does it mean not to install any plugin but mandate "requireAuthenticationEntity"? Would it create the user entity? Would it force the user to create one? 
Don't we want to install the default auth plugins?

packages/amplication-server/src/core/billing/billing.service.ts
```diff
@@ -294,6 +294,18 @@ export class BillingService {
...
 
+  async provisionPreviewCustomer(workspaceId: string): Promise<null> {
+    if (!this.isBillingEnabled) {
+      return;
+    }
+
+    await this.stiggClient.provisionCustomer({
+      customerId: workspaceId,
+      shouldSyncFree: false,
+      subscriptionParams: { planId: BillingPlan.Enterprise }, // TODO: CHANGE IT TO PREVIEW PLAN
```
mulygottlieb, line 305: How will we not forget this part (set a preview plan)? Perhaps add it to the Enum (BillingPlan) already if possible?

packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -106,6 +106,30 @@ export class ProjectService {
     return project;
   }
 
+  async createPreviewProject(
```
mulygottlieb, line 109: This seems repetitive of the usual createProject method. I'm not sure it's really a big problem, but perhaps you can think of refactoring out some of the parts so you will keep DRY.

abrl91: The difference is the report usage, we don't need to report usage for the preview account and I don't want to add a pram "doReportUsage" or something like that

packages/amplication-server/src/core/billing/billing.service.ts
abrl91: the TODO is a keyword on my editor and this is also one of the reasons this pr is still a draft

packages/amplication-server/src/core/auth/auth.service.ts
abrl91: plugins to install are "other plugins". This is for future usage. Have a look at the `createPrevewService` implementation. We can move the logic to here but I though that here we just need to send a list of plugins to install that are not the "default". The resource service handles the default behavior

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -381,6 +389,88 @@ export class ResourceService {
...
+    const defaultDBPlugin: PluginInstallationCreateInput = {
+      displayName: "postgres",
+      pluginId: "db-postgres",
+      npm: "@amplication/plugin-db-postgres",
+      version: "latest",
+      enabled: true,
+      resource: { connect: { id: resource.id } },
+    };
+
+    const plugins = [
```
mulygottlieb, line 454: So this answers another question I asked about installing the default plugins. I find it confusing that this method accepts "pluginsToInstall" as a parameter but still might install up to 3 other plugins. This is very confusing / not intuitive. Why aren't those plugins passed in from the caller as "pluginsToInstall", or remove this parameter and allow a separate way to install plugins?

packages/amplication-server/src/core/auth/auth.service.ts
abrl91: Yep. This is based on the spec.

packages/amplication-client/src/User/SignupPreviewAccount.tsx
mulygottlieb, line 1: Can you please ask Mor to review this single file? I don't feel I know enough to give this file a really good review

packages/amplication-server/src/core/auth/auth.service.ts
mulygottlieb: I wrote a separate comment on that, so I am moving the discussion there and marking this one as resolved. However, I think if the parameter name would change from `pluginsToInstall` to something like `nonDefaultPluginsToInstal` or `additionalPluginsToInstall` it will become more readable

packages/amplication-server/src/core/project/project.service.ts
mulygottlieb: Perhaps at least extract the common parts to utility methods so we won't have the same `prisma.create` statement twice, etc. (if the schema changes later, we would need to touch two places)

packages/amplication-server/src/core/resource/resource.service.ts
mulygottlieb: Let's just change the parameter name from `pluginsToInstall` to `nonDefaultPluginsToInstall` or `additionalPluginsToInstall` or similar, to make it clearer

abrl91: as I see it, `pluginsToInstall` (additionalPluginsToInstall) should be plugins that in the future we might want to use for other preview features. For now, it passed as a hard-coded value (empty array), but in the future, it might be a calculated value based on the preview account type.

`createPreviewService` handles the installation of the default plugins as without these plugins the service will not function as expected.

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { useHistory } from "react-router-dom";
+import { gql, useMutation } from "@apollo/client";
+import { formatError } from "../util/error";
+import "./Signup.scss";
+import { setToken } from "../authentication/authentication";
+
+const SignupPreviewAccount = () => {
+  const history = useHistory();
+
+  const [signupPreviewAccount, { loading, error }] = useMutation<{
```
morhag90, line 11: Move the mutation to hook

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+  }>(SIGNUP_PREVIEW_ACCOUNT);
+
+  useEffect(() => {
+    const queryString = window.location.search;
+    const urlParams = new URLSearchParams(queryString);
+    const email = urlParams.get("email");
+    const previewAccountType = urlParams.get("previewAccountType");
+
+    if (email && previewAccountType) {
+      signupPreviewAccount({
```
morhag90, line 27: Move this logic to a separated hook and expose a function that received the relevant params

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+        variables: {
+          data: {
+            previewAccountEmail: email,
+            previewAccountType: previewAccountType,
+          },
+        },
+        onCompleted: ({ signupPreviewAccount }) => {
+          console.log(signupPreviewAccount, "signupPreviewAccount");
```
morhag90, line 35: delete the log

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+            signupPreviewAccount;
+
+          setToken(token);
+          history.push(`/${workspaceId}/${projectId}/${resourceId}`);
+        },
+      }).catch(console.error);
+    }
+  }, [signupPreviewAccount, window.location.search]);
+
+  const errorMessage = formatError(error);
```
morhag90, line 46: expose this error from the hook

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  errorMessage && <p>{errorMessage}</p>;
+
+  return loading && <p>Please wait...</p>;
```
morhag90, line 50: Use    <CircularProgress centerToParent /> instead of <p>

packages/amplication-client/src/User/SignupPreviewAccount.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+
+export default SignupPreviewAccount;
```
morhag90, line 53: move this to query file

morhag90 (approved):

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -423,6 +539,125 @@ export class AuthService {
...
+        description,
+        resourceType: EnumResourceType.Service,
+        project: {
+          connect: {
+            id: projectId,
+          },
+        },
+        serviceSettings: {
+          authProvider: EnumAuthProviderType.Jwt,
+          authEntityName,
```
mulygottlieb, line 646: didn't we say we want the preview account to be created with no auth and no User entity?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -381,6 +389,97 @@ export class ResourceService {
...
+      "create resource"
+    );
+
+    await this.prisma.resourceRole.create({
+      data: { ...USER_RESOURCE_ROLE, resourceId: resource.id },
+    });
+
+    if (requireAuthenticationEntity) {
+      await this.entityService.createDefaultEntities(resource.id, user);
+      serviceSettings.authEntityName = USER_ENTITY_NAME;
```
mulygottlieb, line 417: Shouldn't this come from a parameter from the caller? (the name of the auth entity)?

mulygottlieb (approved):
Nice work!

PR #7737, opened 2024-01-03
chore(server): update stigg sdk version

abrl91:
Part of: amplication/private-issues#108

## PR Details

In order to provision a customer with a plan, skip the trial, and have a strong solution, we need to upgrade the Stigg SDK version that we are using.

Without it, the solution is to provision a customer, and then provision a subscription with a plan and `{skipTrial:true}`.

The property `skipTrial` is about to be deprecated, and the new version doesn't contain breaking changes.

## PR Checklist


overbit (approved):
Very well prepared pr. Nice work @abrl91 💯!

PR #7740, opened 2024-01-04
chore(client): add support for realtime user monitoring

overbit:
Close: https://github.com/amplication/amplication-cluster-configuration/issues/190

## PR Details

Add client monitoring leveraging RUM (Realtime-User-Monitoring). 
image

## PR Checklist


yuval-hazaz:
@overbit I like this change
We need to make sure the client works normally without the keys and connection to the RUM account.

overbit:
@yuval-hazaz  @mulygottlieb we could have perhaps a quick chat about the next actions?

packages/amplication-client/src/util/monitoring.ts
```diff
@@ -0,0 +1,51 @@
...
+    const APPLICATION_REGION = "us-east-1";
+
+    monitoring = new AwsRum(
+      REACT_APP_RUM_APP_ID,
+      APPLICATION_VERSION,
+      APPLICATION_REGION,
+      config
+    );
+  } catch (error) {
+    // Ignore errors thrown during CloudWatch RUM web client initialization
```
abrl91, line 38: I think we should add `console.error(error);`

overbit: No, I don't think we need to do so as it won't give us any thing we would act on.

abrl91 (approved):

overbit:
Reopened as we had to revert due to issues with tracing

PR #7748, opened 2024-01-04
refactor(server+client): analytics for preview user

abrl91:
Close: https://github.com/amplication/private-issues/issues/114

## PR Checklist
- use the preview email for preview users
- refacotor current "user" (account) and create hook + use the right data for previeww user analytics

### DIFF
--- a/packages/amplication-client/src/Components/UserBadge.tsx
+++ b/packages/amplication-client/src/Components/UserBadge.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from "react";
-import { gql, useQuery } from "@apollo/client";
+import { useQuery } from "@apollo/client";
 
 import * as models from "../models";
 
@@ -8,6 +8,7 @@ import { UserAvatar, Tooltip } from "@amplication/ui/design-system";
 
 import "./UserBadge.scss";
 import { identity } from "../util/analytics";
+import { GET_USER } from "../Profile/accountQueries";
 
 type TData = {
   me: {
@@ -47,17 +48,3 @@ function UserBadge() {
 }
 
 export default UserBadge;
-
-export const GET_USER = gql`
-  query getUser {
-    me {
-      account {
-        id
-        email
-        firstName
-        lastName
-        createdAt
-      }
-    }
-  }
-`;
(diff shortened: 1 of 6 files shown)


packages/amplication-client/src/Profile/hooks/UseCurrentUser.ts
```diff
@@ -0,0 +1,36 @@
...
+import { GET_USER } from "../accountQueries";
+import { identity } from "../../util/analytics";
+
+type TData = {
+  me: {
+    account: models.Account;
+  };
+};
+
+export const useCurrentUser = () => {
```
barshimi, line 14: this is redundant. on every route we are using  useAuthenticated, so you can use it from there and also you need to remember that when you using a hook on a component, every time yo rernder it you'll restart your hook. so you better place it on a higher level or do it on a route level/App.
also, we are calling GET_USER at the higher level, so it will create another call for it

packages/amplication-client/src/Profile/accountQueries.ts
```diff
@@ -0,0 +1,25 @@
+import { gql } from "@apollo/client";
+
+export const GET_USER = gql`
+  query getUser {
+    me {
+      account {
+        id
+        email
+        previewAccountEmail
```
barshimi, line 9: we'll need to add the previewAccountType as well

packages/amplication-client/src/Profile/ProfileForm.tsx
```diff
@@ -45,8 +46,7 @@ const FORM_SCHEMA = {
 const CLASS_NAME = "profile-form";
 
 const ProfileForm = () => {
-  const { data, error, refetch } = useQuery<TData>(GET_USER);
-
+  const { currentUser, error, refetch } = useCurrentUser();
```
barshimi, line 49: you should pass it as props from the context or the route itself

packages/amplication-client/src/Components/UserBadge.tsx
```diff
@@ -1,63 +1,27 @@
...
-  };
-};
 const TOOLTIP_DIRECTION = "sw";
 
 function UserBadge() {
-  const authenticated = useAuthenticated();
-  const { data } = useQuery<TData>(GET_USER, {
-    skip: !authenticated,
-  });
+  const { currentUser } = useCurrentUser();
```
barshimi, line 10: take a look at my comment at the hook itself

packages/amplication-client/src/Profile/accountQueries.ts
abrl91: why?

abrl91: I added it

morhag90 (approved):

PR #7751, opened 2024-01-05
Feat(server): preview account limitations

abrl91:
Close: https://github.com/amplication/private-issues/issues/112

## PR Details

- block creation of a workspace
- block commit (build) 
- both "create workspace" and "build" are boolean entitlements that if the `hasAccess` is true - that means that it needs to be blocked.
- block invite a member based on usage (metered entitlement). That was already implemented but just a small refactoring - moving the logic to a separated function and check for the current usage and the limit instead of the hasAccess.

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/project/project.service.ts
+++ b/packages/amplication-server/src/core/project/project.service.ts
@@ -276,10 +276,6 @@ export class ProjectService {
       })
       .workspace();
 
-    if (!workspace) {
-      throw new Error("workspace not found");
-    }
-
     const blockBuildEntitlement =
       await this.billingService.getBooleanEntitlement(
         workspace.id,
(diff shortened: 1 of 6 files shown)


packages/amplication-server/src/core/project/project.service.ts
```diff
@@ -263,6 +263,32 @@ export class ProjectService {
...
+    const workspace = await this.prisma.user
+      .findUnique({
+        where: {
+          id: userId,
+        },
+      })
+      .workspace();
+
+    if (!workspace) {
+      throw new Error("workspace not found");
```
barshimi, line 280: do we want to throw error and stop the service or we just want to make it clear that there is no workspace and continue

overbit (approved):
LGTM 🚀

Issue #7760, opened 2024-01-07
CI is failing on build and dockerize due to Nx cache issues


abrl91:
<img width="1046" alt="image" src="https://github.com/amplication/amplication/assets/39680385/b1a09628-7289-47cd-a27d-6b1ea97705f1">

<img width="1046" alt="image" src="https://github.com/amplication/amplication/assets/39680385/b5ee7c82-9478-405d-a82e-78ccff0edfd1">

<img width="1057" alt="image" src="https://github.com/amplication/amplication/assets/39680385/68b60a43-0eb5-42aa-a283-7834c88de3b4">

overbit:
Fixed by clearing the node_module cache

overbit:
Still happening
https://github.com/amplication/amplication/actions/runs/7503351067/job/20427925039#step:10:82

mulygottlieb:
Looks good after upgrading the nx version:


![Image](https://github.com/amplication/amplication/assets/1912933/71f0fe28-6846-493b-b225-663d17a42ceb)


Will re-open if this surfaces again, for now it seems okay so moving this to passed QA

PR #7761, opened 2024-01-07
Fix(workflows): nx ci reset nx cache

abrl91:
Close: https://github.com/amplication/amplication/issues/7760

## PR Details

With this addition, the `ci.yaml` will handle cache resetting in the CI job (by default false), and the `nx.template.yml` will do the same in the NX job. 

When we need to reset the cache, we can set nx-reset-cache to true via the workflow dispatch inputs.

## PR Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,11 @@ on:
         type: boolean
         description: Forces Nx to consider all projects (apps and libs) as affected.
         default: false
+      nx-reset-cache:
+        type: boolean
+        description: Resets Nx cache.
+        required: false
+        default: false
   pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -64,6 +70,10 @@ jobs:
         if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
         run: npm ci
 
+      - name: NX Reset Cache
+        if: ${{ env.NX_RESET_CACHE == 'true' }}
+        run: npx nx reset
```
overbit, line 75: as per docs, this command will also stop the nx deamon  so we minght need something different

Issue #7742, opened 2024-01-04
`@amplication/util/nestjs/kafka` KafkaProducer must validate message against schema before emitting 


overbit:
### Feature description

Currently KafkaProducerService doesn't perform any data validation against the schema registry defined event key and value before emitting a Kafka event.
This can be the source of big issues where a message not conform to the specs is published as all the consumer might fail and stop processing other events due to unexpected data.

Instead, the producer should validate and through relevant exceptions in case of validation failures to avoid issues down the line.

xonx4l:
Add:KafkaProducer validate message against schema before emitting . #7776.  Feedback@overbit.

PR #7776, opened 2024-01-10
Add:KafkaProducer validate message against schema before emitting .

xonx4l:
Close: #7742

## PR Details
This  method first validates the message against the provided schema IDs. If the message does not conform to the specified schema, it throws an error. Then, it proceeds to serialize the message using the provided schema IDs and emits the message to the Kafka client. This way, the KafkaProducerService ensures that only valid messages are published to the Kafka topic, preventing potential issues for the consumers.

## PR Checklist

- [x ] `npm test` doesn't throw any error


yuval-hazaz:
closed due to inactivity

Issue #6849, opened 2023-09-05
As a User - I'd like to see a mention of the service built in the generated PR 


overbit:
### Feature description

When using a PR per service, I'd like to see PR names containing the service created to make easier to understand what is what.

Current:
![image](https://github.com/amplication/amplication/assets/2861984/b261e265-99c7-4b31-ac8f-6f45d9f319d7)

Ideal example:
![image](https://github.com/amplication/amplication/assets/2861984/ae26bb85-7e89-4e04-ad4b-7a6f89bda874)

PR naming rules:
- for unique PR created : `chore(amplication): updated generated code`
- for one PR per service: `chore(amplication): updated generated code for <serviceName>`

mulygottlieb:
Passed QA:

![Image](https://github.com/amplication/amplication/assets/1912933/87ad3b2e-30d0-4c39-aba3-74888bb1238f)

PR #7781, opened 2024-01-10
feat(git): include service-name in generated PRs

Shurtu-gal:
Close: #6849 

## PR Details

- Add service name to generated PRs for users with separate branches per resource.

## PR Checklist


libs/util/git/src/git-client.service.ts
```diff
@@ -347,7 +350,7 @@ export class GitClientService {
       });
     }
 
-    return pullRequest.url;
+    return pullRequest?.url || "";
```
overbit, line 353: As empty string is a valid value, I'd suggest to use the nullish coalescing operator instead.

```suggestion
    return pullRequest?.url ?? "";
```

Shurtu-gal: Done @overbit.

overbit (approved):
Thank you @Shurtu-gal! 💯

PR #7810, opened 2024-01-14
Feat(server+client): preview account complete signup

abrl91:
Close: https://github.com/amplication/private-issues/issues/117

## PR Checklist


packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -316,6 +316,57 @@ export class AuthService {
...
+        throw Error("Failed to create new Auth0 user");
+    }
+
+    const userEmail = existingAuth0User
+      ? currentAccount.previewAccountEmail
+      : auth0User.data.email;
+
+    const resetPassword = await this.resetAuth0UserPassword(userEmail);
+
+    if (!resetPassword.data)
```
morhag90, line 341: Could this object return undefined results?

packages/amplication-server/src/core/auth/auth.resolver.ts
```diff
@@ -54,6 +54,14 @@ export class AuthResolver {
         });
   }
 
+  @Mutation(() => String)
+  @UseGuards(GqlAuthGuard)
+  async completeSignupWithBusinessEmail(
```
overbit, line 59: is this related to the PR?

abrl91: of course..it's the resolver

abrl91: I don't know why it seems like I did it. It's what Shimi did. Maybe something with the merges...

packages/amplication-server/src/core/auth/auth.service.ts
abrl91: The data will be undefined, so it's ok...

packages/amplication-server/src/core/auth/auth.resolver.ts
abrl91: yes...it's the resolver of my new endpoing

overbit: Then the naming is misleading as it might assume that is any business email and not preview only

morhag90 (approved):

Issue #7818, opened 2024-01-16
Expose BTM AI recommendation for an easy display to the user


overbit:
Once BTM generation is completed, the suggestion must be persisted and be retrievable by any client through a GraphQL query. 
The query must be protected by auth, and the resulting data must be returned only to the users of the workspace where the resource belong to.
In any other case, an unauthorised error should be returned.

## Tech notes

GQL query: 
```gql
query FinalizeBreakServiceIntoMicroservices($userActionId: String!) {
  finalizeBreakServiceIntoMicroservices(userActionId: $userActionId) {
    data {
      microservices {
        name
        functionality
        dataModels {
          originalEntityId
          name
        }
      }
    }
    status
    originalResourceId
  }
}
```

The query above will be easy to map to the data needed in the UI [#7819](https://github.com/amplication/amplication/issues/7819) and the Architecture UI https://github.com/amplication/amplication/issues/7693

overbit:
```ts
resources.map(resource => {
    name: resource.name
    entities: entities.filter(entity => entity.targetResourceId === resource.id).map(entity => {
        name: entity.name
    })
})
```

abrl91:
The recommendation output was changed to
```
"microservices": [
  {
    "name": string // service name
    "functionality": string, // description
    "dataModels": [
      {
        "originalEntityId": string,
        "name": string // entity name
      }
    ]
  },
]
```

overbit:
Test result ✅ 

![Image](https://github.com/amplication/amplication/assets/2861984/13a41d45-b3b2-45af-a0ed-382540137c62)

PR #7834, opened 2024-01-18
feat(server): add `btmRecommendationModelChanges` query for client recommendation collection

overbit:
Close: #7818

## PR Details

Expose a new gql Query for collecting BTM recommendation resource.

Documentation has been added to the gql schema itself to simplify integration. Check `btmRecommendationModelChanges` in `server` playground or apollo studio.

## PR Checklist


packages/amplication-server/src/core/ai/ai.service.ts
```diff
@@ -222,7 +227,7 @@ export class AiService {
               name: resource.name,
```
morhag90, line 227: Why do you need the where filter for the resource entities?

packages/amplication-server/src/core/ai/btm-manager.service.ts
```diff
@@ -56,17 +53,25 @@ export class BtmManagerService {
...
+              const entityNameIdMap = originalResource.entities.reduce(
+                (map, entity) => {
+                  map[entity.name] = entity;
+                  return map;
+                },
+                {} as Record<string, EntityPartial>
+              );
+
+              return {
+                id: undefined,
```
morhag90, line 66: Is this the tempResourceId field? If it is, Why not generate a tempId like a random guid?

packages/amplication-server/src/core/ai/ai.service.ts
overbit: Where do you mean?

packages/amplication-server/src/core/ai/dto/btm-recommendation-model-changes-args.dto.ts
abrl91, line 1: The name of the file is different from what we are doing. Even though what you did is the recommended practice, I think we should be consistent (same for other DTOs file names)

overbit: handled here https://github.com/amplication/amplication/pull/7842

abrl91 (approved):

Issue #6549, opened 2023-07-21
🐛 Bug Report: When a user creates a service and then tries to discard the changes - he gets errors and then a blank page


abrl91:
### What happened?

https://github.com/amplication/amplication/assets/39680385/f4b2323f-8768-4f6e-93fc-5cf78b299296

### What you expected to happen

- Discard pending changes should not throw an error
- After discarding pending changes, we should not see a blank page. We should see the entities page with 0 entities like we see after a refresh

### How to reproduce

see the screen recording

### Amplication version

1.8.0

mulygottlieb:
Note to self: perhaps this was already fixes with the fix for creating new service settings if they were deleted by discard - need to try and reproduce

mulygottlieb:
In general, when you create a new service with the wizard and then immediately after the wizard you hit "discard pending changes" you will get an "Internal Server Error" response:


![Image](https://github.com/amplication/amplication/assets/1912933/7e33ea21-ad1e-42df-a7e4-d7f080e999bc)

mulygottlieb:
@Shurtu-gal  I am re-opening this one, see the video attached here. I am still getting errors after discarding the changes.
(not a blocker for version release as this is not a regression)


https://github.com/amplication/amplication/assets/1912933/36444604-f552-4ae2-976f-1c5482dcd37f

Shurtu-gal:
> @Shurtu-gal I am re-opening this one, see the video attached here. I am still getting errors after discarding the changes. (not a blocker for version release as this is not a regression)
> Screen.Recording.2024-01-15.at.11.15.02.mov

Sure, let me look into this again.

mulygottlieb:
@Shurtu-gal, this is still not working as expected.
I created a new service with the wizard, chose the order management template, and no auth. Then, when I finished the wizard and hit discard, I get:

![Image](https://github.com/amplication/amplication/assets/1912933/5d506284-d2c1-419f-a94b-d861367130f9)


If I do the same thing and choose an empty service but with Auth, I get another error that I cannot delete the Auth entity (User):

![Image](https://github.com/amplication/amplication/assets/1912933/031ad9d4-08b3-4713-9717-806c974dff06)

mulygottlieb:
@PazYanoverr  this we re-opened back in January and we didn't do anything with it since. Please review and LMK if you want to place this in the todo somewhere (yours to prioritize) or we can move this to the backlog

PR #7846, opened 2024-01-20
fix: there may be no plugins installed

Shurtu-gal:
Close: #6549 

## PR Details

There may be no installed plugins in a newly generated PR. Therefore, the following lines give an error:- 

https://github.com/amplication/amplication/blob/ba64ad704d4eb4b9492928f9ca6fe9f8e2e73c21/packages/amplication-client/src/Entity/EntityList.tsx#L82-L86

image

[Check Video](https://www.loom.com/share/faf8d5e742b84b3c98c11e0b5893590d?sid=4910ffca-3aac-4c65-a197-07cbcb8b4f17 "Check Video")

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Entity/EntityList.tsx
+++ b/packages/amplication-client/src/Entity/EntityList.tsx
@@ -83,7 +83,7 @@ const EntityList: React.FC<Props> = ({ match, innerRoutes }) => {
     pluginInstallations?.filter(
       (x) =>
         x?.configurations?.requireAuthenticationEntity === "true" && x.enabled
-    ).length > 0;
+    )?.length > 0;
 
   const handleNewEntityClick = useCallback(() => {
     setNewEntity(!newEntity);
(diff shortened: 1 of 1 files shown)


packages/amplication-client/src/Entity/EntityList.tsx
```diff
@@ -82,7 +82,7 @@ const EntityList: React.FC<Props> = ({ match, innerRoutes }) => {
   const isUserEntityMandatory =
     pluginInstallations?.filter(
       (x) =>
-        x.configurations?.requireAuthenticationEntity === "true" && x.enabled
+        x?.configurations?.requireAuthenticationEntity === "true" && x.enabled
```
abrl91, line 85: what will happen if the first equation is undefined or null? I mean this part: `x?.configurations?.requireAuthenticationEntity`
I think the `.length` will throw

Shurtu-gal: Yeah, missed that would be adding it. Thanks 🙇🏻

Shurtu-gal: Added that @abrl91.

mulygottlieb (approved):

PR #7848, opened 2024-01-21
chore(devops): configuration for local DSG dockerfile

mulygottlieb:
Relates to: https://github.com/amplication/amplication/pull/7338/

Closes: https://github.com/amplication/amplication/issues/7337

## PR Checklist


yuval-hazaz (approved):

yuval-hazaz:
@overbit can you also review this please?
@mulygottlieb are we ready to push this to next ?

overbit:
The changes in general look good. 
The only downside is that it won't force contributor to create separate PR for deployable as it won't require packages to be published first.
We need to remember that it's a monorepo and not a monolith.

packages/data-service-generator/scripts/package-container.sh
```diff
@@ -21,5 +21,7 @@ if [[ $PRODUCTION_TAGS = "true" ]]; then
 
     npx nx internal:package:container data-service-generator --prod
 else
-    npx nx internal:package:container data-service-generator 
+    echo "Local deployment detected"
+    echo "Using Dockerfile-dev to include latest code-gen-types"
+    npx nx internal:package:container data-service-generator --dev
```
overbit, line 26: `--dev` is an invalid option. Only production configuration has the alias `--prod`

```suggestion
    npx nx internal:package:container data-service-generator --configuration dev
```

overbit (changes requested):

overbit (approved):

PR #7849, opened 2024-01-21
Feat/project architecture

morhag90:
Close: #127

## PR Details

Add Architecture tab logic (break the monolith page). 
## PR Checklist


packages/amplication-server/src/core/resource/dto/ResourceCreateCopiedEntitiesInput.ts
```diff
@@ -1,17 +1,55 @@
...
+  @Field(() => [CopiedEntity], {
+    nullable: false,
+  })
+  entitiesToCopy!: CopiedEntity[];
+}
+
+@InputType({
+  isAbstract: true,
+})
+export class ModelGroupResource {
```
yuval-hazaz, line 26: why are we using the "ModelGroup" name on the server?

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
```diff
@@ -13,6 +13,11 @@ export class ResourceCreateInput {
   })
   name!: string;
 
+  @Field(() => String, {
+    nullable: true,
+  })
+  tempId?: string;
```
yuval-hazaz, line 19: why is that here?

packages/amplication-server/src/core/resource/resource.resolver.ts
```diff
@@ -160,20 +160,14 @@ export class ResourceResolver {
...
+  @Mutation(() => [Resource], { nullable: false })
   @Roles("ORGANIZATION_ADMIN")
-  @AuthorizeContext(
-    AuthorizableOriginParameter.ProjectId,
-    "data.project.connect.id"
-  )
-  async createResourceEntitiesFromExistingResource(
-    @Args() args: CreateResourceEntitiesArgs,
+  @AuthorizeContext(AuthorizableOriginParameter.ProjectId, "data.projectId")
+  async copiedEntities(
```
yuval-hazaz, line 166: This name is not good

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -480,43 +483,177 @@ export class ResourceService {
     return resource;
   }
 
-  async createResourceEntitiesFromExistingResource(
-    args: CreateResourceEntitiesArgs,
+  async createModelGroupService(
```
yuval-hazaz, line 486: we should not use "modelgroup" in the server. This is a name of the concept used in the client to group models - meaning using a resource to hold entities

yuval-hazaz, line 486: I dont think this function is the best decision.
we need to use a function that know how to create a service with default settings. it should be generic, and not something specific to the BTM

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -571,66 +708,61 @@ export class ResourceService {
...
+      throw new Error(
+        `Failed to create entity "${entityToCopy.name}" due to ${error.message}`
+      );
     }
 
-    return copiedEntities;
+    return copiedEntity;
   }
 
   async createCopiedEntityIdField(copiedEntityId: string): Promise<void> {
```
yuval-hazaz, line 768: the name "copiedEntities" does not mean anything to anyone who is not familiar with the context

packages/amplication-server/src/models/Resource.ts
```diff
@@ -16,6 +16,11 @@ export class Resource {
   })
   id!: string;
 
+  @Field(() => String, {
+    nullable: true,
+  })
+  tempId?: string;
+
```
yuval-hazaz, line 23: why is that here?

yuval-hazaz (changes requested):
@morhag90 impressive work!!! 
see my comments. I am happy to review together if needed

mulygottlieb (approved):

Issue #7693, opened 2023-12-26
As a user - I want to be able to copy/move entities between my services - client side


morhag90:
Tasks todo: 

- [x] Select “current” interactive service => mark the selected service as the edited one, the rest of the services on disabled mode. (disable movement or add extent to non-interactive services) => css issues. 
- [x] Show service bar map by Figma design.
- [x] Add service serach. 
- [x] Add cancel changes.
- [x] Add the ability to create a new service.  (the new service/s will be created only after clicking the apply changes button). 
- [x] Add a new mutation to update the new changes in the server. 
- [x] Save the entire list of nodes and edges in local storage. 
- [x] Improve design and responsiveness of group custom drag handle. 
- [ ] ~~Update pointer class for groups and models when draggable or not.~~
        tracked by https://github.com/amplication/amplication/issues/7990
- [x] Create a new entitlement "feature-redesign-architecture and check it on client" . 
- [x] https://github.com/amplication/amplication/issues/7864

overbit:
## Test result: ❌ 

>   Update pointer class for groups and models when draggable or not.

https://github.com/amplication/amplication/assets/2861984/f02e1740-5181-47e3-b6af-6db12638a09a

mulygottlieb:
I opened https://github.com/amplication/amplication/issues/7990 to track the last issue, the mouse pointer.
I will QA everything else without the mouse pointer, and if everything else is good - I will move this one to Passed QA and leave the other ticket about the mouse pointer for prioritization by @PazYanoverr.

mulygottlieb:
Other than the mouse pointer (tracked by #7990), this passed QA:


https://github.com/amplication/amplication/assets/1912933/a81f5e9f-eb98-4ba6-b26c-8e0c82382481

PR #7882, opened 2024-01-29
fix(client): reload localStorage when projectId changed

morhag90:
Close: #7693 

## PR Details

 reload localStorage when projectId changed,

## PR Checklist


packages/amplication-client/src/Project/ArchitectureConsole/hooks/useModelOrganizer.tsx
```diff
@@ -136,6 +143,7 @@ const useModelOrganization = () => {
     setCurrentTheme,
     showRelationDetails,
     currentProject,
+    currentResourcesData,
```
mulygottlieb, line 146: What is this? Should it be `currentResourcesStorageData` ?

morhag90: Yes, Fixed

mulygottlieb (approved):

Issue #7492, opened 2023-11-23
By default, the "Username" field in the "User" entity should be a "Unique Field"


PazYanoverr:
As for now, the "Username" field in the "User" entity is not a unique field by default.
This should be changed so the "Username" field is unique by default and disabled so users won't be able to change it.
This change should apply to new services only.

abrl91:
@PazYanoverr 
Is this requirement is only for new services? or also for existing?
If it is also for existing services, we will need a migration + it will change the users' schema and they will also need a migration for their DB

PazYanoverr:
@abrl91 - Good question. This change should apply to new services only. I will update the ticket's description accordingly. Thanks!

mulygottlieb:
Passed QA:



![Image](https://github.com/amplication/amplication/assets/1912933/4f8740a8-8a54-4551-93cf-3381796a02b2)

PR #7910, opened 2024-02-01
feat(client): disable unique flag for auth entity specific field types

overbit:
Close: #7492

## PR Details

Restrict users options on authentication entities specific field types like `Username` and `Password`.

In the specific, users won't be able to modify the unique flag as it will default to:
- Username: `unique=true`
- Password: `unique=false`

## PR Checklist


yuval-hazaz (approved):
good job

PR #7916, opened 2024-02-02
refactor(server): business email signup flow to allow tracking at the end of the process

overbit:
Fixes: https://github.com/amplication/private-issues/issues/131

## PR Details

This pull request includes changes to the `amplication-server` package. The changes involve business email signup flow, refactoring the identity provider types in the `account.service.ts` file and removing the `Auth0Signup` preview account type in the `auth.controller.ts` and `auth.service.spec.ts` files.

The drive of this refactoring/rework was due to the requirement to track a signup event ONLY at the registration completed, aka password reset and successful login. 

### Changes
- Remove amplication account creation on business email signup request and leverage OIDC login account autocreation functionality (same as any other OIDC authorisation grant flow.  
- Refactored the identity provider types in `account.service.ts`. The `identityProvider` parameter in the `createAccount` method now accepts `IdentityProvider | IdentityProviderPreview` instead of `string`.
- Removed the `Auth0Signup` preview account type as it will to reflect the nature of the preview account. 
  👀 have a look to the [modified migration script](https://github.com/amplication/amplication/pull/7916/files#diff-58915755139c9f202b7831c96d761f154129172ac89866afbe54b74d6b896991R9-R11
) that will allow us to avoid manual migrations 

## PR Checklist

### DIFF
--- a/packages/amplication-prisma-db/prisma/migrations/20240202103529_remove_auth0_preview_account_type_from_enum/migration.sql
+++ b/packages/amplication-prisma-db/prisma/migrations/20240202103529_remove_auth0_preview_account_type_from_enum/migration.sql
@@ -1,0 +1,16 @@
+/*
+  Warnings:
+
+  - The values [Auth0Signup] on the enum `PreviewAccountType` will be removed. If these variants are still used in the database, this will fail.
+
+*/
+-- AlterEnum
+BEGIN;
+CREATE TYPE "PreviewAccountType_new" AS ENUM ('None', 'BreakingTheMonolith');
+ALTER TABLE "Account" ALTER COLUMN "previewAccountType" DROP DEFAULT;
+ALTER TABLE "Account" ALTER COLUMN "previewAccountType" TYPE "PreviewAccountType_new" USING ("previewAccountType"::text::"PreviewAccountType_new");
+ALTER TYPE "PreviewAccountType" RENAME TO "PreviewAccountType_old";
+ALTER TYPE "PreviewAccountType_new" RENAME TO "PreviewAccountType";
+DROP TYPE "PreviewAccountType_old";
+ALTER TABLE "Account" ALTER COLUMN "previewAccountType" SET DEFAULT 'None';
+COMMIT;
--- a/packages/amplication-prisma-db/prisma/migrations/20240202103529_remove_auth0_preview_account_type_from_enum/migration.sql
+++ b/packages/amplication-prisma-db/prisma/migrations/20240202103529_remove_auth0_preview_account_type_from_enum/migration.sql
@@ -6,6 +6,9 @@
 */
 -- AlterEnum
 BEGIN;
+UPDATE "Account"
+SET "previewAccountType" = 'None'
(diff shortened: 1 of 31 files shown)


packages/amplication-prisma-db/prisma/schema.prisma
```diff
@@ -616,5 +616,4 @@ enum CodeGeneratorVersionStrategy {
 enum PreviewAccountType {
   None
   BreakingTheMonolith
-  Auth0Signup
```
abrl91, line 619: This was already merged to master and deployed to production. Are we good with removing it?

packages/amplication-server/src/core/auth/auth.types.ts
abrl91, line 1: 👍

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -386,7 +363,7 @@ export class AuthService {
         previewAccountType,
         previewAccountEmail,
       },
-      identityProvider: `${IDENTITY_PROVIDER_PREVIEW_ACCOUNT}_${previewAccountType}`,
+      identityProvider,
```
abrl91, line 366: I think we still need to add the preview account type, although we have only one now, but refactoring this only when we have another one will change the event name (if I'm not mistaken)

packages/amplication-prisma-db/prisma/schema.prisma
overbit: forgot to push the manual data migration script. Good catch!

overbit: Pushed [6108abc](https://github.com/amplication/amplication/pull/7916/commits/6108abcd06665e4005477a4628f9e8cdfeb6fa52)

packages/amplication-server/src/core/auth/auth.service.ts
overbit: I agree, this is done here at line 356: https://github.com/amplication/amplication/pull/7916/files/25550b92b9c582f87967b2ff0b2f6982414a5387#diff-368a0067447ce7f9a858f275793696cb4a7fc7111e2c14caba62cfb5febf3eddR356

abrl91 (approved):

yuval-hazaz (commented):
@overbit the code changes look OK to me, but I want to make sure the logic remains... 
So, when the user starts the flow, we still doing all the same except of creating the account for the user, and when the flow ends on the other side we anway create the user as in a normal flow - that's right? 

also, please verify that we don't miss any events at the beginning (startSignupWithEmail?) and also, that the new user ends up on the onboarding on the client side by the end of the flow

packages/amplication-prisma-db/prisma/migrations/20240202103529_remove_auth0_preview_account_type_from_enum/migration.sql
mulygottlieb, line 1: Nice!

PR #7927, opened 2024-02-02
chore(server): add complete signup email event

overbit:
Fixes: amplication/private-issues#131

## PR Details

Added new tracking event for CompleteEmailSignup. (step 3 of amplication/private-issues#131)

:warning: This PR depends on id token custom claims implemented here https://github.com/amplication/auth0/pull/14 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/auth/auth.controller.ts
+++ b/packages/amplication-server/src/core/auth/auth.controller.ts
@@ -184,8 +184,7 @@ export class AuthController {
     }
 
     this.authService.trackCompleteEmailSignup(
-      user.id,
-      user.createdAt,
+      user.account,
       profile,
       existingUser
     );
--- a/packages/amplication-server/src/core/auth/auth.controller.ts
+++ b/packages/amplication-server/src/core/auth/auth.controller.ts
@@ -173,6 +173,7 @@ export class AuthController {
       },
     });
 
+    const existingUser = !!user;
     if (!user) {
       user = await this.authService.createUser(profile);
       isNew = true;
@@ -181,6 +182,13 @@ export class AuthController {
       user = await this.authService.updateUser(user, { githubId: profile.sub });
       isNew = false;
     }
+
+    this.authService.trackCompleteBusinessEmailSignup(
+      user.id,
+      user.createdAt,
+      profile,
+      existingUser
+    );
 
     // @todo update the token to include the auth0 expiry / issued at / etc
     const token = await this.authService.prepareToken(user);
--- a/packages/amplication-server/src/core/auth/auth.controller.ts
+++ b/packages/amplication-server/src/core/auth/auth.controller.ts
@@ -183,7 +183,7 @@ export class AuthController {
       isNew = false;
     }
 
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/auth/auth.controller.ts
```diff
@@ -182,6 +183,13 @@ export class AuthController {
       isNew = false;
```
overbit, line 183: Discuss if a new event is needed to detect user uses switches between different login types (github or sso or email+password)

packages/amplication-server/src/core/auth/auth.controller.ts
```diff
@@ -182,6 +183,13 @@ export class AuthController {
...
     }
 
+    this.authService.trackCompleteEmailSignup(
+      user.id,
```
mulygottlieb, line 187: You need to pass the Account.id rather than the User.id (or pass the entire user object and let the `trackCompleteEmailSignup` extract the account) since the identifier for the Analytics is always the Account.id and not the User.id

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -129,6 +130,55 @@ export class AuthService {
...
+    existingUser: boolean
+  ): void {
+    const { identityOrigin, loginsCount } = profile;
+
+    if (loginsCount != 1) {
+      return;
+    }
+
+    const userData: IdentifyData = {
+      userId: userId,
```
mulygottlieb, line 147: Make sure to use the **Account** Id and not the User.Id for Analytics key

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -129,6 +130,55 @@ export class AuthService {
...
+    void this.analytics.identify(userData).catch((error) => {
+      this.logger.error(
+        `Failed to identify user ${userId} in segment analytics`,
+        error
+      );
+    });
+    //we send the userData again to prevent race condition
+    void this.analytics
+      .track({
+        userId: userData.userId,
```
mulygottlieb, line 163: Same. Use the Account.id

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -568,7 +620,6 @@ export class AuthService {
     if (apiToken.count === 1) {
       return true;
     }
-    return false;
```
mulygottlieb, line 571: Don't you need to return false here (there is no return in this flow, unless you mean an implicit false?

mulygottlieb (changes requested):
Please see my comments

packages/amplication-server/src/core/auth/auth.controller.ts
overbit: done! Nice catch!

packages/amplication-server/src/core/auth/auth.service.ts
overbit: Nice one! this change was not on purpose

mulygottlieb (approved):

PR #7943, opened 2024-02-05
refactor(server): move oidc logic to service instead of leaving it in controller

overbit:
Fixes: amplication/private-issues#131

## PR Details

consolidate auth throgh OIDC and JWT generation in auth service.

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/auth/auth.service.ts
+++ b/packages/amplication-server/src/core/auth/auth.service.ts
@@ -1,6 +1,7 @@
 import { Injectable, forwardRef, Inject } from "@nestjs/common";
 import { JwtService } from "@nestjs/jwt";
 import { subDays } from "date-fns";
+import { Response } from "express";
 import { ConfigService } from "@nestjs/config";
 import cuid from "cuid";
 import { Env } from "../../env";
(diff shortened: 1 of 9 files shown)


packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -763,6 +774,59 @@ export class AuthService {
...
+      user = await this.createUser(profile);
+      isNew = true;
+    }
+    if (!user.account.githubId || user.account.githubId !== profile.sub) {
+      user = await this.updateUser(user, { githubId: profile.sub });
+      isNew = false;
+    }
+
+    this.trackCompleteEmailSignup(
+      user.id,
```
mulygottlieb, line 796: Pay attention here that the Anaytics tracking key should be the Account.id and not the User.id

mulygottlieb (changes requested):
See one comment

packages/amplication-server/src/core/auth/auth.service.ts
overbit: done

mulygottlieb (approved):

PR #7957, opened 2024-02-07
Feat/preview flow add schemes

abrl91:
Close: #[issue-number]

## PR Details

Add three new schema options in the preview user flow for BTM (Formbricks, Abby, Papermark) 
Formbricks and Papermark were added instead of NextCrm and Webstudio

image

## PR Checklist


ee/packages/git-sync-manager/src/models.ts
```diff
@@ -836,7 +836,10 @@ export type EnumResourceTypeFilter = {
 };
 
 export enum EnumSchemaNames {
```
mulygottlieb, line 838: This comment is optional for your consideration. I think that EnumSchemaNames is too generic. I would rename it to EnumPreviewUserSchemaNames or any other naming idea that explains what are those schema names.

packages/amplication-server/src/core/dbSchemaImport/dto/EnumSchemaNames.ts
```diff
@@ -2,6 +2,9 @@ import { registerEnumType } from "@nestjs/graphql";
 
 export enum EnumSchemaNames {
```
mulygottlieb, line 3: Do we need to re-define it outside of the models again?

mulygottlieb: I mean can we only do the registerEnumType here and bring the type from models.ts?

mulygottlieb (approved):
Approved but please consider my comments

packages/amplication-server/src/core/dbSchemaImport/dto/EnumSchemaNames.ts
abrl91: It wil not be in the models without having it here with the registerEnumType

PR #7976, opened 2024-02-07
Feat/btm add server client data validations

morhag90:
Close: https://github.com/amplication/private-issues/issues/138
## PR Details

Add server data validation according to the issue description. 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resource.service.ts
+++ b/packages/amplication-server/src/core/resource/resource.service.ts
@@ -1,6 +1,7 @@
 import {
   EnumActionStepStatus,
   RedesignProjectMovedEntity,
+  RedesignProjectNewService,
 } from "@amplication/code-gen-types/models";
 import { Lookup } from "@amplication/code-gen-types/types";
 import { KAFKA_TOPICS } from "@amplication/schema-registry";
@@ -61,7 +62,10 @@ import { UserAction } from "../userAction/dto";
 import { EnumUserActionType } from "../userAction/types";
 import { UserActionService } from "../userAction/userAction.service";
 import { ReservedEntityNameError } from "./ReservedEntityNameError";
-import { REDESIGN_PROJECT_INITIAL_STEP_DATA } from "./constants";
+import {
+  REDESIGN_PROJECT_INITIAL_STEP_DATA,
+  VALIDATE_PROJECT_REDESIGN_CHANGES_DATA,
+} from "./constants";
 import {
   CreateOneResourceArgs,
   FindManyResourceArgs,
@@ -73,6 +77,7 @@ import {
 import { RedesignProjectArgs } from "./dto/RedesignProjectArgs";
 import { ProjectConfigurationExistError } from "./errors/ProjectConfigurationExistError";
 import { EnumRelatedFieldStrategy } from "../entity/dto/EnumRelatedFieldStrategy";
+import { ActionStep } from "../action/dto";
 
 const USER_RESOURCE_ROLE = {
   name: "user",
@@ -88,6 +93,12 @@ export const INVALID_DELETE_PROJECT_CONFIGURATION =
 
 const DEFAULT_PROJECT_CONFIGURATION_DESCRIPTION =
   "This resource is used to store project configuration.";
+
+const SERVICE_LIMITATION_ERROR =
+  "Can not create new services, The workspace reached your plan's resource limitation";
+
+const ENTITIES_PER_SERVICE_LIMITATION_ERROR =
+  "Your current plan permits only one active Service";
 
 export type CreatePreviewServiceArgs = {
   args: CreateOneResourceArgs;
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -89,6 +90,12 @@ export const INVALID_DELETE_PROJECT_CONFIGURATION =
 const DEFAULT_PROJECT_CONFIGURATION_DESCRIPTION =
   "This resource is used to store project configuration.";
 
+const SERVICE_LIMITATION_ERROR =
+  "Can not create new services, The workspace reached your plan's resource limitation";
+
+const ENTITIES_PER_SERVICE_LIMITATION_ERROR =
+  "Your current plan permits only one active Service";
```
mulygottlieb, line 97: The error text here does not seem to fit the "entities per resource" limitation

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -625,16 +668,6 @@ export class ResourceService {
       }
 
       // 3.group moved entities by target resource
```
mulygottlieb, line 670: This comment should pass along with the block that moved up

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+        //service limitation validation
+        const featureServices = await this.billingService.getMeteredEntitlement(
+          project.workspaceId,
+          BillingFeature.Services
+        );
+
+        if (
+          !featureServices.hasAccess ||
+          (!featureServices.isUnlimited &&
+            featureServices.usageLimit <= projectResources.length)
```
mulygottlieb, line 896: If multiple services are added, you are only checking if one can be added (same validation inside the for loop for each service, not taking into account newServices.length)

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+      for (const newService of newServices) {
+        // duplicate name validation
+        const duplicateService = projectResources.find(
+          (resource) =>
+            resource.name.toLocaleLowerCase() ===
+            newService.name.toLocaleLowerCase()
+        );
+        if (duplicateService) {
+          throw new AmplicationError(
+            `Resource : ${newService.name} is already exist in project: ${project.name}.`
```
mulygottlieb, line 884:
```suggestion
            `Resource : ${newService.name} already exists in project: ${project.name}.`
```

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+            id: movedEntity.entityId,
+          },
+        });
+        // authEntity validation
+        if (
+          serviceSettings.authEntityName &&
+          serviceSettings.authEntityName === currentEntity.name
+        ) {
+          throw new AmplicationError(
+            `cannot move auth entity : ${currentEntity.name}.`
```
mulygottlieb, line 941:
```suggestion
            `Cannot move Auth entity : ${currentEntity.name}.`
```

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+        const resourceEntities = await this.entityService.entities({
+          where: {
+            resourceId: movedEntity.targetResourceId,
+            deletedAt: null,
+          },
+        });
+
+        //duplicate entities names validation
+        if (resourceEntities.length > 0) {
+          for (const resourceEntity of resourceEntities) {
```
mulygottlieb, line 954: Consider replace for with find

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+            if (resourceEntity.name === currentEntity.name) {
+              throw new AmplicationError(
+                `Entity : ${currentEntity.name} is already exist in resource: ${movedEntity.targetResourceId}.`
```
mulygottlieb, line 957: Let's discuss if we want to bring the targetResource name here or not

mulygottlieb: IIUC, targetResourceId is the same as resourceId at this point (in the map) and we already get the current resource in the next lines

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -827,6 +860,123 @@ export class ResourceService {
...
+      }
+
+      //pass limitation validation
+      const currentResource = await this.findOne({
+        where: {
+          id: resourceId,
+        },
+      });
+      if (
+        (!currentResource && !project.licensed) ||
```
mulygottlieb, line 971:
```suggestion
        ( !project.licensed) ||
```

mulygottlieb (approved):

Issue #7968, opened 2024-02-07
BTM - better Handle new services creation in the Arch tab


PazYanoverr:
See https://github.com/amplication/amplication/issues/7864 for more context.

1. Enable the two APIs (REST, GQL), and admin UI for any new service created in the Arch tab (Redesign mode)
2. Base folders - make sure the new services do not override prev services folders.
     i. In case the base directory is empty (in the monolith service)- leave it empty 
     ii. Check for the last '/' in the base directory, and replace the string *after* with < service-name > and < service-name >-admin
     iii. In case no '/' - replace all string with < service-name > and < service-name >-admin

abrl91:
[mono-repe]: Test 2 was created during the architecture redesign.

![Image](https://github.com/amplication/amplication/assets/39680385/2af66a13-449e-48d5-b0dd-29d834f1498f)

[poly-repo]: Test 3 poly was created during the architecture redesign.

![Image](https://github.com/amplication/amplication/assets/39680385/0a13d890-c79b-44ad-b909-44ea1b068d5b)

PR #7977, opened 2024-02-08
feat(server): improve serviceSettings =>

morhag90:
Close: #7968 

## PR Details

Update service settings values according to the original resource. 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resource.service.ts
+++ b/packages/amplication-server/src/core/resource/resource.service.ts
@@ -539,16 +539,35 @@ export class ResourceService {
       event: EnumEventType.ArchitectureRedesignApply,
     });
 
+    const originalResourceId = movedEntities[0]?.originalResourceId;
+
+    const originalResourceSettings =
+      await this.serviceSettingsService.getServiceSettingsValues(
+        {
+          where: {
+            id: originalResourceId,
+          },
+        },
+        user
+      );
+
+    const originalResourceAdminPath =
+      originalResourceSettings &&
+      originalResourceSettings.adminUISettings.adminUIPath;
+    const originalResourceServerPath =
+      originalResourceSettings &&
+      originalResourceSettings.serverSettings.serverPath;
+
     const defaultServiceSettings: ServiceSettingsUpdateInput = {
       adminUISettings: {
-        generateAdminUI: false,
+        generateAdminUI: true,
         adminUIPath: "",
       },
       serverSettings: {
-        generateGraphQL: true, //@todo: take value from original service
-        generateRestApi: false, //@todo: take value from original service
+        generateGraphQL: true,
+        generateRestApi: true,
         generateServer: true,
-        serverPath: "", //@todo: take path from original service and use the same base path
+        serverPath: "",
       },
       authProvider: EnumAuthProviderType.Jwt,
     };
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -580,6 +599,30 @@ export class ResourceService {
...
+            0,
+            originalResourceServerPath.lastIndexOf("/") + 1
+          );
+
+        const baseAdminPath =
+          !originalResourceId || !originalResourceAdminPath
+            ? ""
+            : !originalResourceAdminPath.includes("/")
+            ? `${newService.name}-admin`
+            : `${adminPathWithoutLastFolder}${newService.name}-admin`;
```
mulygottlieb, line 618: Did you check we are not losing the "/" between the LastFolder and the newService.name?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -580,6 +599,30 @@ export class ResourceService {
...
+        const baseAServerPath =
+          !originalResourceId || !originalResourceAdminPath
```
mulygottlieb, line 620:
```suggestion
          !originalResourceId || !originalResourceServerPath
```

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -580,6 +599,30 @@ export class ResourceService {
...
+            originalResourceServerPath.lastIndexOf("/") + 1
+          );
+
+        const baseAdminPath =
+          !originalResourceId || !originalResourceAdminPath
+            ? ""
+            : !originalResourceAdminPath.includes("/")
+            ? `${newService.name}-admin`
+            : `${adminPathWithoutLastFolder}${newService.name}-admin`;
+        const baseAServerPath =
```
mulygottlieb, line 619:
```suggestion
        const baseServerPath =
```

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -539,16 +539,35 @@ export class ResourceService {
       event: EnumEventType.ArchitectureRedesignApply,
     });
 
+    const originalResourceId = movedEntities[0]?.originalResourceId;
```
mulygottlieb, line 542: Should we use `resourceId` from line 526 here?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -580,6 +599,30 @@ export class ResourceService {
...
+        const serverPathWithoutLastFolder =
+          originalResourceServerPath?.substring(
+            0,
+            originalResourceServerPath.lastIndexOf("/") + 1
+          );
+
+        const baseAdminPath =
+          !originalResourceId || !originalResourceAdminPath
+            ? ""
+            : !originalResourceAdminPath.includes("/")
```
mulygottlieb, line 616: originalResourceAdminPath might be undefined here?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -580,6 +599,30 @@ export class ResourceService {
...
+            ? `${newService.name}-admin`
+            : `${adminPathWithoutLastFolder}${newService.name}-admin`;
+        const baseAServerPath =
+          !originalResourceId || !originalResourceAdminPath
+            ? ""
+            : !originalResourceServerPath.includes("/")
```
mulygottlieb, line 622: originalResourceServerPath might be undefined here?

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -590,7 +633,25 @@ export class ResourceService {
...
+            serviceSettings: {
+              ...defaultServiceSettings,
+
+              adminUISettings: {
+                adminUIPath: baseAdminPath,
+                generateAdminUI:
+                  defaultServiceSettings.adminUISettings.generateAdminUI,
+              },
+              serverSettings: {
+                serverPath: baseAServerPath,
```
mulygottlieb, line 645:
```suggestion
                serverPath: baseServerPath,
```

mulygottlieb (changes requested):

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @mulygottlieb  no, because of the first check:  !originalResourceId || !originalResourceAdminPath

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @mulygottlieb no, because of the first check: !originalResourceId || !originalResourceAdminPath

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: but I see that I need to update the  to originalResourceServerPath

packages/amplication-server/src/core/resource/resource.service.ts
morhag90: @mulygottlieb no, because it gives us a default value if there is no new entities, we don't want this in this case

mulygottlieb (approved):

PR #7983, opened 2024-02-08
feat(git): update commit message for pull request

### LINKED ISSUE #7879: Update commit's comment
For every code commit by Amplication (see example below), the comment should include a link back to Amplication, with the following text-

"If you're looking to bring in a new data model, tweak existing ones, handle APIs, or spice things up with Amplication plugins, just swing by the Amplication platform at https://app.amplication.com/login. Tell us what you need, and we'll handle all the coding magic for you."

Image

Meyazhagan:
Close #7879 

## PR Checklist


libs/util/git/src/constants.ts
```diff
@@ -1,9 +1,7 @@
 export const accumulativePullRequestTitle =
   "chore(Amplication): Update Generated Code";
 
-export const accumulativePullRequestBody = `This pull request includes multiple commits by Amplication with the latest generated code.
-The pull request may also include commits from several resources and/or projects.
-You can safely merge this pull request to your default branch.
+export const accumulativePullRequestBody = `If you're looking to bring in a new data model, tweak existing ones, handle APIs, or spice things up with Amplication plugins, just swing by the Amplication platform at $link_to_service. Tell us what you need, and we'll handle all the coding magic for you.
```
overbit, line 4: @Meyazhagan be careful as the requirements in the issue state to **add** that text and not replace it.

libs/util/git/src/git-client.service.ts
```diff
@@ -256,6 +260,8 @@ export class GitClientService {
     baseBranch: string;
     repositoryGroupName?: string;
     pullRequestTitle?: string;
+    resourceName?: string;
+    resourceUrl?: string;
```
overbit, line 264: It would make more sense to pass the `pullRequestBody` already manipulated to this function as in case that changes, the logic won't be flexible enough.

libs/util/git/src/git-client.service.ts
```diff
@@ -332,7 +340,10 @@ export class GitClientService {
         repositoryName,
         repositoryGroupName,
         pullRequestTitle: `${accumulativePullRequestTitle} ${pullRequestTitle}`,
-        pullRequestBody: accumulativePullRequestBody,
+        pullRequestBody: accumulativePullRequestBody.replace(
+          "$link_to_service",
+          `[${resourceName}](${resourceUrl})`
+        ),
```
overbit, line 346: move the logic at the parent function for a more flexible result

overbit (changes requested):

Meyazhagan:
@overbit I updated the code

libs/util/git/src/git-client.service.ts
```diff
@@ -224,11 +227,13 @@ export class GitClientService {
...
             branchName,
             commitMessage,
-            pullRequestBody,
             preparedFiles,
             baseBranch,
             repositoryGroupName,
             pullRequestTitle,
+            pullRequestBody:
+              getServicePullRequestMessage(resourceName, resourceUrl) +
+              pullRequestBody,
```
overbit, line 236: @Meyazhagan sorry, I was probably too vague. I meant that the change should not live in the `util-git` library at all. 
My expectation is to see only changes in the services.

overbit (changes requested):

libs/util/git/src/constants.ts
```diff
@@ -16,3 +16,13 @@ The Amplication Team`;
...
+
+export const getServicePullRequestMessage = (
+  serviceName: string | undefined,
+  serviceLink: string | undefined
+) => {
+  return `
+  If you're looking to bring in a new data model, tweak existing ones, handle APIs, or spice things up with Amplication plugins, just swing by the Amplication platform at [${serviceName}](${serviceLink}). Tell us what you need, and we'll handle all the coding magic for you.
+  
+  `;
+};
```
overbit, line 28: @Meyazhagan as mention previously, in order to make this feature more flexible and keeping the util-git library generic and content agnostic, we would not expect any change to the util-git library in this PR. Please move this logic to the appropriate place (git-sync-manager).

overbit (changes requested):

Meyazhagan:
@overbit @mulygottlieb can you review this PR?

Issue #7960, opened 2024-02-07
BTM - Validation on moving auth entity  (client)


PazYanoverr:

mulygottlieb:
Passed QA take 2:


![Image](https://github.com/amplication/amplication/assets/1912933/5872ce14-3e14-4293-ba45-556d47228b0e)



@PazYanoverr

PR #7989, opened 2024-02-11
feat(client): add auth-entity validation

morhag90:
Close: #7960 

## PR Details

Add auth-entity validation when trying to move it to another service. 
## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Project/ArchitectureConsole/hooks/useModelOrganizer.tsx
+++ b/packages/amplication-client/src/Project/ArchitectureConsole/hooks/useModelOrganizer.tsx
@@ -76,8 +76,7 @@ const useModelOrganization = ({ projectId, onMessage }: Props) => {
   const [duplicateEntityError, setDuplicateEntityError] =
     useState<boolean>(false);
 
-  const [duplicateEntityErrorMessage, setDuplicateEntityErrorMessage] =
-    useState<string>(null);
+  const [errorMessage, setErrorMessage] = useState<string>(null);
 
   const [userAction, setUserAction] = useState<models.UserAction>(null);
   const { data: applyChangesResults } = useUserActionWatchStatus(userAction);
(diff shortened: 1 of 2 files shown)


packages/amplication-client/src/Project/ArchitectureConsole/hooks/useModelOrganizer.tsx
```diff
@@ -577,11 +583,25 @@ const useModelOrganization = ({ projectId, onMessage }: Props) => {
...
         ) {
+          const baseErrorMessage = `Cannot move entity to service: ${currentTargetResource.data.payload?.name}`;
           currentNode.parentNode = currentNode.data.originalParentNode;
+          if (authEntity) {
+            setDuplicateEntityErrorMessage(
+              `${baseErrorMessage} because the authentication entity cannot be deleted.`
+            );
+          } else {
+            setDuplicateEntityErrorMessage(
+              `${baseErrorMessage} because the entity name already exists.`
```
yuval-hazaz, line 602: can we use a logic a head of time and prevent the node from being moved? 
we can use the node.draggable= false 
we can keep this logic anyway if needed

packages/amplication-client/src/Project/ArchitectureConsole/hooks/useModelOrganizer.tsx
```diff
@@ -73,6 +76,9 @@ const useModelOrganization = ({ projectId, onMessage }: Props) => {
   const [duplicateEntityError, setDuplicateEntityError] =
     useState<boolean>(false);
 
+  const [duplicateEntityErrorMessage, setDuplicateEntityErrorMessage] =
```
yuval-hazaz, line 79: let's just call this ErrorMessage so we can use this for any future message (it is already not just used for duplicate.
Also, I don't think we need two state  - it is enough to use the "ErrorMessage" as string | null

yuval-hazaz (changes requested):

yuval-hazaz (approved):

Issue #7395, opened 2023-11-08
Update react-admin in admin-ui (dsg) to v4


Shurtu-gal:
### Feature description

Currently, the react-admin version is [3.19.12](https://github.com/amplication/amplication/blob/b0eabb424b127dacdfc91977e880aa243541b21b/packages/data-service-generator/src/admin/package-json/package.json#L13). However some things have been [depreciated in v4 of react-admin](https://marmelab.com/react-admin/Upgrade.html) hence the need to upgrade.

### Use case

I was trying to make a demo for WebSockets to be implemented in plugin, in which there are some conflicts and breaking changes b/w major versions.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

Meyazhagan:
@Shurtu-gal can I take this

Shurtu-gal:
Sure @Meyazhagan.

mulygottlieb:
Closing the issue for now since this is not where we want to focus our efforts.

PR #7998, opened 2024-02-13
feat(dsg): upgrade react-admin to v4

Meyazhagan:
Close: #7395 

## PR Checklist


overbit:
Thanks @Meyazhagan! we will review it ASAP.

Shurtu-gal:
@overbit should we test if all plugins work as expected after this version bump?

overbit:
@Shurtu-gal yep, that will be necessary. Enterprise users currently have the ability to select the DSG version so they could stay with the current plugin versions but all others will need to have all the latest official plugin versions fully working with react admin 4.

Special attention might be required for the Auth plugins as with the latest react admin, there are changes to the AuthProvider interface and hooks.

@mulygottlieb this needs to be taken into consideration

overbit (changes requested):

packages/data-service-generator/src/admin/entity/create-field-value.ts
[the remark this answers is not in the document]
Meyazhagan: yeah sure

packages/data-service-generator/src/admin/entity/entity-list-component/entity-list-component.template.tsx
```diff
@@ -1,20 +1,14 @@
...
+export const ENTITY_LIST = (): React.ReactElement => {
   return (
-    <List
-      {...props}
-      bulkActionButtons={false}
-      title={ENTITY_PLURAL_DISPLAY_NAME}
-      perPage={50}
-      pagination={<Pagination />}
-    >
+    <List>
```
Shurtu-gal, line 11: Are the list props no longer needed ?

Shurtu-gal (changes requested):

packages/data-service-generator/src/admin/entity/entity-list-component/entity-list-component.template.tsx
Meyazhagan: `bulkActionButtons` is deprecated. we can have rest of props.

Meyazhagan: I saw in pagination component, we extending rowsPerPageOptions, so I revert that change

Meyazhagan: > Are the list props no longer needed ?

yes

mulygottlieb:
@Meyazhagan  @Shurtu-gal @overbit I am sorry, but I'm not sure why we are investing time in this.
We focus on backend services generation, and the react admin UI is a nice-to-have utility. Spending time on this + adding the potential risk of breaking existing plugins and requiring significant regression testing doesn't make sense to me.

PR #8008, opened 2024-02-14
test(server): add and fix test based on new requirements

abrl91:
Close: https://github.com/amplication/private-issues/issues/142

## PR Details

Refactor `prepareBtmRecommendations` function:
This function prepares the AI recommendation for the Break the Monolith result
   * It filters out the tables that the AI result has that the original resource doesn't have
   * It makes sure that there are no duplicated tables in the microservices by removing the duplicates and putting them on the microservice with the least amount of tables
   * It makes sure that the original resource will be added to the microservices result with tables that the AI missed
   * It makes sure that the result will not includes microservices with no tables

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
+++ b/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
@@ -277,6 +277,16 @@ describe("ResourceBtmService", () => {
       const expectedResult: BreakServiceToMicroservicesData = {
         microservices: [
           {
+            name: "product",
+            functionality: "manage products",
+            tables: [
+              {
+                name: "product",
+                originalEntityId: "product",
+              },
+            ],
+          },
+          {
             name: "order",
             functionality: "manage orders, prices and payments",
             tables: [
@@ -287,16 +297,6 @@ describe("ResourceBtmService", () => {
               {
                 name: "orderItem",
                 originalEntityId: "orderItem",
-              },
-            ],
-          },
-          {
-            name: "product",
-            functionality: "manage products",
-            tables: [
-              {
-                name: "product",
-                originalEntityId: "product",
               },
             ],
           },
@@ -459,6 +459,16 @@ describe("ResourceBtmService", () => {
       const expectedResult: BreakServiceToMicroservicesData = {
         microservices: [
           {
+            name: "product",
+            functionality: "manage products",
+            tables: [
+              {
+                name: "product",
+                originalEntityId: "product",
+              },
+            ],
+          },
+          {
             name: "order",
             functionality: "manage orders, prices and payments",
             tables: [
@@ -469,16 +479,6 @@ describe("ResourceBtmService", () => {
               {
                 name: "orderItem",
                 originalEntityId: "orderItem",
-              },
-            ],
-          },
-          {
-            name: "product",
-            functionality: "manage products",
-            tables: [
-              {
-                name: "product",
-                originalEntityId: "product",
               },
             ],
           },
@@ -783,6 +783,16 @@ describe("ResourceBtmService", () => {
       const expectedResult: BreakServiceToMicroservicesData = {
         microservices: [
           {
+            name: "product",
+            functionality: "manage products",
+            tables: [
+              {
+                name: "product",
+                originalEntityId: "product",
+              },
+            ],
+          },
+          {
             name: "order",
             functionality: "manage orders, prices and payments",
             tables: [
@@ -793,16 +803,6 @@ describe("ResourceBtmService", () => {
               {
                 name: "orderItem",
                 originalEntityId: "orderItem",
-              },
-            ],
-          },
-          {
-            name: "product",
-            functionality: "manage products",
-            tables: [
-              {
-                name: "product",
-                originalEntityId: "product",
               },
             ],
           },
--- a/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
+++ b/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
@@ -128,7 +128,7 @@ describe("ResourceBtmService", () => {
     });
   });
 
-  describe("translateToBtmRecommendation", () => {
+  describe("prepareBtmRecommendations", () => {
     it("should map the prompt result to a btm recommendation", async () => {
       const promptResult: BreakTheMonolithOutput = {
         microservices: [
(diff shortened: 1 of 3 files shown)


packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -211,44 +221,77 @@ export class ResourceBtmService {
...
+            originalResourceEntityNamesSet.has(tableName) &&
+            !isDuplicatedAlreadyUsed
+          );
+        }),
+      }))
+      .filter((microservice) => microservice.tables.length > 0);
+
+    // add the original resource to the microservices
+    promptResultObj.microservices.push({
+      name: `${originalResource.name}Original`,
```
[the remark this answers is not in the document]
mulygottlieb: This is already almost the needed collision handling - just instead of adding "original" to the original one, add "-new" to the new one...

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -211,44 +221,77 @@ export class ResourceBtmService {
...
+
+    // remove duplicates and put on the microservice with the least amount of tables
+    promptResultObj.microservices = promptResultObj.microservices
+      .sort((a, b) => a.tables.length - b.tables.length)
+      .map((microservice) => ({
+        name: microservice.name,
+        functionality: microservice.functionality,
+        tables: microservice.tables.filter((tableName) => {
+          const isDuplicatedAlreadyUsed = usedDuplicatedEntities.has(tableName);
+          if (duplicatedEntities.has(tableName)) {
```
mulygottlieb, line 258:
```suggestion
          if (!isDuplicatedAlreadyUsed && duplicatedEntities.has(tableName)) {
```

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -211,44 +221,77 @@ export class ResourceBtmService {
...
+            (map, entity) => {
+              map[entity.name] = entity;
+              return map;
+            },
+            {} as Record<string, EntityDataForBtm>
+          );
+
+          return {
+            name: tableName,
+            originalEntityId: entityNameIdMap[tableName]?.id,
```
mulygottlieb, line 291: I wonder if the '?' here is needed and if it's in fact confusing since I believe we expect that *every* table will exist in the map if there are not bugs, right?

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -211,44 +221,77 @@ export class ResourceBtmService {
     const usedDuplicatedEntities = new Set<string>();
 
     const originalResource = await this.getResourceDataForBtm(resourceId);
-    const originalResourceEntitiesSet = new Set(
-      originalResource.entities.map((entity) => entity.name)
+    const originalResourceEntityNames = originalResource.entities.map(
```
mulygottlieb, line 224: Not sure why you separated the line to have both `originalResourceEntityNames` and `originalResourceEntityNamesSet` since you only need the Set (in `findUniqueEntities` you turn it into a set again and only use the set)

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -276,12 +319,36 @@ export class ResourceBtmService {
     );
   }
 
+  findUniqueEntities(
```
mulygottlieb, line 322: @abrl91 why not use Set.difference() function which is meant for this exactly (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/difference) ?

packages/amplication-server/src/core/resource/resourceBtm.service.ts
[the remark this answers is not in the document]
mulygottlieb: @abrl91 after talking to Yuval about his implementation of moving the AI results into the client' ModelOrganizer, it seems that adding the original service into the microservices property is not needed for the client (which already has the original service anyway) and in fact adding it in this way will break Yuval's implementation. So, what I suggest, is to augment the promptResultObj with a **new** property `originalService` which will have the same structure as the other microservices (name, functionality, tables) but will be outside of the `microservices` property. It will allow the clients of the AI result to easily distinguish the original service from the new ones, and decide per usecase what to do with it.

packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
```diff
@@ -324,11 +345,17 @@ describe("ResourceBtmService", () => {
             functionality: "manage orders, prices and payments",
             tables: ["order", "orderItem"],
           },
+          {
```
overbit, line 348: Review this test as the title and the expectations are not aligned

abrl91: https://github.com/amplication/amplication/pull/8008/commits/25259309dccd5010d780844262e478c4b40cf1e9

mulygottlieb (approved):

packages/amplication-server/src/core/resource/resourceBtm.service.ts
overbit, line 1: `findDuplicatedEntities` is O(n^2) since the last PR, I'd suggest to improve this as it is an easy problem that can be implemented with O(n) complexity

overbit:
```ts
    const duplicates = new Set<string>();
    const seen = new Set<string>();

    for (const str of strings) {
        if (seen.has(str)) {
            duplicates.add(str);
        } else {
            seen.add(str);
        }
    }

    return duplicates;
```

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -211,29 +220,37 @@ export class ResourceBtmService {
...
       microservices: promptResultObj.microservices
-        .sort((microservice) => -1 * microservice.tables.length)
+        .sort((a, b) => a.tables.length - b.tables.length)
         .map((microservice) => ({
           name: microservice.name,
           functionality: microservice.functionality,
           tables: microservice.tables
-            .filter((dataModelName) => {
+            .filter((tableName) => !inventedEntitiesByGpt.includes(tableName))
+            .filter((tableName) => {
```
overbit, line 239: no need for 2 filters right? you should be able to do everything in a single run

overbit (approved):

Issue #7974, opened 2024-02-07
BTM (AI)- Handle service name collisions from suggested services VS existing services


PazYanoverr:
We should handle name collisions that may occur as part of the BTM flow-

Service names- the LLM process of BTM may cause a name collision (between services)-
In case the name of a new service suggested by OpenAI already exists, the new service should be called [name]-< random suffix >
(The BTM spec was updated accordingly)

overbit:
Test result ✅ 



https://github.com/amplication/amplication/assets/2861984/2687df42-a463-482e-983b-6068b51e07ff

PR #8017, opened 2024-02-16
test(server): add test for the new collision handling

abrl91:
Close: https://github.com/amplication/amplication/issues/7974

## PR Details

- handle microservice name potential collisions between GPT recommendation and the exiting services in the project
- refactor the prepare method

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
+++ b/packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
@@ -16,6 +16,11 @@ import { BillingService } from "../billing/billing.service";
 import { EnumSubscriptionPlan } from "../subscription/dto";
 import { EnumEventType } from "../../services/segmentAnalytics/segmentAnalytics.types";
 import { MockedSegmentAnalyticsProvider } from "../../services/segmentAnalytics/tests";
+import { v4 as uuidv4 } from "uuid";
+
+jest.mock("uuid", () => ({
+  v4: jest.fn().mockReturnValue("123456"),
+}));
 
 const resourceIdMock = "resourceId";
 const userIdMock = "userId";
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -195,11 +197,28 @@ export class ResourceBtmService {
...
+  handleProjectServicesCollision(
+    projectServices: { name: string }[],
+    serviceName: string
+  ): string {
+    let validServiceName = serviceName;
+    projectServices.forEach((service) => {
+      const isDuplicated = service.name === serviceName;
+
+      if (isDuplicated) {
+        const suffix = v4().split("-")[0];
```
mulygottlieb, line 209: The ticket says the service name in case of collision should be suffixes with "_new", so why are you using a random suffix here?

overbit: I guess to avoid collision with multiple runs of btm by ai. I like the randomised name more because it also looks more professional and less hacky

mulygottlieb: I am perfectly fine with this. It just needs to be agreed with Paz since it's a product spec/requirement. If he is good with the randomized solution, I agree.

packages/amplication-server/src/core/resource/resourceBtm.service.spec.ts
```diff
@@ -813,6 +820,213 @@ describe("ResourceBtmService", () => {
...
+          },
+          {
+            name: "customer",
+            functionality: "manage customer",
+            tables: ["customer"],
+          },
+        ],
+      };
+      const originalResource: ResourceDataForBtm = {
+        name: "order",
```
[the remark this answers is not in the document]
abrl91: This is the original service name.
The GPT results return a microservice with this name (that may/may not be the original one).
When it comes to collision, as we are not preserving the original resource (the one we break) in the server, it doesn't matter if the original resource name is in the GPT result.
What's important is that the `resourceFindManyMock` returns two services, one named "resourceName", and one named "product".

packages/amplication-server/src/core/resource/resourceBtm.service.ts
abrl91: I talked with Paz and he confirmed this implementation

Issue #7965, opened 2024-02-07
BTM - Discarding pending changes (trash bin) while *inside* the Arch screen


PazYanoverr:
When discarding pending changes (trash bin) while *inside* the Arch screen *and* not inside redesign mode - refresh the state (fetch)

overbit:
Test result ✅

PR #8049, opened 2024-02-19
feat(client): reset changes when discarding changes inside BTM

morhag90:
Close: #7965 

## PR Details

Reset changes when discarding changes inside BTM.

## PR Checklist


packages/amplication-client/src/Project/ArchitectureConsole/ModelOrganizer.tsx
```diff
@@ -59,7 +59,13 @@ const edgeTypes = {
 };
 
 export default function ModelOrganizer() {
-  const { currentProject } = useAppContext();
+  const { currentProject, resetPendingChangesIndicator } = useAppContext();
+
+  useEffect(() => {
+    if (!resetPendingChangesIndicator) return;
+    resetChanges();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
overbit, line 67: Is this on purpose?
Shall we add resetChanges to the deps?

morhag90: @overbit  Yes, It is; it causes an infinity loop. this is just a callback function that reset all the changes.

Issue #7967, opened 2024-02-07
BTM - Entity selection context menu - Select all related entities (first relatives)


PazYanoverr:
**NOTE: We need to plan and estimate this before starting, since the decision on implementing or not will depend on the estimation**

Add this part of the design / spec.
- [ ] Add the context menu on entities (the "...")
- [ ] In the context menu add a single option "Select all related entities"
- [ ] If clicked, all *first order* related entities will be in selected mode
- [ ] The user can (as usual) add/remove selection with command key and click

overbit:
@PazYanoverr what is the expectation for Win users? 
`cmd` + `left mouse click` is only available for Mac. 
Is the `control` + `click` the correct mapping?

📝 Suggestion:  It would make sense to have a legend with these common user interaction commands

PazYanoverr:
@overbit - For Win users- Ctrl + Click

overbit:
Test result: ✅ 
https://jam.dev/c/f73d3b88-ea34-4ad1-a717-d92bea9cdb80

PR #8061, opened 2024-02-20
Feat/7967 fix context menu branch

morhag90:
Close: #7967 

## PR Details

Add a context menu for each resource in the Architecture tab; include one option: select all related entities.

## PR Checklist


packages/amplication-client/src/Components/EntityContextMenuButton.tsx
```diff
@@ -0,0 +1,77 @@
...
+const CLASS_NAME = "entity-context-menu-button";
+
+type Props = {
+  isContextMenuEnable: boolean;
+  onSelectRelatedEntities: () => void;
+};
+
+export default function EntityContextMenuButton({
+  onSelectRelatedEntities,
+  isContextMenuEnable,
```
mulygottlieb, line 26: Minor, but I suggest to change to `isContextMenuEnabled` (add "d" at the end)

packages/amplication-client/src/Project/ArchitectureConsole/ModelOrganizer.tsx
```diff
@@ -349,6 +359,7 @@ export default function ModelOrganizer({ restrictedMode = false }: Props) {
                 onNodeDragStop={onNodeDragStop}
                 onEdgesChange={onEdgesChange}
                 connectionMode={ConnectionMode.Loose}
+                onNodeClick={onNodeClick}
```
mulygottlieb, line 362: Is this only effective for the entity nodes and not the resource nodes?

mulygottlieb (approved):

packages/amplication-client/src/Project/ArchitectureConsole/ModelOrganizer.tsx
morhag90: @mulygottlieb is effective for Both of them, but in the callback function, there is a check that prevents the process from continuing,

PR #8093, opened 2024-02-21
chore(server): improve error handling in ai suggestion flow

overbit:
Fixes: https://github.com/amplication/private-issues/issues/155

## PR Details

- Fix issue related to storing of gpt responses. These responses are string and they should not be stored as parsed JSON (otherwise errors will happen)
- Improve error message for bad formatted ai suggestions
- refactor resourcesBtm service to not use userActionService dependency

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resourceBtm.service.ts
+++ b/packages/amplication-server/src/core/resource/resourceBtm.service.ts
@@ -28,7 +28,6 @@ import { BillingFeature } from "@amplication/util-billing-types";
 import { EnumResourceType } from "./dto/EnumResourceType";
 import { v4 } from "uuid";
 import { BREAK_THE_MONOLITH_AI_ERROR_MESSAGE } from "./constants";
-import { validate } from "class-validator";
 
 @Injectable()
 export class ResourceBtmService {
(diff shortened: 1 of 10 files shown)


packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -165,16 +160,20 @@ export class ResourceBtmService {
...
-      originalResourceId: resourceId,
-      data: recommendations,
-    };
+      return {
+        status: EnumUserActionStatus.Completed,
+        originalResourceId: resourceId,
+        data: recommendations,
+      };
+    } catch (error) {
+      throw new AmplicationError(BREAK_THE_MONOLITH_AI_ERROR_MESSAGE);
```
mulygottlieb, line 175: Let's log the original `error` object here to our login infra so we don't lose it

packages/amplication-server/src/core/resource/resourceBtm.service.ts
```diff
@@ -360,7 +359,7 @@ export class ResourceBtmService {
         })),
       };
     } catch (error) {
-      throw new GptBadFormatResponseError(JSON.stringify(promptResult), error);
+      throw new GptBadFormatResponseError(promptResult);
```
mulygottlieb, line 362: Please add the original `error` that was removed so we get it logged and don't lose information

packages/amplication-server/src/core/resource/resourceBtm.service.ts
overbit: done

mulygottlieb (approved):

Issue #8080, opened 2024-02-21
BTM (Preview) - After signup - add an explanation message


PazYanoverr:
After the preview user signups to Amplication and login (after reset password flow), the following message should appear **in a modal** on top of the first screen he sees = project overview page:

**Title**: Welcome Aboard! 🚀 

Congrats on joining us! With Amplication, you've unlocked streamlined backend code generation and boundless possibilities.

Now, let's dive into your application journey. Explore our intuitive interface, craft solutions effortlessly, and enjoy a free 14-day trial with unlocked Enterprise features.

Embrace the adventure ahead, and let's create magic together! ✨

PazYanoverr:
@mulygottlieb - Updated the description. Please assign it to the team.

abrl91:
@PazYanoverr 
The first screen the user sees is the schema list page

PazYanoverr:
@abrl91 - NSIU which page is it. The Project overview?

abrl91:
✅ passed QA


![Image](https://github.com/amplication/amplication/assets/39680385/8047365d-e4cf-49b2-add0-5f6672b3e068)

PR #8137, opened 2024-02-25
Feat/8080 => Add Dialog for Preview user first login after a signup

morhag90:
Close: #8080

## PR Details

 Add Dialog for Preview user first login after a signup. 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Project/ProjectPage.scss
+++ b/packages/amplication-client/src/Project/ProjectPage.scss
@@ -1,7 +1,9 @@
 @import "../style/index.scss";
+$line-height: 24px;
+
 .project-page {
   .amp-text {
-    line-height: 24px;
+    line-height: $line-height;
   }
 }
 
--- a/packages/amplication-client/src/Project/ProjectPage.scss
+++ b/packages/amplication-client/src/Project/ProjectPage.scss
@@ -1,5 +1,8 @@
 @import "../style/index.scss";
 .project-page {
+  .amp-text {
+    line-height: 24px;
+  }
 }
 
 .delete__panel {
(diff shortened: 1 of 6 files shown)


packages/amplication-client/src/Project/ProjectPage.scss
```diff
@@ -1,5 +1,8 @@
 @import "../style/index.scss";
 .project-page {
+  .amp-text {
+    line-height: 24px;
```
mulygottlieb, line 4: This should be in a $ variable according to our guidelines :)

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -763,11 +764,10 @@ export class AuthService {
     let isNew: boolean;
     const existingUser = !!user;
 
+    const isFromPreviewUser =
+      user.account.previewAccountType !== EnumPreviewAccountType.None;
```
mulygottlieb:
```suggestion
      user && user.account.previewAccountType !== EnumPreviewAccountType.None;
```

packages/amplication-server/src/core/auth/auth.service.ts
```diff
@@ -763,11 +764,10 @@ export class AuthService {
...
     // to complete the preview account signup, after the user reset the password in Auth0 we need to update the account type and workspace subscription
-    if (
-      user &&
-      user.account.previewAccountType !== EnumPreviewAccountType.None
-    ) {
+    if (user && isFromPreviewUser) {
```
mulygottlieb:
```suggestion
    if (isFromPreviewUser) {
```

morhag90: @mulygottlieb Already updated :)

mulygottlieb (approved):

Issue #8167, opened 2024-03-04
[Epic] As a User - I want to have an Analytics Dashboard to see the value I get from Amplication over time


mulygottlieb:
Notion Spec:
https://www.notion.so/amplication/Analytics-Dashboard-5285374805704dfea58395db4a2156d6?pvs=4

**Technical Requirements:**
- A new service - Analytics Service - to store the relevant analytical information, aggregate it as needed, and expose it to the client with GQL
- The new service will be a dogfooding service created with Amplication. We can start with a regular Postgres DB or consider if there is any benefit to MongoDB for this use case
- The new service will listen to various Kafka topics that will allow it to update its internal DB with analytics for required metrics
- The actual metrics should be easily extendible (so in the future we will be able to hold new metrics easily)
- Metrics should be saved with timestamp (e.g. `createdAt`), and be linked to a `resourceId` (and perhaps also `projectId` and `workspaceId`) so we can collect all metric of a certain "metric type" for a given resource/project/workspace over a specified time range
- Metrics (where applicable) should be stored as numbers so we can perform numerical operations like sum, max, avg, median, etc.
- Required metrics to hold:
   - Number of Builds (builds over time)
   - Number of LOC per build
   - Number of entities changed/created
   - Number of APIs (ModuleAction) changed/created per build (only "enabled" ones? all? count enabling/disabling itself as something?)
   - Number of Plugins installed / updated
  

**Concerns that need to be discussed:**
- Initial data migration - do we create all the historical data or this dashboard will only contain data starting from the date it was introduced (similar question will arise in the future for new metrics we will add)
- If we calculate LOC changed in a PR (Amplication Build), and the user does **not** merge the PR, and then creates another build - we may count the same LOC twice - do we care? can this be considered a "known limitation" or even a "by design feature" (if I add and remove the same thing and build in between - perhaps I am actually touching more LOC)
- How do we enforce permissions for this data in the new service? Is the new service accessed directly from the client, or relayed via the server?  
- Do we count/represent changes that were done in the platform **without** building or just save data for each new build? E.g. I may be working a lot with the system and changing things but not pushing new PRs

overbit:
Test result: ❔ 

- Unnecessary newline
<img width="720" alt="image" src="https://github.com/amplication/amplication/assets/2861984/3e9925e6-3e33-4378-95bb-a1c6cccdb708">
-

PR #8176, opened 2024-03-06
Feat(server): analyze Amplication benefit

abrl91:
Close: https://github.com/amplication/amplication/issues/8167

## PR Checklist


ee/packages/git-sync-manager/src/pull-request/pull-request.controller.ts
```diff
@@ -85,10 +85,11 @@ export class PullRequestController {
...
 
     try {
-      const pullRequest = await KafkaPacemaker.wrapLongRunningMethod<string>(
-        context,
-        () => this.pullRequestService.createPullRequest(validArgs)
-      );
+      const { pullRequestUrl, diffStat } =
+        await KafkaPacemaker.wrapLongRunningMethod<{
+          pullRequestUrl: string;
+          diffStat: any;
```
mulygottlieb, line 91: If diffStat is actually a string, why use `any`?

abrl91: probably a leftover. It should be a string here

packages/amplication-server/src/core/analytics/types.ts
```diff
@@ -0,0 +1,19 @@
+export enum EnumTimeFrame {
+  LastMonth = "LAST_MONTH",
+  LastThreeMonths = "LAST_THREE_MONTHS",
+  LastYear = "LAST_YEAR",
+  Custom = "CUSTOM",
+}
+
+export type TimeFrame = {
```
mulygottlieb, line 8: I think it's confusing that the type `TimeFrame` has a property `timeFrame`

packages/amplication-server/src/core/analytics/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  timeFrame: EnumTimeFrame;
+  startDate?: Date;
+  endDate?: Date;
+};
+
+export type ProjectBuildsArgs = {
```
mulygottlieb, line 14: Why does this have `startDate` and `endDate` and not `TimeFrame` ?

packages/amplication-server/src/core/analytics/analytics.service.ts
```diff
@@ -0,0 +1,63 @@
...
+    workspaceId,
+    projectId,
+    startDate,
+    endDate,
+  }: ProjectBuildsArgs): Promise<number> {
+    const aggregatedLoc = await this.prisma.build.aggregate({
+      where: {
+        createdAt: {
+          gte: startDate,
+          lte: endDate,
```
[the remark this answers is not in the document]
abrl91: discussed in a call

packages/amplication-server/src/core/analytics/types.ts
abrl91: I need to remove the time frame enum and type

packages/amplication-server/src/core/analytics/types.ts
abrl91: I need to remove the time frame enum and type

packages/amplication-server/src/core/UsageInsights/dtos/BlockChanges.args.ts
mulygottlieb, line 1: If I understand correctly, this type is only used internally in the server and is not part of the args to any GQL query - so I don't think we need all the GQL decorators like ArgsType and Field.

packages/amplication-server/src/core/UsageInsights/usageInsights.service.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  async countBlockChanges({
+    projectIds,
+    startDate,
+    endDate,
+    blockType,
+    timeGroup,
+  }: BlockChangesArgs): Promise<UsageInsights> {
+    let results: QueryRawResult[];
+    switch (blockType) {
```
mulygottlieb, line 106: It's not critical, but I still feel we can use a mapper from BlockType to string somehow and avoid duplication here.

packages/amplication-server/src/core/UsageInsights/usageInsights.service.ts
```diff
@@ -0,0 +1,230 @@
...
+
+  private async evaluateCodeQuality(linesOfCode: number) {
+    const multiplier = 14;
+    const divisor = 1000;
+    const bugsPrevented = (multiplier * linesOfCode) / divisor;
+
+    return Math.round(bugsPrevented);
+  }
+
+  private translateToAnalyticsResults(results: QueryRawResult[]) {
```
mulygottlieb, line 214: Here again, why do we need this - is it just to rename time_group to timeGroup? Can we avoid the need for this translation?

abrl91: The parsing later is not only for the name, but also for type. Moreover, one of the inputs of the function named timeGroup

packages/amplication-server/src/core/UsageInsights/usageInsights.service.ts
abrl91: As I told you, it didn't work. I can show you what I did, maybe I missed something

packages/amplication-server/src/core/UsageInsights/types.ts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+export type ParsedQueryRowResult = {
+  year: number;
+  month: number;
+  timeGroup: number;
+  count: number;
+};
+
+export type QueryRawResult = {
```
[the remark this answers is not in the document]
abrl91: Because it's different. I don't know why but the type of count from the query raw is big int

mulygottlieb (approved):

abrl91:
was implemented in: https://github.com/amplication/amplication/pull/8176

Issue #8180, opened 2024-03-08
🐛 Bug Report: Generated code fails to run after release of TypeScript 5.4


abrl91:
### What happened?

Getting the following errors:
![image](https://github.com/amplication/amplication/assets/39680385/022a9da2-9bbc-4196-ba99-ee3ca23236a4)

### What you expected to happen

To work 😄 

### How to reproduce

`cd {service-name} && npm i && npm run build`

overbit:
Test result ✅

abrl91:
Passed QA ✅ 



![Image](https://github.com/amplication/amplication/assets/39680385/8fdea4b1-e7e8-4e6e-b92b-016a1ad4a761)



`npm run run build` finished successfully.

![Image](https://github.com/amplication/amplication/assets/39680385/32e53eaf-84e0-4aca-ab1e-790680a0e704)

PR #8181, opened 2024-03-08
hotfix(dsg): ts version

abrl91:
Close: https://github.com/amplication/amplication/issues/8180

## PR Details

pin ts version in the generated app: ` "typescript": "5.3.*"`

## PR Checklist

### DIFF
--- a/packages/data-service-generator/src/server/package-json/package.json
+++ b/packages/data-service-generator/src/server/package-json/package.json
@@ -64,7 +64,7 @@
     "prisma": "^5.4.2",
     "supertest": "^6.3.3",
     "ts-jest": "^29.1.1",
-    "typescript": "5.2.2"
+    "typescript": "5.3.*"
   },
   "jest": {
     "preset": "ts-jest",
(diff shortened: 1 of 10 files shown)


mulygottlieb (dismissed):

packages/data-service-generator/src/server/package-json/package.json
```diff
@@ -64,7 +64,7 @@
     "prisma": "^5.4.2",
     "supertest": "^6.3.3",
     "ts-jest": "^29.1.1",
-    "typescript": "^5.2.2"
```
overbit, line 67: This would potentially bring breaking changes to our users generated code as it potentially force them to an earlier version of TS.

The real issue is related to the types mapping we generate that are not

overbit (changes requested):

packages/data-service-generator/src/server/package-json/package.json
abrl91: @overbit we tested it and it works both on run and build.
The problematic version is `5.4.2` and the chance that some user has custom code that relies on that version is very minor.

overbit (approved):

mulygottlieb (approved):

Issue #7713, opened 2023-12-31
[Epic] Custom Actions and Types- Phase 2


PazYanoverr:
Spec-
https://www.notion.so/amplication/Custom-Types-Actions-Phase-2-b83fb9234c78419584bbb47a0fc31634?pvs=4
Design- 
TBD

**P0 issues (MUST HAVE before beta release)**
- [x]  DSG support for Custom Enum generation
- [x]  Add entitlement validation on the server as needed
- [x] Define and implement migration strategy to create default DTOs
- [x]  InputType(), ArgsType() and other graphQL decorators
- [x]  Disallow DTO deletion *and disabling* (client)
- [x]  refactor customDtoNameToPath to support multiple custom DTOs with the same name in different modules (import)

**P0.5 (before beta release)**
- [x] Add id property to DTO property or validate on change rename
- [x]  When deleting an entity (or custom module)- check if all children (actions, dtos (? might not want to delete DTOs that might be in used), props…) are also deleted and are not left orphans
- [x] Analytics (detailed in the spec)
- [x] Beta indications (where?)

The remainder of the issues are in: https://github.com/amplication/amplication/issues/8280

PR #8205, opened 2024-03-12
feat(DSG): add custom model creation

morhag90:
Close: #7713 

## PR Details

Add custom model creation. 

## PR Checklist


packages/amplication-server/src/core/block/block.util.ts
```diff
@@ -40,3 +42,35 @@ export const mergeAllSettings = (
...
+    default:
+      return;
+  }
+}
+
+async function validateCustomActionsEntitlement(
+  workspaceId: string
+): Promise<void> {
+  try {
+    const stiggClient = await this.getStiggClient();
```
mulygottlieb, line 64: 1. I am not sure where `this.getStiggClient()` is implemented, I believe this is a copy-paste from the billing service which actually has a `getStiggClient()` function. What am I missing?
2. Perhaps it's better not to have stigg in the block.util anyway, and instead use the relevant function from the billing service (and inject it here - e.g. `this.billingService.getBooleanEntitlement()`)

packages/amplication-server/src/core/block/block.util.ts
```diff
@@ -40,3 +42,35 @@ export const mergeAllSettings = (
     ...mergedObj,
   };
 };
+
+export async function validateEntitlement(
```
mulygottlieb, line 46: let's rename to `validateBlockEntitlement` or `validateEntitlementForBlockType` so it's more clear

packages/data-service-generator/src/server/custom-module/custom-service/service.template.ts
```diff
@@ -0,0 +1,7 @@
+import { Injectable } from "@nestjs/common";
+import { PrismaService } from "../prisma/prisma.service";
+
+@Injectable()
+export class SERVICE {
+  constructor(protected readonly prisma: PrismaService) {}
```
mulygottlieb, line 6: Do we need to inject prisma into custom module services?

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
```diff
@@ -0,0 +1,119 @@
...
+  dtoNameToPath: Record<string, string>
+): Promise<ModuleMap> {
+  const context = DsgContext.getInstance;
+  const { moduleContainers } = context;
+
+  const customModuleContainers = moduleContainers.filter(
+    (module) => module.entityId === undefined
+  );
+
+  const customModules = new ModuleMap(DsgContext.getInstance.logger);
```
mulygottlieb, line 23:
```suggestion
  const customModules = new ModuleMap(context.logger);
```

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
```diff
@@ -0,0 +1,119 @@
...
+  for await (const customModule of customModuleContainers) {
+    await customModules.merge(
+      await createCustomModuleModules(customModule, dtoNameToPath)
+    );
+  }
+
+  return customModules;
+}
+
+async function createCustomModuleModules(
```
mulygottlieb, line 33: Just a suggestion, not a must, but the naming is very confusing with `createCustomModuleModules` and `createCustomModulesModules`.
Perhaps rename this one to something like `createSingleCustomModuleModules` just for easier readability?

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
```diff
@@ -0,0 +1,119 @@
...
+    (appInfo.settings.serverSettings.generateRestApi &&
+      (await createCustomModuleControllerModules(
+        resource,
+        customModuleName,
+        serviceModule.path,
+        dtoNameToPath
+      ))) ||
+    new ModuleMap(DsgContext.getInstance.logger);
+
+  const [controllerModule] = controllerModules.modules();
```
mulygottlieb, line 65: if `generateRestApi` is false, and there is an empty module map here,  wouldn't this line throw? or will you just get an `undefined` value as `controllerModule` ?

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
```diff
@@ -0,0 +1,119 @@
...
+
+  const resolverModules =
+    (appInfo.settings.serverSettings.generateGraphQL &&
+      (await createCustomModuleResolverModules(
+        customModuleName,
+        serviceModule.path,
+        dtoNameToPath
+      ))) ||
+    new ModuleMap(DsgContext.getInstance.logger);
+  const [resolverModule] = resolverModules.modules();
```
mulygottlieb, line 89: same question here

packages/data-service-generator/src/server/custom-module/custom-service/service.template.ts
morhag90: removed

packages/data-service-generator/src/server/custom-module/custom-controller/create-controller.ts
mulygottlieb, line 1: Not sure if it's critical, but there is a lot of usage of "Entity*" variables here and this might be confusing for a custom module controller that isn't really linked to any entity. Perhaps we can consider naming it better?

packages/data-service-generator/src/server/custom-module/custom-module/create-module.ts
mulygottlieb, line 1: same comment here about using "entity*" so much

packages/amplication-server/src/core/block/block.util.ts
morhag90: @mulygottlieb  Logic moved to blockType service

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
morhag90: @mulygottlieb  will you just get an undefined value as controllerModule  (same logic as in the existing controller creation)

packages/data-service-generator/src/server/custom-module/create-custom-module.ts
morhag90: @mulygottlieb will you just get an undefined value as controllerModule (same logic as in the existing controller creation)

mulygottlieb (approved):

Issue #8220, opened 2024-03-15
Add support for custom sendGrid notifications


yuval-hazaz:
### Feature description

When sending transactional emails, we want to use SendGrid templates instead of the built-in Novu templates

mulygottlieb:
Passed QA:


![Image](https://github.com/amplication/amplication/assets/1912933/677fc1e9-cd03-4ed0-a27d-8b815c0f6f8f)

PR #8221, opened 2024-03-15
feat(Notifications): Support custom SendGrid Templates 

yuval-hazaz:
Close: #8220
## PR Details

Allow using the overrides property in Novu API to use custom template when sending emails from SendGrid via Novu

## PR Checklist

### DIFF
--- a/packages/notification-service/src/notification-packages/previewUserGenerationCompleted.ts
+++ b/packages/notification-service/src/notification-packages/previewUserGenerationCompleted.ts
@@ -2,6 +2,7 @@ import { KAFKA_TOPICS } from "@amplication/schema-registry";
 import { NotificationContext } from "../util/novuTypes";
 
 const NOTIFICATION_KEY = "complete-preview-onboarding";
+const SENDGRID_TEMPLATE_ID = "d-f8d34551e2c7436681b34f8af0782788";
 
 export const previewUserGenerationCompleted = async (
   notificationCtx: NotificationContext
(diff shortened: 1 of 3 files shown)


packages/notification-service/src/notification-packages/previewUserGenerationCompleted.ts
```diff
@@ -2,6 +2,7 @@ import { KAFKA_TOPICS } from "@amplication/schema-registry";
 import { NotificationContext } from "../util/novuTypes";
 
 const NOTIFICATION_KEY = "complete-preview-onboarding";
+const SENDGRID_TEMPLATE_ID = "d-f8d34551e2c7436681b34f8af0782788";
```
overbit, line 5: Should be this configure as environment variable? Is sendgrid supporting versioning?

yuval-hazaz: I don't think so, as it is specific to the notification package. 
Meaning, each notification package may have a different template ID, just like the NOTIFICATION_KEY

mulygottlieb (approved):

packages/notification-service/src/notification-packages/previewUserGenerationCompleted.ts
overbit: I meant for testing in other envs. But it ok :)

overbit (approved):

Issue #8058, opened 2024-02-20
🐛 Bug Report: generated app Swagger UI doesn't work with `auth-basic` plugin  


overbit:
### What happened?

The generated app's Swagger UI is configured to use JWT Bearer token regardless if `auth-basic` plugin was installed or not.

### What you expected to happen

If a service has  `auth-basic` plugin, the swagger-ui should be generated with the `BasicAuth` option instead of `BearerAuth` one. 

### How to reproduce

- create a service
- install auth-core and auth-basic plugins
- generate the code
- run the code and try to make requests with auth

voytovichdm:
It seems I've found a solution, can I do this task?

PR #8245, opened 2024-03-25
fix(client): editing SwaggerUI with auth-basic

voytovichdm:
Close: #[8058](https://github.com/amplication/amplication/issues/8058)

## PR Details

Added the necessary authentication type to the ServiceSettings.authProvider field when modifying authentication plugins.

## PR Checklist


packages/amplication-client/src/Plugins/PluginsCatalog.tsx
overbit, line 1: Thanks @voytovichdm for the PR. Unfortunately we cannot merge this as it goes against the amplication design. 
### TL;DR; the change must live within the `auth-basic` plugin instead of `amplication-client`


With plugins, `data-service-generator` should not be aware of which plugins are installed and each plugin will perform the necessary operations to produce the expected generated code.

In the old days, client, server and dsg where more aware of things like authentication strategy, today fully managed by this plugins. Of course, we would need to clean up all code that is currently still rely on this  `authProvider` property and I'm opening an issue for this.

overbit (changes requested):

PR #8253, opened 2024-03-27
Feat/7713 custom dtos migration

morhag90:
Close: #7713 

## PR Details

Create a default dtos migration script for existing entities 
## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/workspace/workspace.controller.ts
+++ b/packages/amplication-server/src/core/workspace/workspace.controller.ts
@@ -3,7 +3,10 @@ import { AmplicationLogger } from "@amplication/util/nestjs/logging";
 import { WorkspaceService } from "./workspace.service";
 import { ApiTags } from "@nestjs/swagger";
 import { ConfigService } from "@nestjs/config";
-import { CreateWorkspacesResourcesDefaultCustomActionsMigrationInput } from "./dto/CreateWorkspacesResourcesDefaultCustomActionsMigrationInput";
+import {
+  CreateWorkspacesResourcesDefaultCustomActionsMigrationInput,
+  CreateWorkspacesResourcesDefaultCustomDtosMigrationInput,
+} from "./dto/CreateWorkspacesResourcesDefaultCustomActionsMigrationInput";
 
 @ApiTags("workspace")
 @Controller("migrate-custom-actions")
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/workspace/workspace.controller.ts
```diff
@@ -59,4 +62,24 @@ export class WorkspaceController {
...
+  async createWorkspacesResourcesDefaultCustomDtosMigration(
+    @Param("token") token: string,
+    @Body()
+    data: CreateWorkspacesResourcesDefaultCustomDtosMigrationInput
+  ): Promise<boolean> {
+    this.logger.info("createWorkspacesResourcesDefaultCustomDtosMigration....");
+    const { quantity } = data;
+
+    if (
+      this.configService.get<string>("CUSTOM_DTO_MIGRATION_TOKEN") !== token
```
mulygottlieb, line 76: We need a "twin" PR in cluster-configuration that adds CUSTOM_DTO_MIGRATION_TOKEN to all envs, based on an AWS secret. We can do that part together.

packages/amplication-server/src/core/workspace/workspace.service.ts
```diff
@@ -694,6 +696,71 @@ export class WorkspaceService {
...
+      where: {
+        projects: {
+          some: {
+            deletedAt: null,
+            resources: {
+              some: {
+                deletedAt: null,
+                archived: { not: true },
+                resourceType: EnumResourceType.Service,
+                blocks: { none: { blockType: EnumBlockType.ModuleDto } },
```
mulygottlieb, line 712: do we want "blocks:  { none: ..."  or to remove this condition to allow finding services where not ALL ModuleDTOs were created?
In addition, if someone has ModuleDTOs from custom modules (e.g., Yuval), this will not run the migration for them for the same reason.

I think we should consider how we only find entities that don't have ModuleDTOs, perhaps?

packages/amplication-server/src/core/workspace/workspace.service.ts
```diff
@@ -817,6 +884,57 @@ export class WorkspaceService {
...
+          resources,
+          workspaceUser
+        );
+
+        if (hasChanges) {
+          try {
+            await this.projectService.commit(
+              {
+                data: {
+                  message: "this is automatic commit for update custom actions",
```
mulygottlieb, line 905:
```suggestion
                  message: "this is automatic commit to update custom actions and create default DTOs",
```

packages/amplication-server/src/core/workspace/workspace.controller.ts
morhag90: Changed to the existing var for custom-action migrations

mulygottlieb (approved):

PR #8268, opened 2024-04-01
feat(server): add module dto validation before deleting an entity

morhag90:
Close: #7713 

## PR Details

Add module dto's validation before deleting an entity.

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -835,14 +835,6 @@ export class EntityService {
         );
       }
 
-      for (const relatedEntityField of relatedEntityFields) {
-        await this.deleteField(
-          { where: { id: relatedEntityField.id } },
-          user,
-          fieldStrategy
-        );
-      }
-
       try {
         await this.moduleService.deleteDefaultModuleForEntity(
           entity.resourceId,
@@ -850,11 +842,25 @@ export class EntityService {
           user
         );
       } catch (error) {
+        const amplicationError = error as AmplicationError;
+        if (
+          amplicationError.cause === "dtoInUse" ||
+          amplicationError.cause === "ActionDtoInUse"
+        )
+          throw new AmplicationError(error);
         //continue to delete the entity even if the deletion of the default module failed.
         //This is done in order to allow the user to workaround issues in any case when a default module is missing
         this.logger.error(
           "Continue with EntityDelete even though the default entity could not be deleted or was not found ",
           error
+        );
+      }
+
+      for (const relatedEntityField of relatedEntityFields) {
+        await this.deleteField(
+          { where: { id: relatedEntityField.id } },
+          user,
+          fieldStrategy
         );
       }
 
--- a/packages/amplication-server/src/core/module/module.service.ts
+++ b/packages/amplication-server/src/core/module/module.service.ts
@@ -18,6 +18,11 @@ import { ConfigService } from "@nestjs/config";
 import { Env } from "../../env";
 import { BillingService } from "../billing/billing.service";
 import { AmplicationLogger } from "@amplication/util/nestjs/logging";
+import {
+  EnumModuleActionType,
+  EnumModuleDtoPropertyType,
+  EnumModuleDtoType,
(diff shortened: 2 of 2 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -835,21 +835,19 @@ export class EntityService {
...
           entity.id,
           user
         );
       } catch (error) {
+        const amplicationError = error as AmplicationError;
+        if (
+          amplicationError.cause === "dtoInUse" ||
+          amplicationError.cause === "ActionDtoInUse"
+        )
+          throw new AmplicationError(error);
```
mulygottlieb, line 850: The comments and log below no longer make sense (lines 851-856).

morhag90: Why? In case the error is thrown for another reason, the process will and should continue @mulygottlieb

packages/amplication-server/src/core/module/module.service.ts
```diff
@@ -223,6 +228,109 @@ export class ModuleService extends BlockTypeService<
...
+        (moduleDto.dtoType === EnumModuleDtoType.Custom ||
+          moduleDto.dtoType === EnumModuleDtoType.CustomEnum)
+    );
+
+    allOtherResourceCustomModuleDtos.forEach((moduleDto) => {
+      moduleDto.properties.forEach((prop) => {
+        if (prop) {
+          prop.propertyTypes.forEach((propType) => {
+            const currentDto = moduleModuleDtos.find(
+              (x) => x.id === propType.dtoId
```
mulygottlieb, line 268: what if this property is a primitive and not a DTO? you would not have propType.dtoId (maybe that's okay in this case actually, but we need to make sure)

packages/amplication-server/src/core/module/module.service.ts
```diff
@@ -223,6 +228,109 @@ export class ModuleService extends BlockTypeService<
...
+
+    const allOtherResourceCustomModuleActions = allResourceModuleActions.filter(
+      (moduleAction) =>
+        moduleAction.parentBlockId !== moduleId &&
+        moduleAction.actionType === EnumModuleActionType.Custom
+    );
+
+    allOtherResourceCustomModuleActions.forEach((moduleAction) => {
+      if (
+        moduleAction.inputType.type === EnumModuleDtoPropertyType.Dto ||
```
mulygottlieb, line 312: Shouldn't we add some more || clauses for the outputType? (perhaps the input is a primitive but the output is a DTO?)

packages/amplication-server/src/core/module/module.service.ts
mulygottlieb, line 1: what about when deleting a custom module (not via entity)? Will this be handled separately?

packages/amplication-server/src/core/module/module.service.ts
morhag90: @mulygottlieb  Done

packages/amplication-server/src/core/module/module.service.ts
morhag90: @mulygottlieb, In this case, it will not have the dtoId field

packages/amplication-server/src/core/module/module.service.ts
morhag90: Is this scenario in the current task scope?  @mulygottlieb

morhag90: Now I see in the issue that there is a mention of a custom module. Checking it @mulygottlieb

morhag90: Handle (I added the same function in the delete method)

packages/amplication-server/src/core/module/module.service.ts
mulygottlieb: but don't you need at least `propType?.dtoId` or will it just be an existing field with `undefined` value?

morhag90: An existing field with an undefined value

mulygottlieb (approved):

PR #8273, opened 2024-04-02
Feat/7713 analytics

morhag90:
Close: #7713 

## PR Details

Add analytics to custom-modules/actions/dtos services. 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/module/module.service.ts
+++ b/packages/amplication-server/src/core/module/module.service.ts
@@ -18,6 +18,8 @@ import { ConfigService } from "@nestjs/config";
 import { Env } from "../../env";
 import { BillingService } from "../billing/billing.service";
 import { AmplicationLogger } from "@amplication/util/nestjs/logging";
+import { SegmentAnalyticsService } from "../../services/segmentAnalytics/segmentAnalytics.service";
+import { EnumEventType } from "../../services/segmentAnalytics/segmentAnalyticsEventType.types";
 const DEFAULT_MODULE_DESCRIPTION =
   "This module was automatically created as the default module for an entity";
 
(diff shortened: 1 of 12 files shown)


packages/amplication-server/src/core/module/module.service.ts
```diff
@@ -61,13 +64,55 @@ export class ModuleService extends BlockTypeService<
...
+      });
+    }
+
+    return super.findMany(prismaArgs);
+  }
+
+  async create(
+    args: CreateModuleArgs,
+    user: User,
+    trackEvent = true
```
mulygottlieb, line 94: should it default to true or false?

packages/amplication-server/src/core/module/module.service.ts
```diff
@@ -100,6 +145,20 @@ export class ModuleService extends BlockTypeService<
...
+
+    const subscription = await this.billingService.getSubscription(
+      user.workspace?.id
+    );
+
+    await this.analytics.trackWithContext({
+      properties: {
+        name: args.data.name,
+        planType: subscription.subscriptionPlan,
+        operation: "rename",
```
mulygottlieb, line 157: Is this the only possible update for Module (only renaming)?

packages/amplication-server/src/core/moduleAction/moduleAction.service.ts
```diff
@@ -175,6 +222,19 @@ export class ModuleActionService extends BlockTypeService<
...
+      user.workspace?.id
+    );
+
+    await this.analytics.trackWithContext({
+      properties: {
+        name: moduleAction.name,
+        operation: "delete",
+        planType: subscription.subscriptionPlan,
+      },
+      event: EnumEventType.InteractUserAction,
```
mulygottlieb, line 236: Do you know this is never on an implication action here (e.g. use InteractAmplicationAction in some cases)?

packages/amplication-server/src/core/module/module.service.ts
morhag90: true, sending false from the inner function

packages/amplication-server/src/core/module/module.service.ts
morhag90: name and description @mulygottlieb

mulygottlieb (approved):

Issue #8280, opened 2024-04-02
[Epic] Custom Actions and Types - Phase 3


mulygottlieb:
Spec-
https://www.notion.so/amplication/Custom-Types-Actions-Phase-2-b83fb9234c78419584bbb47a0fc31634?pvs=4
Design- 
TBD

First part of this was handled in: https://github.com/amplication/amplication/issues/7713

**P1 issues**
- [ ]  Union type (without DTOs) ? null type ? DTO vs primitives; Remove the union type?
- [ ]  Use a flag to determine where to generate the custom DTOs (dto folder/ base folder, entity folder…) **(to discuss)**
    - We currently generate custom actions in the base file. should we keep it?
- [ ]  Creating services trigger hundreds of records (**to discuss**)
- [ ]  Use server-side logic to guess DTO props types like with entity field (**quick win?**)
- [x]  Create action/dto/enum modal (with Mor’s new modal) (**UX**)
- [ ] Create module modal L&F (**UX**)
- [ ]  Free and Trial plan experience should be adjusted to meet the behavior of other Enterprise features (tooltips etc.)
    - The following actions should be enabled for Enterprise plan only (and Trial) - "Add Module", "Add Action", "Add DTO", "Add Enum", enable / disable action (two places- in the module screen, and in the action screen itself), edit action (besides add / edit its description), edit DTO / Enum (besides add / edit its description) 
- [x]  Bug: DTO can have a member of type Enum, and a member of type DTO of type Enum. Confusing... The Enum type should be removed.
- [ ]  Bug: "The users will be able to delete empty modules that they created." - for now a custom module can be deleted even in case there are actions / DTOs in it.
- [x]  Bug: When searching for a specific module (in the "all modules" screen), choosing / pressing one module, and back to the "all modules" screen- the search string is "stuck" although not visible.

**UI/UX issues (P1/P2)**

- [X]  We need to think how we can make it clearer what is expected from a user who wants to define a new API? (**P1**)
- [ ]  The modules / Actions / DTOs should be sorted alphabetically **(P1)**
- [ ]  From the spec: "Amplication’s Actions, and DTOs will be displayed with a special indication, and they cannot be deleted/moved between modules." **(P1)**
    - Be able to filter out default actions/DTOs for better navigation
- [ ]  Bug: Choosing a module in the "all modules" search should highlight it in the menu **(P1)**.
[...]

PR #8296, opened 2024-04-08
feat(server): add lazy migration to custom-dtos

morhag90:
Close: #8280 

## PR Details

Add lazy migration to custom-dtos. 
## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/workspace/workspace.resolver.ts
+++ b/packages/amplication-server/src/core/workspace/workspace.resolver.ts
@@ -93,6 +93,10 @@ export class WorkspaceResolver {
       currentUser
     );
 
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -93,6 +93,10 @@ export class WorkspaceResolver {
       currentUser
     );
 
+    await this.workspaceService.dataMigrateWorkspaceResourcesCustomDtos(
+      currentUser.workspace?.id
+    );
```
overbit, line 98: As we are passing a potentially undefined parameter, we should always check it the method call or first thing inside the method logic (param validation)

Issue #8297, opened 2024-04-08
🐛 Bug Report: Insight Panel is broken on smaller screens


yuval-hazaz:
### What happened?

![image](https://github.com/amplication/amplication/assets/43705455/1b41595a-d3f0-4095-acde-d340f26f8cb7)

### What you expected to happen

The page should be responsive

### How to reproduce

use smaller screen

PR #8298, opened 2024-04-08
feat(client):responsive container

yuval-hazaz:
Close: #8297 

## PR Details

Add responsive container on pages with the insights panel to allow breakpoints on smaller screens
image

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Components/ResponsiveContainer.tsx
+++ b/packages/amplication-client/src/Components/ResponsiveContainer.tsx
@@ -1,0 +1,62 @@
+import classNames from "classnames";
+import { useEffect, useLayoutEffect, useRef, useState } from "react";
+
+import useResizeObserver from "@react-hook/resize-observer";
+
+const DEFAULT_MIN_WIDTH_CLASSES: { [key: string]: number }[] = [
+  { xxl: 1400 },
+  { xl: 1200 },
+  { lg: 992 },
+  { md: 768 },
+  { sm: 576 },
+];
+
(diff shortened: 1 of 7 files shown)


packages/amplication-client/src/Components/ResponsiveContainer.tsx
```diff
@@ -0,0 +1,62 @@
...
+        (minWidthClass) => width > Object.values(minWidthClass)[0]
+      );
+
+      const classes = breakPoint.map(
+        (minWidthClass) => `${className}--${Object.keys(minWidthClass)[0]}`
+      );
+
+      setResponsiveClassName(classes.join(" "));
+    }
+  }, [className, minWidthClasses, width]);
```
overbit, line 53: Is the responsive class name made up all the class names of sizes over the width on purpose?
What is the expected result for width=100?

yuval-hazaz: yes
the parent component provides a className, for example - "panel-wrapper"
the results should be `class="panel-wrapper panel-wrapper-ms panel-wrapper-md panel-wrapper-lg panel-wrapper-xl"`

The parent container should also do something like this:
```css
.panel-wrapper{
    flex-direction:column;
}

.panel-wrapper-lg .panel-wrapper{.  
    flex-direction:row;
}
```
this way, for all sizes from "lg" and above, the flex-direction will be "row"

overbit (approved):

PR #8315, opened 2024-04-11
chore(server): remove workspace migration scripts

overbit:
Close: #7713

## PR Details

Remove all migration code to avoid issues on new migrations

## PR Checklist


packages/amplication-server/src/core/workspace/workspace.resolver.ts
```diff
@@ -88,11 +88,6 @@ export class WorkspaceResolver {
       currentUser
     );
 
-    await this.workspaceService.migrateWorkspace(
```
mulygottlieb, line 91: Are we confident that we finished with the old migration and we no longer need this lazy migration?

mulygottlieb (commented):
Let's wait a couple of weeks to see we really no longer need any of it

Issue #7402, opened 2023-11-10
Plugin development: main.ts before and after events


codingmickey:
### Feature description

As I was working on the #7043 issue there was a line of code that was needed to be added in the `main.ts` file [Ref](https://github.com/codingmickey/test/commit/18892ce8e7c6ff5e02bca085fb0549e501e547d1#diff-09e5808b2021b1aa80c9b870910b028a5ab125cc0534479b0f949bd09646fdad) 

as there are no events on the createMainTs or something like this.. should I create another field in the context such as what is used for  `hasBigIntFields` .. and then add this line of import and all to the template

That wouldn't be an optimal solution but.. so creating events for main.ts file generation too like others (but I don't have any idea how long that would take.. as I don't have a complete idea about the internals of the dsg but would love to know one day)

cc: @overbit 

### Use case

Improvements in the plugin development

### Are you willing to submit PR?

Yes I am willing to submit a PR!

overbit:
@PazYanoverr  / @mulygottlieb what do you think?

Shurtu-gal:
It makes sense regarding createMain.ts as there is no way to change the template. @codingmickey  It would be good to add a plugin wrapper and make changes to the template file in the plugin.

Shurtu-gal:
For `hasBigIntFields` we have [main.with-bigint.template.ts](https://github.com/amplication/amplication/blob/master/packages/data-service-generator/src/server/create-main/main.with-bigint.template.ts).

You just need to pass it in context,
https://github.com/amplication/amplication/blob/a89dfd6279619fe803baf10d7f043eb915cbc4b3/packages/data-service-generator/src/server/create-main/create-main-file.ts#L9

Shurtu-gal:
> > For `hasBigIntFields` we have [main.with-bigint.template.ts](https://github.com/amplication/amplication/blob/master/packages/data-service-generator/src/server/create-main/main.with-bigint.template.ts).
> > You just need to pass it in context,
> > https://github.com/amplication/amplication/blob/a89dfd6279619fe803baf10d7f043eb915cbc4b3/packages/data-service-generator/src/server/create-main/create-main-file.ts#L9
> 
> No no I meant "like `hasBigIntFields`" so as to creating for my use case such as `hasFileUploadPlugin` (but this wouldn't make any sense to be here as it would just increase the if-else complexity if there are more changes in the future modifying the `main.ts` file)

Understood. I think it's better to use a plugin-wrapper then. Any thoughts @overbit?

PR #8316, opened 2024-04-11
feat(dsg): add create main file event

Shurtu-gal:
Close: #7402

## PR Details

Adds createMainFile events.

## PR Checklist

### DIFF
--- a/packages/data-service-generator/src/server/create-main/create-main-file.ts
+++ b/packages/data-service-generator/src/server/create-main/create-main-file.ts
@@ -7,30 +7,53 @@ import {
 } from "@amplication/code-gen-types";
 import { MAIN_TS_FILE_NAME, MAIN_TS_WITH_BIGINT_FILE_NAME } from "../constants";
 import DsgContext from "../../dsg-context";
-import { formatCode, print, readFile } from "@amplication/code-gen-utils";
+import {
+  formatCode,
+  parse,
+  print,
+  readFile,
+} from "@amplication/code-gen-utils";
 import pluginWrapper from "../../plugin-wrapper";
+import fs from "fs";
+import { namedTypes } from "ast-types";
 
 export async function createMainFile(): Promise<ModuleMap> {
   const mainFilePath = path.resolve(__dirname, MAIN_TS_FILE_NAME);
-  const mainWithBigintFilePath = path.resolve(
-    __dirname,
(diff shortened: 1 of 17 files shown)


packages/data-service-generator/src/server/create-main/create-main-file.ts
```diff
@@ -1,38 +1,50 @@
...
     __dirname,
     MAIN_TS_WITH_BIGINT_FILE_NAME
   );
-  const mainFileContent = await fs.readFile(mainFilePath, "utf-8");
-  const mainWithBigintFileContent = await fs.readFile(
-    mainWithBigintFilePath,
-    "utf-8"
-  );
+  const template = await readFile(mainFilePath);
+  const bigIntTemplate = await readFile(mainWithBigintFilePath);
```
overbit, line 20: I'd recommend to move the logic that handle the template selection before so the plugin event would need to only take care of one template, making it simplier and more maintainable. @Shurtu-gal what do you think?

Shurtu-gal: I could do that, however there is a difference between the two, which one should the template pass ?

diff: https://github.com/amplication/amplication/blob/2e4118c079dfc69b516598b843ec97fc7310ccfa/packages/data-service-generator/src/server/create-main/main.with-bigint.template.ts#L26-L31

overbit: we can unify the two with the following patch 
[refactor-main.ts.patch](https://github.com/amplication/amplication/files/14981999/refactor-main.ts.patch)

Shurtu-gal: So basically you want this to be applied in the AST tree, if I am able to understand it correctly.

Shurtu-gal: Done @overbit.

overbit:
well done @Shurtu-gal. 
Can you please update the snapshot too `npx nx update-snapshot data-service-generator`?
We also need to add the new event in out docs site https://github.com/amplication/docs

overbit (approved):

Shurtu-gal:
> npx nx update-snapshot data-service-generator

@overbit done.

overbit (approved):

mulygottlieb (approved):

PR #8317, opened 2024-04-13
feat(server):add kafka pubsub for the assistant

yuval-hazaz:
Fixes: https://github.com/amplication/private-issues/issues/171

## PR Checklist


yuval-hazaz:
@overbit please have a careful look at the kafka.pubsub.service 
I tried to follow the standards from the publisher service. I would love to get your feedback

libs/util/nestjs/kafka/src/kafka.pubsub.service.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const kafka = new Kafka({
+      logLevel: logLevel.ERROR,
+      clientId: kafkaEnv.getClientId() + `-${randomUUID()}`,
+      brokers: kafkaEnv.getBrokers(),
+      connectionTimeout: 25000,
+      retry: {
+        retries: 3,
+        maxRetryTime: 3000,
+      },
```
overbit, line 29: kafka configurations should match the `createNestjsKafkaConfig` to allow a single common setup and potentially customised that on top if needed.

libs/util/nestjs/kafka/src/kafka.pubsub.service.ts
```diff
@@ -0,0 +1,45 @@
+import { Injectable, OnModuleInit } from "@nestjs/common";
+import { Kafka, logLevel } from "kafkajs";
+import { KafkaPubSub } from "graphql-kafkajs-subscriptions";
```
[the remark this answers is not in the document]
overbit: To fully understand the risks, you can start 2 instances of the server on your local and see what happens

libs/util/nestjs/kafka/src/Kafka.module.ts
```diff
@@ -26,8 +25,7 @@ import { KafkaPubSubService } from "./kafka.pubsub.service";
       useClass: KafkaMessageJsonSerializer,
     },
     KafkaProducerService,
-    KafkaPubSubService,
   ],
-  exports: [KAFKA_SERIALIZER, KafkaProducerService, KafkaPubSubService],
+  exports: [KAFKA_SERIALIZER, KafkaProducerService],
```
overbit, line 29: Removed the service from the kafkaModule otherwise all services using this will have a kafka consumer reading all the partition of the shared topic even if they never need it.

packages/amplication-server/.env
```diff
@@ -77,6 +77,7 @@ CRON_SECRET_KEY=[cron-secret-key]
 
 FEATURE_CUSTOM_ACTIONS_ENABLED=true
 
-CHAT_OPENAI_KEY = "CHAT_OPENAI_KEY"
-CHAT_ASSISTANT_ID = "CHAT_ASSISTANT_ID"
-FEATURE_AI_ASSISTANT_ENABLED = false
\ No newline at end of file
+CHAT_OPENAI_KEY="CHAT_OPENAI_KEY"
+CHAT_ASSISTANT_ID="CHAT_ASSISTANT_ID"
+FEATURE_AI_ASSISTANT_ENABLED=true
```
overbit, line 82: spaces in a dotenv file between vars and `=` and values are invalid

overbit:
@yuval-hazaz  @mulygottlieb  I've recorded the local testing too with some additional explanations https://jam.dev/c/653a24ca-5ef6-43ea-9610-3edfc3998c0b

libs/util/nestjs/kafka/tsconfig.lib.json
```diff
@@ -7,5 +7,8 @@
     "types": ["node"]
   },
   "exclude": ["jest.config.ts", "**/*.spec.ts", "**/*.test.ts"],
-  "include": ["**/*.ts"]
+  "include": [
+    "**/*.ts",
+    "../../../../packages/amplication-server/src/graphql/graphqlSubscriptionKafka.service.ts"
+  ]
```
yuval-hazaz, line 13: why is that needed?

packages/amplication-server/src/graphql/graphql.module.ts
```diff
@@ -0,0 +1,52 @@
...
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            "graphql-ws": true,
+          },
+        };
+      },
+      inject: [ConfigService],
+    }),
+  ],
+  controllers: [],
+  providers: [GraphqlSubscriptionPubSubKafkaService],
```
yuval-hazaz, line 45: Should we really merge the graphql module with the kafka subscription service?
I am wondering whether it will make more sense to use two seperate modules for clarity

packages/amplication-server/src/graphql/graphqlSubscriptionPubSubKafka.service.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  async onModuleInit() {
+    const config = createNestjsKafkaConfig().options;
+
+    const kafka = new Kafka(config.client as KafkaConfig);
+
+    this.pubsub = await KafkaPubSub.create({
+      kafka,
+      topic: KAFKA_TOPICS.SHARED_GRAPHQL_SUBSCRIPTION_PUBSUB_TOPIC,
+      groupIdPrefix: config.consumer.groupId + "-graphql-subscrition",
```
mulygottlieb, line 19: Is this groups ID constant for all server instances? something like "amplication-server-group-prod" ?

libs/util/nestjs/kafka/tsconfig.lib.json
overbit, line 13: Absolutely an error. Good spot! 
```suggestion
  "include": ["**/*.ts"]
```

packages/amplication-server/src/graphql/graphqlSubscriptionPubSubKafka.service.ts
overbit: yep, but currently that is a prefix as the graphql-kafkajs-subscriptions lib doesn't support custom fixed groupids. 
i.e. if we have 2 runtime of server running we will see 3 consumer groups for server like:
image

packages/amplication-server/src/graphql/graphql.module.ts
overbit: It makes sense for code organisation. Moving to the on only module that is using it and we can refactor if/when we will need it in other services

overbit (approved):

Issue #8307, opened 2024-04-09
Jovu: Add installPlugin function


PazYanoverr:
Add installPlugin function to Jovu. The function requires to have the project, the service, and the plugin name / id.

abrl91:
It installs the plugin but, we can see it only after a refresh


https://github.com/amplication/amplication/assets/39680385/6aecf108-1111-4d14-b1e2-c38ecce7f0b9

PR #8341, opened 2024-04-28
Epic: Jovu Phase 2

yuval-hazaz:
Fixes: #8307

## PR Details

Adds the plugins API client on amplication-server to make the catalog available for Jovu

## PR Checklist


packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -665,5 +674,58 @@ export class AssistantService {
...
+        (plugin) => omit(plugin, ["__typename"])
+      );
+      this.logger.debug(
+        `Chat: Plugins: ${JSON.stringify(plugins)}`,
+        null,
+        context
+      );
+      return plugins;
+    },
+    installPlugins: async (
```
overbit, line 688: we are missing the unit tests here. Try the edge case of `this.pluginCatalogService.getPluginWithLatestVersion` returning  `undefined`

packages/amplication-server/src/core/pluginCatalog/pluginCatalog.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      const plugin =
+        data.plugins && data.plugins.length ? data.plugins[0] : undefined;
+      if (plugin) {
+        const versions = plugin.versions.filter((version) => version.isLatest);
+        return {
+          ...plugin,
+          versions,
+        };
+      }
+      return undefined;
```
overbit, line 71: undefined shouldn't be valid as the return type is Promise<PluginCatalogItem>

packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -665,5 +674,58 @@ export class AssistantService {
             : "Entity",
       }));
     },
+    getPlugins: async (args: undefined, context: AssistantContext) => {
+      const plugins = (await this.pluginCatalogService.getPlugins()).map(
+        (plugin) => omit(plugin, ["__typename"])
```
overbit, line 679: keep the data manipulation close to the action. 
`PluginCatalogItem` doesn't contain `__typename` so this line should have not effect. A unit tests would have caught it.

packages/amplication-server/src/core/pluginCatalog/pluginCatalog.service.ts
```diff
@@ -0,0 +1,103 @@
+import { Injectable } from "@nestjs/common";
+import { ApolloClient, InMemoryCache, gql } from "@apollo/client";
```
overbit, line 2: To avoid any React dependency, import the apollo client from the core instead.
```suggestion
import { ApolloClient, InMemoryCache, gql } from "@apollo/client/core";
```

packages/amplication-client/src/Assistant/AssistantMessage.tsx
```diff
@@ -28,7 +30,14 @@ const AssistantMessage = ({ message, onOptionClick }: Props) => {
...
           <>
-            <Icon icon="ai" color={EnumTextColor.ThemeTurquoise} size="large" />
+            <img
+              src={jovu}
+              alt="jovu"
+              width={30}
+              height={30}
+              style={{ background: "white", borderRadius: "50%" }}
+            />
+            {/* <Icon icon="ai" color={EnumTextColor.ThemeTurquoise} size="large" /> */}
```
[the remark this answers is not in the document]
yuval-hazaz: we are still working on the logo, so we will get back to it soon anyway.

packages/amplication-server/src/core/pluginCatalog/pluginCatalog.service.ts
yuval-hazaz: nice!

overbit (approved):
Approved but it's missing unit tests. All these new services are missing them.

Issue #8306, opened 2024-04-09
Jovu: Add createAPI function


PazYanoverr:
Add createAPI function to Jovu. The function requires to have the project, the service, and the module information.

PR #8342, opened 2024-04-29
feat(server):add module functions for assistant service

yuval-hazaz:
Close: https://github.com/amplication/amplication/issues/8306

## PR Checklist


packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -35,6 +38,13 @@ enum EnumAssistantFunctions {
   GetProjectPendingChanges = "getProjectPendingChanges",
   GetPlugins = "getPlugins",
   InstallPlugins = "installPlugins",
+  GetServiceModules = "getServiceModules",
+  CreateModule = "createModule",
+  GetModuleDtosAndEnums = "getModuleDtosAndEnums",
+  CreateModuleDto = "createModuleDto",
+  // CreateModuleEnum = "createModuleEnum",
+  // GetModuleActions = "getModuleActions",
+  // CreateModuleAction = "createModuleAction",
```
overbit, line 47: Are they missed or not yet implemented? 
If the second, a better approach is to implement a `throw NotImplemented` exception instead.

packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -490,7 +501,7 @@ export class AssistantService {
           data: {
             displayName: args.name,
             pluralDisplayName: pluralDisplayName,
-            name: camelCase(args.name),
+            name: pascalCase(args.name),
```
overbit, line 504: would it make sense to move this string manipulation inside the entityService to avoid inconsistencies due to other calls to the same `.entityService.createOneEntity` methods?

packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
```diff
@@ -227,7 +228,7 @@ export class ModuleDtoService extends BlockTypeService<
         ...args,
         data: {
           ...args.data,
-          properties: [],
+          properties: (args.properties as unknown as JsonArray) ?? [],
```
overbit, line 231: ⚠️ pay attention to this casting as it will hide potential type mismatches.
Type casting should be avoided in typescript projects

yuval-hazaz: that is true. 
the BlockCreateInput uses a specific implementation of [key: string]: JsonValue; which causes a type conflict here. 
If you think you have a quick workaround, please suggest one, otherwise it will not make sense to make a big change to fix it now

yuval-hazaz: @overbit  please note that because of the `[key: string]: JsonValue;` BlockCreateInput anyway accepts all types, and the type validation is done on the previous layer per specific block type

mulygottlieb (approved):

PR #8357, opened 2024-04-30
Feat/168 dotnet promote

morhag90:
Close: https://github.com/amplication/private-issues/issues/168

## PR Details

Add the .NET promote feature.

## PR Checklist


packages/amplication-client/src/util/analytics-events.types.ts
```diff
@@ -166,4 +166,7 @@ export enum AnalyticsEventNames {
 
   // break the monolith preview user
   PreviewUser_GenerateCode = "PreviewUser_GenerateCode",
+
+  ChangedToDotNet = "ChangedToDotNet",
+  ChoseDotNetUsage = "ChoseDotNetUsage",
```
yuval-hazaz, line 171: is that uses?

yuval-hazaz (approved):

packages/amplication-client/src/util/analytics-events.types.ts
morhag90: Good point.
I added the missing analytics.

yuval-hazaz (approved):

PR #8358, opened 2024-04-30
Feat/jovu usage limits

yuval-hazaz:
Close: https://github.com/amplication/private-issues/issues/183

## PR Details

Add enforcement for the number of Jovu requests per a new entitlement. 
image

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Assistant/Assistant.tsx
+++ b/packages/amplication-client/src/Assistant/Assistant.tsx
@@ -1,6 +1,12 @@
 import {
+  EnumContentAlign,
+  EnumFlexDirection,
+  EnumGapSize,
+  EnumItemsAlign,
+  EnumTextAlign,
   EnumTextColor,
   EnumTextStyle,
+  FlexItem,
   Icon,
   Text,
   TextField,
@@ -18,6 +24,8 @@ import classNames from "classnames";
 import { useAppContext } from "../context/appContext";
 import { Link } from "react-router-dom";
 import jovu from "../assets/jovu.svg";
+import { BillingFeature } from "@amplication/util-billing-types";
+import { useStiggContext } from "@stigg/react-sdk";
 type SendMessageType = models.SendAssistantMessageInput;
 
 const INITIAL_VALUES: SendMessageType = {
@@ -53,6 +61,12 @@ const WIDTH_STATE_SETTINGS: Record<
 const Assistant = () => {
   const { currentWorkspace } = useAppContext();
 
+  const { stigg } = useStiggContext();
+
+  const { hasAccess } = stigg.getMeteredEntitlement({
+    featureId: BillingFeature.JovuRequests,
+  });
+
   const [open, setOpen] = useState(true);
   const [widthState, setWidthState] = useState(WIDTH_STATE_DEFAULT);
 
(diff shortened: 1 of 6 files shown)


yuval-hazaz:
@PazYanoverr @mulygottlieb  please note we need to add the new entitlement in the prod env
https://app.stigg.io/development-0239228d/features/feature-jovu-requests

packages/amplication-client/src/Assistant/Assistant.tsx
```diff
@@ -53,6 +63,16 @@ const WIDTH_STATE_SETTINGS: Record<
 const Assistant = () => {
   const { currentWorkspace } = useAppContext();
 
+  const { stigg } = useStiggContext();
+
+  const { hasAccess } = stigg.getMeteredEntitlement({
+    featureId: BillingFeature.JovuRequests,
+  });
+
+  const { data } = useQuery(GET_CONTACT_US_LINK, {
```
mulygottlieb, line 72: With all the recent changes in meeting URLs (yours, mine, Paz, Oren) in the product, this specifically will get you the meeting URL that is defined in the server's .env which is currently Oren's meeting link, but Paz asked me to change it to his...
So - which meeting URL do you want/expect here?

packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -186,14 +214,7 @@ export class AssistantService {
     threadId: string,
     context: AssistantContext
   ): Promise<AssistantThread> {
-    if (!this.assistantFeatureEnabled)
-      throw new AmplicationError("The assistant AI feature is disabled");
-
-    if (context.user.workspace.allowLLMFeatures === false) {
```
mulygottlieb, line 192: You deleted this check (allowLLMFeatures) from the server side and left it only on the client side. Is that intentional?

packages/amplication-client/src/Assistant/Assistant.tsx
yuval-hazaz: let's merge this PR and we can align again on next one

packages/amplication-server/src/core/assistant/assistant.service.ts
yuval-hazaz: it was moved to a common function and used from 2 places in the server side
look on the new code inthis PR

packages/amplication-client/src/Assistant/Assistant.tsx
mulygottlieb: Okay, as long as you don't mind which exact meeting link will be here (Paz/Oren)

packages/amplication-server/src/core/assistant/assistant.service.ts
mulygottlieb: In this PR there are 3 places that allowLLMFeatures appears, two in the client  and one here where it was removed from the server. The other check for assistantFeatureEnabled is indeed in the common function but this check isn't.

mulygottlieb: LOL, GH diff view fail. Just a folded part of the function that I didn't see...

mulygottlieb (approved):

Issue #8388, opened 2024-05-04
Jovu - better support for single field creation and improve relation fields creation


yuval-hazaz:
### Feature description

Improve the way Jovu creates fields, especially relation fields. 
Avoid creation of relation fields of the FK format (e.g. customerId)
Allo creation of single fields

overbit:
Test ✅

PR #8389, opened 2024-05-04
Feat(server): Jovu - Separate field creation from entity creation

yuval-hazaz:
Close: https://github.com/amplication/amplication/issues/8388

## PR Details

Improve the way Jovu creates fields, especially relation fields.
Avoid the creation of relation fields with the FK format - instead of  customerId, use customer
Allo creation of single fields separately from entity creation

## PR Checklist


packages/amplication-server/src/core/assistant/assistant.service.ts
```diff
@@ -347,6 +360,7 @@ export class AssistantService {
         const result = await this.assistantFunctions[functionName].apply(null, [
           args,
           context,
+          loggerContext,
```
mulygottlieb, line 363: Don't you need to cloneDeep the loggerContext here as well?

mulygottlieb (approved):

Issue #8390, opened 2024-05-04
Jovu - improved chat UX


yuval-hazaz:
### Feature description

1. Allow the user to use enter to submit the message while using shift + enter for a new line
2. expand the height of the text input when writing a multi-line message
3. Enable the text input even when the button is disabled (chat is processing) - to allow the use to start writing the next message without waiting for the end of the response (do not allow submit)

overbit:
Test ✅ 



![Image](https://github.com/amplication/amplication/assets/2861984/eedd7be6-b177-477a-a296-25a1e03fc112)

PR #8391, opened 2024-05-04
feat(client):improve chat ux

yuval-hazaz:
Close: #8390

## PR Details

- Allow the user to use enter to submit the message while using shift + enter for a new line
- expand the height of the text input when writing a multi-line message
- Enable the text input even when the button is disabled (chat is processing) - to allow the use to start writing the next message without waiting for the end of the response (do not allow submit)

## PR Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -141,6 +141,7 @@
         "react-dropzone": "^14.2.3",
         "react-helmet": "^6.1.0",
         "react-hotkeys": "^2.0.0",
+        "react-hotkeys-hook": "^4.5.0",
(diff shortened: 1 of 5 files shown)


yuval-hazaz:
@mulygottlieb @overbit I added to this PR a new package to work with hotkeys
The package we currently use is not being maintained for several years already and the new package brings more capabilities that we needed. 

I didn't do it as part of this PR, but we should sometime replace the old package with the new packages and remove the old one 

https://npmtrends.com/react-hotkeys-vs-react-hotkeys-hook

mulygottlieb (approved):

Issue #8401, opened 2024-05-04
Jovu - Check access permissions to assistant functions


yuval-hazaz:
### Feature description

Before executing functions, check the permissions to the requested data based on the user in context

overbit:
Test ✅

PR #8402, opened 2024-05-04
feat(server):check permissions to assistant functions

yuval-hazaz:
Close: #8401

## PR Details

Enforce user permissions on assistant functions 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
+++ b/packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
@@ -26,10 +26,89 @@ import { ResourceService } from "../resource/resource.service";
 import { AssistantContext } from "./dto/AssistantContext";
 import { EnumAssistantFunctions } from "./dto/EnumAssistantFunctions";
 import { MessageLoggerContext } from "./assistant.service";
+import { PermissionsService } from "../permissions/permissions.service";
+import { AuthorizableOriginParameter } from "../../enums/AuthorizableOriginParameter";
+import { get } from "lodash";
 
 export const MESSAGE_UPDATED_EVENT = "assistantMessageUpdated";
 
 export const PLUGIN_LATEST_VERSION_TAG = "latest";
+
+const FUNCTION_PERMISSIONS: {
+  [key in EnumAssistantFunctions]: {
+    paramType: AuthorizableOriginParameter;
+    paramPath: string;
+  };
+} = {
+  createEntities: {
+    paramType: AuthorizableOriginParameter.ResourceId,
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
```diff
@@ -26,11 +26,90 @@ import { ResourceService } from "../resource/resource.service";
...
 import { MessageLoggerContext } from "./assistant.service";
+import { PermissionsService } from "../permissions/permissions.service";
+import { AuthorizableOriginParameter } from "../../enums/AuthorizableOriginParameter";
+import { get } from "lodash";
 
 export const MESSAGE_UPDATED_EVENT = "assistantMessageUpdated";
 
 export const PLUGIN_LATEST_VERSION_TAG = "latest";
 
+const FUNCTION_PERMISSIONS: {
```
mulygottlieb, line 37: Since there is so much repetition here, this can be refactored to look nicer, and the top combinations can be reused (ResourceId+"serviced" and ProjectId+"projected"), etc.
Not a blocker to approve but a DRY thing.

packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
```diff
@@ -107,6 +205,38 @@ export class AssistantFunctionsService {
...
+    const permissions = FUNCTION_PERMISSIONS[functionName];
+    if (!permissions) {
+      return false;
+    }
+
+    if (
+      permissions.paramType === AuthorizableOriginParameter.WorkspaceId &&
+      permissions.paramPath === "context.workspaceId"
+    ) {
+      return true;
```
mulygottlieb, line 222: Wouldn't this allow to create a project in another workspace?

packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
yuval-hazaz: I decided to write this this way to force the developer of any future function to add the permissions settings
otherwise the type check will break

the type [key in EnumAssistantFunctions] requires that all enum members will be listed as properties of the object

packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
yuval-hazaz: no, because the workspace ID is taken from the context, this is just a way to declare that no permissions are checked (because all enum members require a definition) - see the spec file for a relevant test

mulygottlieb (approved):

Issue #8407, opened 2024-05-06
🐛 Bug Report: Getting subscription null when trying to develop and build locally


Jayanth-Parthsarathy:
### What happened?

When I try to build and develop the application locally. I get the error that subscription is null after logging in and attempting to create my first service.

![image](https://github.com/amplication/amplication/assets/86654557/e342ef9e-1b44-44c4-83cb-fb5247a4ccb0)

### What you expected to happen

The application should be able to proceed to the home page where I can create services.

### How to reproduce

Follow the README.md on the development section and develop locally on the system. Run the commands to serve the client and the server. 
Then go to http://localhost:3001 and after logging in you get the subscription is null error.

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #8409, opened 2024-05-06
fix(client): fix subscription null issue

Jayanth-Parthsarathy:
Close: #8407 

## PR Details
This pull request resolves the issue when the user gets the error `subscription is null` when trying to develop the application locally. This fixes this issue by checking for the subscription status with the `subscription?.status` expression in the `FeatureIndicatorContainer.tsx` file
## PR Checklist


mulygottlieb:
@Jayanth-Parthsarathy - thanks for your contribution! 
The PR looks good! The only comment I have is to add the dependency for `status` in the last `useEffect` that you changed:
image
(I think you can replace `subscription` with `status` in this list)

After this change, I can approve and we will merge your first contribution to Amplication 🫶

mulygottlieb (approved):

mulygottlieb:
@Jayanth-Parthsarathy Thanks a lot for helping us improve Amplication 💜

Issue #8434, opened 2024-05-08
Jovu: Improve the onboarding flow and experience


PazYanoverr:

yuval-hazaz:
✅passed qa

PR #8458, opened 2024-05-12
Feat/jovu onboarding

yuval-hazaz:
Close: #8434

## PR Details

- Create workspace with Demo Repo on onboarding
- Use default project settings from the project when creating a service with Jovu
- Use a new messageType to give Jovu special instructions for onboarding
- Build new onboarding experience using Jovu

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/resource/resource.service.ts
+++ b/packages/amplication-server/src/core/resource/resource.service.ts
@@ -596,12 +596,7 @@ export class ResourceService {
           authProvider: EnumAuthProviderType.Jwt, //@todo: remove this property
         },
 
-        gitRepository: {
-          isOverrideGitRepository: false,
-          name: "",
-          resourceId: "",
-          gitOrganizationId: "",
-        },
+        gitRepository: null,
       },
     };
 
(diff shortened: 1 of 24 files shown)


packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -596,12 +596,7 @@ export class ResourceService {
           authProvider: EnumAuthProviderType.Jwt, //@todo: remove this property
         },
 
-        gitRepository: {
-          isOverrideGitRepository: false,
-          name: "",
-          resourceId: "",
-          gitOrganizationId: "",
-        },
+        gitRepository: null,
```
mulygottlieb, line 599: This type of change caused some bugs in the past - can you elaborate on the need for this change?

yuval-hazaz: This is indeed a very delicate place, and I would be happy to revisit it one day.
Because I had a few issues in this area, I copied the behavior of the wizard from the client - and null is the behavior we need as far as I understand.
I must say that it was not that easy to track that

mulygottlieb (approved):

Issue #8467, opened 2024-05-13
feat(code-gen-types): add support to DotNet plugins events


morhag90:
Add support to DotNet plugins events.

mulygottlieb:
Passed QA ✅
Tested in: https://github.com/muly-demo-repos/dotnet-tests-1

PR #8468, opened 2024-05-13
feat(code-gen-types): add DotNet plugin events and types

morhag90:
Close: #8467

## PR Details

Add DotNet plugin events and types. 

## PR Checklist

### DIFF
--- a/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
+++ b/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
@@ -100,7 +100,7 @@ export interface CreateServerDockerComposeParams extends EventParams {
   outputFileName: string;
 }
 
-export interface CreateServerDockerComposeDBParams extends EventParams {
+export interface CreateServerDockerComposeDevParams extends EventParams {
   fileContent: string;
   updateProperties: { [key: string]: any }[];
   outputFileName: string;
--- a/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
+++ b/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
@@ -1,0 +1,168 @@
+import {
+  Entity,
+  EntityLookupField,
+  ModuleContainer,
+  ModuleAction,
+} from "./code-gen-types";
+import { IFile } from "./files";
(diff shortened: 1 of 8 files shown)


libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
```diff
@@ -0,0 +1,168 @@
...
+} from "./code-gen-types";
+import { IFile } from "./files";
+import { EventParams } from "./plugins.types";
+import { AstNode } from "@amplication/csharp-ast";
+
+export interface CreateEntityServiceBaseParams extends EventParams {
+  entity: Entity;
+  resourceName: string;
+  apisDir: string;
+  file: IFile<AstNode>;
```
overbit, line 15:
```suggestion
  file: IFile<AstNode>;
```

AstNode is a great solution when we want a super generic solution but passing a more "expected" type like CSharp file that can be of type `Interface`, `Class` or `Enum` will make plugin experience much easier as it would be possible to get properties by name.

i.e.
```cs 
// IFile<Class> file will allow us to extend easily the obj as follow

const createMethod = file.getMethod("createX");

createMethod.AddAnnotation({ /.../ });

//...
```

libs/util/code-gen-types/src/dotnet-plugin-events.types.ts
```diff
@@ -0,0 +1,58 @@
...
+  CreateServerGitIgnoreParams,
+  CreateServerParams,
+  CreateServerSecretsManagerParams,
+  CreateSwaggerParams,
+  LoadStaticFilesParams,
+} from "./dotnet-plugin-events-params.types";
+import { DotnetEventNames, PluginEventType } from "./dotnet-plugins.types";
+
+export type DotnetEvents = {
+  [DotnetEventNames.CreateServerAuth]?: PluginEventType<CreateServerAuthParams>;
```
overbit, line 32: Do we need this if we move to just auth plugins?

libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
overbit: Outdated. We will have types per event that will improve devex

overbit (approved):

Issue #8472, opened 2024-05-13
Jovu: Allow creation of service with an initial set of plugins


yuval-hazaz:
### Feature description

Allow Jovu to install plugins in the same function with the service creation 

### Use case

it will encourage Jovu to install more plugins with new services

yuval-hazaz:
✅- passed QA

ha466:
Hi

PR #8473, opened 2024-05-13
Feat/jovu create serviec with plugins

yuval-hazaz:
Close: #8472

## PR Details

- Add a new param for createService for a list of plugin IDs
- remove the params for base paths, and use default values instead 
- do not install the default DB plugin when jovu asks to install any DB plugin 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/assistant/functions/createService.json
+++ b/packages/amplication-server/src/core/assistant/functions/createService.json
@@ -1,6 +1,6 @@
 {
   "name": "createService",
-  "description": "Create a service in a project. When using this function, it is best to also create a few default entities and fields unless specifically requested not to do so",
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/assistant/functions/createService.json
mulygottlieb, line 1: we need to manually update this in the default project for sandbox and then in the production project for staging+prod (in OpenAI)

mulygottlieb (approved):
See my comment on updating OpenAI platform manually

packages/amplication-server/src/core/assistant/functions/createService.json
yuval-hazaz: yep

PR #8549, opened 2024-05-26
Feat(server): handle auth entity for jovu

yuval-hazaz:
Close: https://github.com/amplication/private-issues/issues/191

## PR Details

1. When jovu creates a new entity with the name "User" or "Users" (or any upper or lower case version of that) - we create the default user entity with all the default fields, and set it as the auth entity, and we also let jovu know that we created the default fields (so it will not try to create the same fields again)
  - if another auth entity already exist we create the new entity as a regular entity with the requested name

2. When jovu installs a plugin that requires auth entity, 
 - if there is already auth entity - all good (probably it alreadt created an entity with the name "User" or in case we are working on an existing service).
 - if there is no auth entity - and the name "User" is available - we create a new user entity with all default fields and set the auth entity
 - if the name "user" is already take - we throw an error back to jovu

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
+++ b/packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
@@ -2,9 +2,9 @@ import { EnumModuleDtoType } from "@amplication/code-gen-types";
 import { AmplicationLogger } from "@amplication/util/nestjs/logging";
 import { Inject, Injectable } from "@nestjs/common";
 import { ConfigService } from "@nestjs/config";
-import { get } from "lodash";
+import { get, isEmpty } from "lodash";
 import { pascalCase } from "pascal-case";
-import { plural } from "pluralize";
+import { plural, singular } from "pluralize";
 import { AuthorizableOriginParameter } from "../../enums/AuthorizableOriginParameter";
 import { Env } from "../../env";
 import { Block } from "../../models";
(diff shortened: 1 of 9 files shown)


packages/amplication-server/src/core/entity/constants.ts
```diff
@@ -100,7 +100,7 @@ type EntityData = Omit<
 
 export const DEFAULT_SINGLE_LINE_TEXT_MAX_LENGTH = 256;
 
-export const DEFAULT_ENTITIES: EntityData[] = [
+export const DEFAULT_USER_ENTITY: EntityData[] = [
```
mulygottlieb, line 103: Why is this still an array then?

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -758,12 +758,12 @@ export class EntityService {
...
   }
 
-  async createDefaultEntities(
+  async createDefaultUserEntity(
     resourceId: string,
     user: User
   ): Promise<Entity[]> {
     return await Promise.all(
-      DEFAULT_ENTITIES.map(async (entity) => {
+      DEFAULT_USER_ENTITY.map(async (entity) => {
```
mulygottlieb, line 766: See my previous comment, if it's not an array, you can get rid of the .map() here.

packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
```diff
@@ -286,6 +292,27 @@ export class AssistantFunctionsService {
...
+      if (
+        configurations &&
+        configurations[REQUIRES_AUTHENTICATION_ENTITY] === "true" &&
+        !authEntityExist
+      ) {
+        const authEntityName = await this.resourceService.getAuthEntityName(
+          serviceId,
+          context.user
+        );
+        if (!isEmpty(authEntityName)) {
```
mulygottlieb, line 304: I would write `if (isEmpty())` and add the `authEntityExist = true` at the end after the if (since it's done in both cases and the code would be more readable IMO)

packages/amplication-server/src/core/entity/constants.ts
yuval-hazaz: because it required a big change e2e including the APIs and the client and I didn't want to include it

packages/amplication-server/src/core/entity/entity.service.ts
yuval-hazaz: I suggest to keep it as an array for now for risk management reasons

packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
yuval-hazaz: thanks, i will pass this time

mulygottlieb (approved):

Issue #8356, opened 2024-04-30
Apply BTM changes fails


PazYanoverr:
When trying to apply changes, the following error appeared-

![Image](https://github.com/amplication/amplication/assets/149934977/f747eae4-3400-4faf-abd8-07619454d959)

(Failed to Yuval as well)

PazYanoverr:
Passed QA (Sandbox)-
![image](https://github.com/amplication/amplication/assets/149934977/53ea71a1-1db8-48c7-8ba0-074e590c3ed3)

PR #8551, opened 2024-05-26
(Fix): server 8356=> BTM apply changes failed

morhag90:
Close: #8356

## PR Details

Fix BTM apply changes failed. 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -836,6 +836,13 @@ export class EntityService {
         );
       }
 
+      for (const relatedEntityField of relatedEntityFields) {
+        await this.deleteField(
+          { where: { id: relatedEntityField.id } },
+          user,
+          fieldStrategy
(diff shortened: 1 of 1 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -836,6 +836,13 @@ export class EntityService {
         );
       }
 
+      for (const relatedEntityField of relatedEntityFields) {
```
yuval-hazaz, line 839: can you please explain this change in a couple of words? how does it fix the issue?

yuval-hazaz (approved):

packages/amplication-server/src/core/entity/entity.service.ts
morhag90: @yuval-hazaz  In the deleteRelatedField function, we are trying to get the Module by the entity id, and it can't find it because we have already deleted the entity module.

Issue #8567, opened 2024-05-29
[Epic:`.NET`] add CreateEntityModel plugin event


morhag90:

mulygottlieb:
Passed QA ✅
Tested in: https://github.com/muly-demo-repos/dotnet-tests-1

PR #8568, opened 2024-05-29
feat(code-gen-types): add CreateEntityModel plugin event

morhag90:
Close: #8567 

## PR Details

Add CreateEntityModel plugin event.

## PR Checklist

### DIFF
--- a/libs/util/code-gen-types/src/dotnet-plugins.types.ts
+++ b/libs/util/code-gen-types/src/dotnet-plugins.types.ts
@@ -127,6 +127,7 @@ export enum DotnetEventNames {
   CreateServerSecretsManager = "CreateServerSecretsManager",
   CreateEntityInterface = "CreateEntityInterface",
   CreateEntityExtensions = "CreateEntityExtensions",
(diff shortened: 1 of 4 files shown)


libs/util/code-gen-types/src/dotnet-plugins.types.ts
```diff
@@ -127,6 +127,7 @@ export enum DotnetEventNames {
   CreateServerSecretsManager = "CreateServerSecretsManager",
   CreateEntityInterface = "CreateEntityInterface",
   CreateEntityExtensions = "CreateEntityExtensions",
+  CreateEntityModel = "CreateEntityModel",
```
overbit, line 130: where is the plugin event internal function for this event?

morhag90: @overbit  Will be implemented on a separated PR from the Dotnet DSG

overbit: Of course. Silly me

overbit (approved):

PR #8578, opened 2024-05-30
feat(code-gen-types): add createResourceDbContextFile plugin event

morhag90:
Close: #8577 
## PR Details

Add createResourceDbContextFile plugin event. 

## PR Checklist

### DIFF
--- a/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
+++ b/libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
@@ -63,6 +63,12 @@ export interface CreateEntityModelParams extends EventParams {
   apisDir: string;
 }
 
+export interface createResourceDbContextFileParams extends EventParams {
+  entities: Entity[];
+  resourceName: string;
+  apisDir: string;
+}
+
 export interface CreateEntityGrpcControllerParams extends EventParams {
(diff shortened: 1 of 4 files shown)


libs/util/code-gen-types/src/dotnet-plugin-events-params.types.ts
```diff
@@ -63,6 +63,12 @@ export interface CreateEntityModelParams extends EventParams {
   apisDir: string;
 }
 
+export interface createResourceDbContextFileParams extends EventParams {
+  entities: Entity[];
+  resourceName: string;
+  apisDir: string;
```
overbit, line 69: Is `apisDir` needed? Or should it be just context folder?

morhag90: @overbit  We are passing it to the internal method, so I think we need to pass it to the plugin event too:

image

overbit: Let's use the same name then. Apisdir is a bit cryptic

morhag90: Done

overbit (approved):

PR #8581, opened 2024-05-31
Feat/code gen dotnet

yuval-hazaz:
Close: https://github.com/amplication/private-issues/issues/189

## PR Details

- Add support for the selection of the code generator when creating a resource
- Select the code generator by default based on the workspace permissions for Jovu and other services that are created automatically (preview user, BTM)
- Allow users with access to select the type of code generator when creating a new service
- Show the type of code generator in the resource list, resource home, and build list 

## PR Checklist


packages/amplication-client/src/Resource/constants.ts
```diff
@@ -161,6 +162,7 @@ export function prepareMessageBrokerObject(
     name: "My message broker",
     description: "",
     resourceType: models.EnumResourceType.MessageBroker,
+    codeGenerator: models.EnumCodeGenerator.NodeJs,
```
mulygottlieb, line 165: Since `codeGenerator` is optional in `Resource`, and we are in a Message Broker, isn't it better to leave it undefined (not to add this line)?

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
```diff
@@ -29,4 +30,9 @@ export class ResourceCreateInput {
 
   @Field(() => ConnectGitRepositoryInput, { nullable: true })
   gitRepository?: ConnectGitRepositoryInput;
+
+  @Field(() => EnumCodeGenerator, {
+    nullable: false,
```
mulygottlieb, line 35: Shouldn't this be `nullable : true` ?

libs/util/billing-types/src/lib/billing-feature.types.ts
```diff
@@ -6,7 +6,7 @@ export enum BillingFeature {
   BranchPerResource = "feature-branch-per-resource",
   ChangeGitBaseBranch = "feature-change-git-base-branch",
   CodeGenerationBuilds = "feature-code-generation-builds",
-  CodeGeneratorName = "feature-code-generator-name",
+  CodeGeneratorDotNet = "feature-code-generator-dotnet",
```
mulygottlieb, line 9: Writing this as a note to myself before merging this PR - we need to create the new entitlement on Stigg prod (and remove the old entitlement no longer in use)

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -384,44 +415,47 @@ export class ResourceService {
...
+        return EnumCodeGenerator.DotNet;
+      }
 
-    return this.prisma.resource.update({
-      where: args.where,
-      data: {
-        codeGeneratorName: args.codeGeneratorName,
-      },
-    });
+      return EnumCodeGenerator.NodeJs;
```
mulygottlieb, line 457: Should this line be outside the `if (license)` block?

packages/amplication-client/src/Resource/constants.ts
yuval-hazaz: the value on the server side is non-nullable so it will be easier to maintain this feature on the long run, for now and as long as we don't have other generator, the default is anyway "Node.js" 
so, I thought that for now, this should be a more maintainable approach that will prevent drifting from the original design as we add more resources

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
yuval-hazaz: see my previous reply

yuval-hazaz:
@mulygottlieb I made some changes based on your review. Please review again. thanks!

packages/amplication-client/src/Workspaces/ResourceListItem.tsx
```diff
@@ -121,7 +121,8 @@ function ResourceListItem({ resource, onDelete }: Props) {
               >
                 {name}
               </Text>
-              <CodeGeneratorImage codeGenerator={codeGenerator} />
+
+              <CodeGeneratorImage resource={resource} />
```
mulygottlieb, line 125: I think you can remove `codeGenerator` from the spread in line 37

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
mulygottlieb: I guess it's confusing since below the actual member `codeGenerator` is defined with a `?` but if that's the right way to define it, feel free to resolve this comment

libs/util/billing-types/src/lib/billing-feature.types.ts
mulygottlieb: Done:
image

packages/amplication-server/src/core/resource/dto/ResourceCreateInput.ts
yuval-hazaz: You are correct, but it means that the property is not actually required on the type, but it is required in the API via gql
we have a fallback for undefined value, so we can support undefined, but we don't want to increase the gap so we make it required from the client

mulygottlieb (approved):

mulygottlieb (approved):

PR #8630, opened 2024-06-10
feat(server):add essential plan

yuval-hazaz:
Close: #8631

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Purchase/PurchasePage.tsx
+++ b/packages/amplication-client/src/Purchase/PurchasePage.tsx
@@ -11,6 +11,7 @@ import {
   EnumButtonStyle,
   EnumIconPosition,
   Modal,
+  Snackbar,
 } from "@amplication/ui/design-system";
 import "./PurchasePage.scss";
 import { useCallback, useContext, useState } from "react";
@@ -22,6 +23,8 @@ import { PROVISION_SUBSCRIPTION } from "../Workspaces/queries/workspaceQueries";
 import { PurchaseLoader } from "./PurchaseLoader";
 import { FAQ } from "./FAQ";
 import { GET_CONTACT_US_LINK } from "../Workspaces/queries/workspaceQueries";
+import { formatError } from "../util/error";
+import { set } from "lodash";
 
 export type DType = {
   provisionSubscription: models.ProvisionSubscriptionResult;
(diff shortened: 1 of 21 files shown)


packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -22,6 +23,8 @@ import { PROVISION_SUBSCRIPTION } from "../Workspaces/queries/workspaceQueries";
 import { PurchaseLoader } from "./PurchaseLoader";
 import { FAQ } from "./FAQ";
 import { GET_CONTACT_US_LINK } from "../Workspaces/queries/workspaceQueries";
+import { formatError } from "../util/error";
+import { set } from "lodash";
```
overbit, line 27: This import is not used

packages/amplication-client/src/Purchase/PurchasePage.tsx
```diff
@@ -76,17 +81,34 @@ const PurchasePage = (props) => {
...
+    {
+      loading: provisionSubscriptionLoading,
+      error: provisionSubscriptionError,
+    },
+  ] = useMutation<DType>(PROVISION_SUBSCRIPTION, {
+    onCompleted: (data) => {
+      const { provisionStatus, checkoutUrl } = data.provisionSubscription;
+      if (provisionStatus === "PAYMENT_REQUIRED") {
+        window.location.href = checkoutUrl;
+      } else if (provisionStatus === "SUCCESS") {
```
overbit, line 95: Ideally we should use the Enum for provisionStatus and not an hardcoded string.

Update server to return the correct enum in `ProvisionSubscriptionResult` as follow:

```ts
import { Field, ObjectType, registerEnumType } from "@nestjs/graphql";
import { ProvisionSubscriptionStatus } from "@stigg/node-server-sdk";

@ObjectType({
  isAbstract: true,
})
export class ProvisionSubscriptionResult {
  @Field(() => ProvisionSubscriptionStatus, {
    nullable: false,
  })
  provisionStatus!: ProvisionSubscriptionStatus;

  @Field(() => String, {
    nullable: true,
  })
  checkoutUrl?: string;
}

registerEnumType(ProvisionSubscriptionStatus, {
  name: "ProvisionSubscriptionStatus",
});
```

packages/amplication-client/src/Workspaces/hooks/useUpgradeButtonData.ts
overbit, line 1: @yuval-hazaz check what these changes will do to the Pro users as they will experience different behaviours

packages/amplication-client/src/Purchase/PurchasePage.tsx
yuval-hazaz: I agree with the need but prefer not to go with this change in the current scope. 
I will keep the existing code as it is, and we can consider making the change in a separate scope after considering the impact and different options

yuval-hazaz: I opened a new issue
https://github.com/amplication/amplication/issues/8639

overbit (approved):

PR #8682, opened 2024-06-17
fix(server): issue related to billing features not being respected

overbit:
Fixes: https://github.com/amplication/private-issues/issues/197

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/entity/entity.service.ts
+++ b/packages/amplication-server/src/core/entity/entity.service.ts
@@ -272,6 +272,39 @@ export class EntityService {
     }
   }
 
+  private async checkServiceEntityLicense(resource: Resource) {
+    if (!this.billingService.isBillingEnabled) {
+      return;
+    }
+    await this.checkServiceLicense(resource);
+
+    const serviceEntityEntitlementPromise =
+      this.billingService.getNumericEntitlement(
+        resource.project.workspaceId,
+        BillingFeature.EntitiesPerService
+      );
+
+    const resourceEntitiesCount = this.prisma.entity.count({
+      where: {
+        resourceId: resource.id,
+        deletedAt: null,
+      },
+    });
+
+    const serviceEntityEntitlement = await serviceEntityEntitlementPromise;
+    if (
+      !serviceEntityEntitlement.hasAccess ||
+      (!serviceEntityEntitlement.isUnlimited &&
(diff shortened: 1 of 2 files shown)


packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -272,6 +272,39 @@ export class EntityService {
     }
   }
 
+  private async checkServiceEntityLicense(resource: Resource) {
+    if (!this.billingService.isBillingEnabled) {
+      return;
+    }
+    await this.checkServiceLicense(resource);
```
morhag90, line 279: @overbit  Why do we need serviceCheck in the entityCheck method?

packages/amplication-server/src/core/entity/entity.service.ts
```diff
@@ -272,6 +272,39 @@ export class EntityService {
...
+      );
+
+    const resourceEntitiesCount = this.prisma.entity.count({
+      where: {
+        resourceId: resource.id,
+        deletedAt: null,
+      },
+    });
+
+    const serviceEntityEntitlement = await serviceEntityEntitlementPromise;
```
morhag90, line 294: @overbit  Why not do the await in the functions calls?

packages/amplication-server/src/core/entity/entity.service.ts
overbit: to avoid duplicating code as it was there before

packages/amplication-server/src/core/entity/entity.service.ts
overbit: has prisma and http call are not dependent on each other, we are going to await only when we need their results

morhag90: This const : serviceEntityEntitlementPromise not related to prisma, 

I mean: 
  const serviceEntityEntitlementPromise = **await** 
      this.billingService.getNumericEntitlement(
        resource.project.workspaceId,
        BillingFeature.EntitiesPerService
      );

Am I missing something??

overbit: The code is firing the following 2 promises: `this.billingService.getNumericEntitlement` and `this.prisma.entity.count`.
As they are not dependent on each other, we can have the await later where is needed to avoid blocking the async execution on one of these statements

morhag90 (approved):

PR #8688, opened 2024-06-17
fix(dsg): validate min and max

abrl91:
Close: https://github.com/amplication/private-issues/issues/169

## PR Checklist


packages/data-service-generator/src/server/resource/dto/create-field-class-property.ts
```diff
@@ -156,6 +156,17 @@ export const PARSE_ID = builders.identifier("parse");
 export const IS_ARRAY_ID = builders.identifier("isArray");
 export const NULLABLE_ID = builders.identifier("nullable");
 
+function createMinMaxDecorator(
+  identifierName: string,
+  minValue: number
```
mulygottlieb, line 161: Since this value can be for minValue, maxValue, or maxLength value, I would rename the function parameter from "minValue" to something more generic like "decoratorValue" or just "value" (if not a reserved word).

abrl91: yeah, in the begining I created 2 function one for min and one for max and I forgot to rename the variable :)

mulygottlieb (approved):

Issue #8556, opened 2024-05-28
[Epic:`.NET`] Align all the build-in DTOs that appear in the UI under the entity-related-modules (e.g. CustomerFindManyArgs) VS the actual dotnet DTOs


mulygottlieb:
We need to decide which "way" to align - either change the dotnet generated code to use the same DTOs we use for Nest (that appear in the UI), or - change the auto-created-DTOs in the server according to the chosen tech-stack for the service (align the blocks and UI to the actual generated code).
This needs to be discussed internally first (Mango)

overbit:
This needs to be reviewed as the current generation is not using the same DTOs @mulygottlieb @morhag90

mulygottlieb:
Moving this back into "Todo" until we have the discussion which side to align to which side

mulygottlieb:
Tech notes for implementation:
Server:
- [ ] Filter out GQL-only DTOs in the server in the ModuleDTO Resolver
- [ ] 
Dotnet DSG:
- [ ] Rename non-GQL DTOs to the names we have in the server
- [ ] Change the [Entity]IdDto - rename, do not inherit in the [Entity] DTO and instead add the id with its id-type directly

morhag90:
Passed QA:

Forbbiden: 
<img width="1469" alt="image" src="https://github.com/amplication/amplication/assets/97830649/78d96590-c02e-4278-97af-681e3c4b17e7">

Authorize + FindOne => Not Found:
<img width="1466" alt="image" src="https://github.com/amplication/amplication/assets/97830649/04ed5aac-9983-4ad4-9d51-7af8fc9ba26d">

Create One entity:

<img width="1309" alt="image" src="https://github.com/amplication/amplication/assets/97830649/b2f6f85e-9299-4658-8670-3ffafb0d0c64">

FindMany:

<img width="1245" alt="image" src="https://github.com/amplication/amplication/assets/97830649/58a0570f-aeda-4fdc-a1d2-1fcb60af7d7a">

Get one entity:

<img width="1386" alt="image" src="https://github.com/amplication/amplication/assets/97830649/c682b379-4ed8-4ec3-bdf5-47bdf5c3ddf0">
 
Create one entity with another entity relation:

<img width="1468" alt="image" src="https://github.com/amplication/amplication/assets/97830649/7b1ec613-d39f-4456-a263-bda5a6eecc50">
 
The other side:

[...]

PR #8713, opened 2024-06-25
Filter-nest-only-dtos

mulygottlieb:
fixes: #8556

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
+++ b/packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
@@ -4,7 +4,7 @@ import {
   getDefaultDtosForEnumField,
   getDefaultDtosForRelatedEntity,
 } from "@amplication/dsg-utils";
-import { Injectable } from "@nestjs/common";
+import { Inject, Injectable, forwardRef } from "@nestjs/common";
 import { UserEntity } from "../../decorators/user.decorator";
 import { EnumBlockType } from "../../enums/EnumBlockType";
 import { AmplicationError } from "../../errors/AmplicationError";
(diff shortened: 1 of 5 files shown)


packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
```diff
@@ -137,8 +157,34 @@ export class ModuleDtoService extends BlockTypeService<
         "AND"
       );
     } else {
-      return [];
+      allDTOs = [];
     }
+
+    const resourceId = args.where.resource?.id;
+    return this.filterDtosByGeneratorName(resourceId, allDTOs);
```
morhag90, line 164: @mulygottlieb  await?

packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
```diff
@@ -137,8 +157,34 @@ export class ModuleDtoService extends BlockTypeService<
...
+
+    if (
+      !resource ||
+      (resource &&
+        CODE_GENERATOR_NAME_TO_ENUM[resource.codeGeneratorName] !==
+          EnumCodeGenerator.DotNet)
+    )
+      return allDTOs;
+
+    return allDTOs.filter((dto) => {
```
morhag90, line 183: @mulygottlieb, you aren't using the results, and it looks like you don't override the original array

packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
mulygottlieb: no need since this is an async function that calls another async function so it returns the promise

packages/amplication-server/src/core/moduleDto/moduleDto.service.ts
mulygottlieb: what do you mean? I return the result from the .filter() call so the return value is the filtered list

morhag90 (approved):

Issue #8741, opened 2024-06-30
Jovu should "understand" failure of entity creation in case maximum allowed entities reached


PazYanoverr:
As for now, Jovu understands a failure of **service** creation in case the maximum allowed services reached -

![Image](https://github.com/amplication/amplication/assets/149934977/c393eb1e-2264-43e0-90d2-8e2619b37ecc)

However, it does not understand a failure of **entity** creation in case the limit is reached-

![Image](https://github.com/amplication/amplication/assets/149934977/a0da76b5-c59d-4a55-a754-ad5957965f65)

morhag90:
Passed QA:

<img width="337" alt="image" src="https://github.com/amplication/amplication/assets/97830649/6aa603bd-163c-44cf-8170-fb7ab9a16497">

PR #8749, opened 2024-07-02
fix(server): Jovu: error create entity limitation

morhag90:
Close: #8741

## PR Details

Jovu: error create entity limitation.

## PR Checklist


packages/amplication-server/src/core/assistant/assistantFunctions.service.ts
```diff
@@ -448,7 +448,7 @@ export class AssistantFunctionsService {
           } catch (error) {
             this.logger.error(
               `Chat: Error creating entity ${entityName}`,
```
mulygottlieb, line 450:
```suggestion
              `Chat: Error creating entity ${entityName} with error: ${error.message}`,
```

PR #8758, opened 2024-07-08
feat(client): add analytics to upgrade button: .NET promote page

morhag90:
Close: https://github.com/amplication/private-issues/issues/210

## PR Details

Add analytics to upgrade button: .NET promote.

## PR Checklist


packages/amplication-client/src/VersionControl/DotNetPromotePage.tsx
```diff
@@ -38,6 +42,14 @@ const DotNetPromotePage = () => {
     history.push(`/${currentWorkspace?.id}/${currentProject?.id}`);
   }, [history, currentProject, currentWorkspace]);
 
+  const handleUpgradeClick = useCallback(() => {
+    window.open(`/${currentWorkspace?.id}/purchase`, "_blank");
```
mulygottlieb, line 46: shouldn't we first track the event and only then call the `window.open()` call?

mulygottlieb (approved):

Issue #8500, opened 2024-05-16
Generate code per service


PazYanoverr:
General: We would like to have the following build / code generation options:
1. Generate code for all services
2. Generate code for all services that have pending changes
3. Generate code for a specific service

Detailed request:
1. When pressing the "Generate the code" option- the code will be generated for all the services
2. The "Generate the code" will also have an "arrow" option that will open a menu with the following options:
    a. Generate the code for all services
    b. Generate the code for services with changes
    c. Generate the code for service [name]
    d. Generate the code for service [name]
3. In case the user chooses option c- and there are pending changes for other services, a message should appear saying-
"**Note:** There are pending changes for other services as well. 
Generating the code for service '[name]' will update all services with their changes, but only the code for the chosen service will be generated."
Buttons: Cancel, Generate
4. We need to modify the commit event to include the chosen type of commit out of the options above (add new metadata to the existing event) - preferably on the server side event

abrl91:
@PazYanoverr 
what about having an icon to "generate code" on each service (like the delete button)
It can be disabled if there are no pending changes.

<img width="395" alt="image" src="https://github.com/amplication/amplication/assets/39680385/5de9468d-82ad-40aa-aceb-d41079231da3">

PazYanoverr:
@abrl91 - It's a good idea! The only problem with it, is that when generating the code for service A, the pending changes and other changes (DSG-originated and such) will be applied for other services as well. I am afraid it's a little bit confusing, so I think we should try to avoid this path in a prominent location.  
WDYT?

mulygottlieb:
Requires completing the client side

mulygottlieb:
@PazYanoverr I am moving this back to "todo" for the client side (server is done)

mulygottlieb:
Passed QA ✅

- Commit only pending changes (1 service out of two):
https://github.com/user-attachments/assets/b7e41086-10df-4e4f-9927-bcfff66836e9


- Commit all services by choosing all in the drop down + verify the "pending changes only" option does not appear when no changes:
https://github.com/user-attachments/assets/8f193b97-ef40-48b4-8f03-7129c703f19a


- Commit all services by just hitting the button itself (default behavior):
https://github.com/user-attachments/assets/b9a0f3ca-f6ae-4696-b6c8-93e0d161deb2


- Commit a single service:
https://github.com/user-attachments/assets/c0a697e2-4fa2-45e0-83fd-f8fa2fa27a78

PR #8768, opened 2024-07-14
Feat(client): Add options list for commit

morhag90:
Close: #8500 

## PR Details

Add options list for commit: 
- all services
- all changed services
- specific service 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
+++ b/packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
@@ -89,7 +89,7 @@ const SyncWithGithubPage: React.FC = () => {
   const handleCommit = useCallback(() => {
     commitChanges({
       message: "",
-      projectId: currentProject.id,
+      project: { connect: { id: currentProject?.id } },
       bypassLimitations: false,
     });
     setOpenPr(false);
--- a/packages/amplication-client/src/VersionControl/Commit.tsx
+++ b/packages/amplication-client/src/VersionControl/Commit.tsx
@@ -49,8 +49,8 @@ export type TCommit = {
 
 const INITIAL_VALUES: TCommit = {
   message: "",
-  selectedService: null,
   commitStrategy: EnumCommitStrategy.All,
+  selectedService: null,
 };
 
 type Props = {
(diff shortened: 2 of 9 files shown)


packages/amplication-client/src/Resource/git/SyncWithGithubPage.tsx
```diff
@@ -94,7 +94,7 @@ const SyncWithGithubPage: React.FC = () => {
   const handleCommit = useCallback(() => {
     commitChanges({
       message: "",
-      projectId: currentProject.id,
+      project: { connect: { id: currentProject?.id } },
```
mulygottlieb, line 97: In all other places in the client, we pass a `projectId` to `commitChanges` so I'm not sure why only here it's different - and if indeed this is the right way to pass this in, then it's better to make this "transformation" inside the hook and not in the caller

packages/amplication-client/src/VersionControl/Commit.tsx
```diff
@@ -88,59 +136,182 @@ const Commit = ({
...
+                          <SelectMenuList>
+                            {COMMIT_STRATEGY_OPTIONS.map((item, index) => (
+                              <CreateCommitStrategyButtonItem
+                                key={index}
+                                item={item}
+                                hasPendingChanges={pendingChanges?.length > 0}
+                                onCommitStrategySelected={
+                                  handleOnSelectCommitStrategyChange
+                                }
+                              ></CreateCommitStrategyButtonItem>
```
mulygottlieb, line 295:
```suggestion
                              />
```

packages/amplication-client/src/VersionControl/Commit.tsx
```diff
@@ -88,59 +136,182 @@ const Commit = ({
...
+                  <MultiStateToggle
+                    className={`${CLASS_NAME}__technology-toggle`}
+                    label=""
+                    name="action_"
+                    options={OPTIONS}
+                    onChange={handleOnSelectLanguageChange}
+                    selectedValue={"node"}
+                  />
+                )}
+                {commitBtnType === CommitBtnType.Button ? (
```
mulygottlieb, line 264: If I understand correctly, you are using the commitBtnType to "understand" in which page you are in right now? and then show or hide the selectMenu options?
If that is the case, let's at least create a small utility method that says if we need to add the SelectMenu or not - inside the function it can still use the commitBtnType to decide but at least it will be encapsulated - when it appears here in this way it does look a bit like a patch or some "code smell"

morhag90: @mulygottlieb Condition removed (Irrelevant anymore).

packages/amplication-client/src/VersionControl/Commit.tsx
morhag90: @mulygottlieb  How? This component has props

mulygottlieb (approved):

PR #8774, opened 2024-07-16
fix(dsg): add stringLiteral("args") for primitive input type in resolver

morhag90:
Close: https://github.com/amplication/private-issues/issues/211

## PR Details

Add stringLiteral("args") for primitive input type in resolver.

## PR Checklist

### DIFF
--- a/packages/data-service-generator/src/utils/create-graphql-args-decorator.ts
+++ b/packages/data-service-generator/src/utils/create-graphql-args-decorator.ts
@@ -2,6 +2,9 @@ import { builders, namedTypes } from "ast-types";
 
 const ARGS_ID = builders.identifier("graphql.Args");
 
-export function createGraphqlArgsDecorator(): namedTypes.Decorator {
-  return builders.decorator(builders.callExpression(ARGS_ID, []));
+export function createGraphqlArgsDecorator(
+  hasArgument = false
+): namedTypes.Decorator {
+  const args = hasArgument ? [builders.stringLiteral("args")] : [];
+  return builders.decorator(builders.callExpression(ARGS_ID, args));
 }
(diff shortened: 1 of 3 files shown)


packages/data-service-generator/src/utils/create-graphql-args-decorator.ts
```diff
@@ -2,6 +2,9 @@ import { builders, namedTypes } from "ast-types";
 
 const ARGS_ID = builders.identifier("graphql.Args");
 
-export function createGraphqlArgsDecorator(): namedTypes.Decorator {
-  return builders.decorator(builders.callExpression(ARGS_ID, []));
+export function createGraphqlArgsDecorator(
+  hasArgument = false
+): namedTypes.Decorator {
+  const args = hasArgument ? [builders.stringLiteral("args")] : [];
```
mulygottlieb, line 8: Isn't it better, instead of passing a Boolean (hasArgument) to pass an optional string which is the name of the argument ("args") and then if the string is passed (and not null / empty string) to add the stringLiteral argument?

I'm asking since this is now hard coding the use of "args" as the name of the parameter (perhaps that's okay, just bringing up the question).

In addition, I think that the name "hasArgument" is a bit misleading since there is always an argument, it's just in most cases a DTO and not a primitive, so perhaps - if using a Boolean - the better name would be "isPrimitiveArgs"

morhag90: @mulygottlieb It is an args decorator, so I think it's fine to leave it hard-coded.
Param rename to: isPrimitiveArgs

mulygottlieb (approved):

Issue #8932, opened 2024-09-14
Move private plugin settings to a tab on the project page


yuval-hazaz:
### Feature description

Move private plugin settings to a tab on the project page

PR #8933, opened 2024-09-14
feat(client):move private plugins to project tabs

yuval-hazaz:
Close: https://github.com/amplication/amplication/issues/8932

## PR Details

Move the private plugins' settings to a separate tab on the project page. 

## PR Checklist

### DIFF
--- a/packages/amplication-client/src/Components/CreateResourceButton.tsx
+++ b/packages/amplication-client/src/Components/CreateResourceButton.tsx
@@ -38,13 +38,6 @@ const ITEMS: CreateResourceButtonItemType[] = [
     route: "create-broker",
     info: "Create a message broker to facilitate communication between services",
   },
-  {
-    type: models.EnumResourceType.PluginRepository,
-    label: "Plugin Repository",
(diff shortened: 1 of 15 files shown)


yuval-hazaz:
@PazYanoverr  this is the first step just to make things clearer and remove the resource from the resource list. 
still to be done:
1. decide and validate the license to access the tab + show indications if needed
2. create the pluginRepo resource with the project by default 
3. prevent deletion of the pluginRepo resource
4.

PazYanoverr:
@yuval-hazaz - 
1. I think that for now, we should keep the same license (entitlement), while users who do not have the license should be able to enter the new tab with a clear message that they should upgrade to Enterprise to use this feature.
2. As discussed- maybe it's not required, and users who do not have any pluginRepo (and have the relevant entitlement) will be "redirected" to the wizard of pluginRepo creation.
3. See # 2. If we go with this behavior, I think we can allow deletion of the last pluginRepo.

morhag90 (approved):

PR #8943, opened 2024-09-16
fix(server): allow workspace creation for the first workspace (new user) regardless of entitlement

mulygottlieb:
Fixes: https://github.com/amplication/private-issues/issues/219

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/workspace/workspace.service.ts
+++ b/packages/amplication-server/src/core/workspace/workspace.service.ts
@@ -191,10 +191,12 @@ export class WorkspaceService {
   async createWorkspace(
     accountId: string,
     args: Prisma.WorkspaceCreateArgs,
+    isNewUserInitialWorkspace = false,
     currentWorkspaceId?: string,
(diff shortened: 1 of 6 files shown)


packages/amplication-server/src/core/workspace/workspace.service.ts
```diff
@@ -191,10 +191,12 @@ export class WorkspaceService {
   async createWorkspace(
     accountId: string,
     args: Prisma.WorkspaceCreateArgs,
+    isNewUserInitialWorkspace = false,
```
morhag90, line 194: @mulygottlieb, you can pass this parameter last, so you will not need to send it from the resolver

morhag90 (approved):

packages/amplication-server/src/core/workspace/workspace.service.ts
mulygottlieb: There are optional args after so it's not possible to pass it last as much as I know

PR #8978, opened 2024-09-23
Epic: Resource Versions

yuval-hazaz:
Close: #[issue-number]

## PR Checklist


packages/amplication-server/src/core/block/block.service.ts
```diff
@@ -942,4 +942,35 @@ export class BlockService {
...
+      where: {
+        ...args.where,
+        resourceId: args.where.resourceId,
+        deletedAt: null,
+      },
+      include: {
+        versions: {
+          where: {
+            versionNumber: {
+              not: CURRENT_VERSION_NUMBER,
```
mulygottlieb, line 959:
```suggestion
              gt: CURRENT_VERSION_NUMBER,
```

packages/amplication-server/src/core/block/block.service.ts
```diff
@@ -942,4 +942,35 @@ export class BlockService {
...
+          take: 1,
+          orderBy: {
+            versionNumber: Prisma.SortOrder.desc,
+          },
+        },
+        parentBlock: true,
+      },
+    });
+
+    return (await blocks).map((block) => {
```
mulygottlieb, line 971: Why do you need to await blocks again? you already await the findMany call above

packages/amplication-server/src/core/block/block.service.ts
yuval-hazaz: CURRENT_VERSION_NUMBER is 0 - so both will give the same results.
we are using not in entities so i prefer to keep the same code
the latest version will always be larger then zero as long as there is a version - so the filter is actually "give me top 1 version from all published versions"

packages/amplication-server/src/core/resourceVersion/resourceVersion.service.ts
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+    const logger = this.logger.child({
+      resourceVersionId: resourceVersion.id,
+      resourceId: resourceVersion.resourceId,
+      userId: resourceVersion.userId,
+      user,
+    });
+
+    const resource = await this.resourceService.findOne({
```
mulygottlieb, line 76: Don't you want to check this before creating the version in the DB at line 51?

mulygottlieb (approved):

packages/amplication-server/src/util/version.ts
```diff
@@ -1 +1 @@
-export const version = "3.3.2";
+export const version = "3.2.3";
```
yuval-hazaz, line 1:
```suggestion
export const version = "3.3.2";

```

packages/amplication-client/src/util/version.ts
```diff
@@ -1 +1 @@
-export const version = "3.3.2";
+export const version = "3.2.3";
```
yuval-hazaz, line 1:
```suggestion
export const version = "3.3.2";

```

Issue #8997, opened 2024-09-30
Add BuildStatus and BuildGitStatus fields on Build model


yuval-hazaz:
### Feature description

instead of using a dynamic property for buildStatus - add a new field on the build model to keep the current status of the build.
also, add a new BuildGitStatus to hold the status of the build sync with git

mulygottlieb:
Passed QA ✅

PR #9003, opened 2024-10-01
fet(server):recalc build status

yuval-hazaz:
part of https://github.com/amplication/amplication/issues/8997

## PR Details

Recalculate the build status of old builds with lazy data migration.
The default value of the new field is Unknown - so use the @FieldResolver to recalculate and update the build status when the status is Unknown, otherwise return the status already saved in the DB

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -1344,7 +1344,12 @@ export class BuildService {
     if (build.status !== EnumBuildStatus.Unknown)
       return build.status as EnumBuildStatus;
 
-    if (!build.action?.steps?.length) return EnumBuildStatus.Invalid;
+    if (!build.action?.steps?.length) {
+      this.logger.error(
+        `calcBuildStatus: Could not find steps for build with id ${buildId}`
+      );
+      return EnumBuildStatus.Invalid;
+    }
     const steps = build.action.steps;
 
     if (steps.every((step) => step.status === EnumActionStepStatus.Success)) {
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -1331,4 +1331,45 @@ export class BuildService {
       message
     );
   }
+
+  async calcBuildStatus(buildId: string): Promise<EnumBuildStatus> {
+    const build = await this.prisma.build.findUnique({
+      where: {
+        id: buildId,
+      },
+      include: ACTION_INCLUDE,
+    });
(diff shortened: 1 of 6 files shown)


packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -1331,4 +1331,45 @@ export class BuildService {
...
+        id: buildId,
+      },
+      include: ACTION_INCLUDE,
+    });
+    //if the build status is not unknown, return the status
+    //the function recalculates the status only if the status is unknown
+    if (build.status !== EnumBuildStatus.Unknown)
+      return build.status as EnumBuildStatus;
+
+    if (!build.action?.steps?.length) return EnumBuildStatus.Invalid;
```
mulygottlieb, line 1347: In this case (Invalid) - don't we want to update in the DB? Is there a chance it will stop being invalid later on?

packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -1331,4 +1331,45 @@ export class BuildService {
...
+    if (steps.some((step) => step.status === EnumActionStepStatus.Failed)) {
+      await this.updateBuildStatuses(
+        buildId,
+        EnumBuildStatus.Failed,
+        EnumBuildGitStatus.Failed
+      );
+      return EnumBuildStatus.Failed;
+    }
+
+    //since the build.status is unknown, it means this is a old record and we need to update the status with failed
```
mulygottlieb, line 1367: What about In Progress?

packages/amplication-server/src/core/build/build.service.ts
yuval-hazaz: it is a great question. Since we don't really expect this scenario I cant tell when it will happen and because of that I prefer not to update the DB and allow it to be "fixed" some other way

packages/amplication-server/src/core/build/build.service.ts
yuval-hazaz: I dont see a case where something is in status "in progress" that is actually in progress - it will probably be stacked in this status, and I prefer to close it. 
If it is really in progress, so the status will be updated once the process completes as expected in a normal build

packages/amplication-server/src/core/build/build.service.ts
mulygottlieb: Sounds good. In that case, the only additional suggestion I have is writing some warning log (or analytics event?) with the buildId so we can potentially track it later to see if/when it happened (Invalid)

mulygottlieb (approved):

Issue #9019, opened 2024-10-02
Publish dsg-utils to NPM to share common utils between versions of DSG


yuval-hazaz:
### Feature description

Publish dsg-utils to NPM to share common utils between versions of DSG

mulygottlieb:
Passed QA:
![Image](https://github.com/user-attachments/assets/f8bdc5fa-f117-46f7-926c-23b79d2c4f0b)

PR #9023, opened 2024-10-02
chore(dsg-utils):move utils from DSG to dsg-utils

yuval-hazaz:
part of: https://github.com/amplication/amplication/issues/9019

## PR Details

move common utils from DSG to dsg-utils library

## PR Checklist


libs/util/dsg-utils/src/logger/logging.ts
```diff
@@ -0,0 +1,16 @@
+import { Logger, LogLevel } from "@amplication/util-logging";
+
+//const { LOG_LEVEL, NODE_ENV } = process.env;
+
+const LOG_LEVEL = process.env["LOG_LEVEL"] || "INFO";
+const NODE_ENV = process.env["NODE_ENV"] || "Development";
```
mulygottlieb, line 6: Note that you need the matching PR in cluster-configuration to add the NODE_ENV to the DSG:

https://github.com/amplication/amplication-cluster-configuration/blob/main/production/applications/data-service-generator/values.yaml

packages/data-service-generator/package.json
mulygottlieb, line 1: I think you are missing to specify the dependency for dsg-utils here

mulygottlieb (approved):

libs/util/dsg-utils/src/logger/logging.ts
yuval-hazaz, line 6:
```suggestion
const NODE_ENV = process.env["NODE_ENV"] || "production";
```

Issue #9005, opened 2024-10-01
Build the infrastructure for an indication / notification system


PazYanoverr:
This system should gather all the platform updates - new template version, new plugin version, new PP, new PP version, (new user?) etc. so this system can be used to add "Tech Debt" screen with all the resources that needed attention, add indications in Service Catalog, Templates Overview, Plugins screens and more.

PR #9034, opened 2024-10-04
feat(server):trigger alerts for new versions of service templates

yuval-hazaz:
Part of : https://github.com/amplication/amplication/issues/9005

## PR Details

Trigger alerts for all services based on their template when a new template version is pubishsed

## PR Checklist


packages/amplication-server/src/core/outdatedVersionAlert/outdatedVersionAlert.service.ts
```diff
@@ -50,4 +52,48 @@ export class OutdatedVersionAlertService {
...
   }
+
+  async triggerAlertsForTemplateVersion(
+    templateResourceId: string,
+    outdatedVersion: string,
+    latestVersion: string
+  ) {
+    const template = await this.resourceService.resource({
+      where: {
+        id: templateResourceId,
```
mulygottlieb, line 63: I would add to the `where` also a validation on the resource type to verify it's a template and not another type of resource

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -1552,10 +1552,28 @@ export class ResourceService {
...
+        args.where.project.id,
+        serviceTemplateId
+      );
+    }
+
     return this.prisma.resource.findMany({
       ...args,
       where: {
-        ...args.where,
+        ...where,
```
mulygottlieb, line 1575: you don't need the `delete` in line 1556  if you are using ...where here (and not using args.where)

packages/amplication-server/src/core/resource/resource.service.ts
```diff
@@ -1552,10 +1552,28 @@ export class ResourceService {
...
+        id: resourceIds ? { in: resourceIds } : where.id,
```
mulygottlieb, line 1576: This method should return multiple resources, so adding `id: where.id` (in the case that `serviceTemplateId` was not specified) doesn't make sense to me

yuval-hazaz: the idea is that we override the provided value for where. id in case we want to filter specific resources by id, otherwise we use the original value of where.id which probably is undefined but we dont really know or care

mulygottlieb (approved):

Issue #9146, opened 2024-10-22
datagrid columns selector


yuval-hazaz:
### Feature description

datagrid columns selector

mulygottlieb:
Passed QA
![Image](https://github.com/user-attachments/assets/7c6ef59d-45e7-4f71-a5f4-c3feef96f4d2)

![Image](https://github.com/user-attachments/assets/c3568942-d8c6-4e1a-a29f-8312ac765164)

PR #9147, opened 2024-10-22
feat(design-system):data grid columns filter

yuval-hazaz:
Close: https://github.com/amplication/amplication/issues/9146

## PR Details

Add columns selector for data grid + use it on catalog and template list

image

Align the style of the git org selector 

image

## PR Checklist

### DIFF
--- a/libs/ui/design-system/src/lib/components/DataGrid/DataGrid.tsx
+++ b/libs/ui/design-system/src/lib/components/DataGrid/DataGrid.tsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useMemo, useState } from "react";
 import "react-data-grid/lib/styles.css";
 
 import ReactDataGrid, {
-  ColumnOrColumnGroup,
+  Column,
   DataGridProps,
   RenderSortStatusProps,
   RowsChangeData,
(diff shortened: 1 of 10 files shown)


libs/ui/design-system/src/lib/components/DataGrid/DataGrid.tsx
```diff
@@ -65,6 +66,11 @@ export function DataGrid<T>({
   const [sortColumns, setSortColumns] = useState<SortColumn[]>([]);
   const [rows, setRows] = useState<T[]>(incomingRows);
 
+  const visibleColumns = useMemo(
+    () => columns.filter((c) => !c.hidden),
```
mulygottlieb, line 70: If "hidden" is undefined (it is optional), wouldn't `!c.hidden` be truth-y ? Is that what you mean to happen?

mulygottlieb (approved):

libs/ui/design-system/src/lib/components/DataGrid/DataGrid.tsx
yuval-hazaz: undefind === not hidden === visible

Issue #9229, opened 2024-11-03
Handle alerts after publishing a new template version (including in-product notifications and emails)


PazYanoverr:

mulygottlieb:
Passed QA:
Published a new version of a template (that already has a service created from it):
![Image](https://github.com/user-attachments/assets/2abd09c4-aa16-49dc-b309-52f1b5455d85)
Click on the alert takes you to the right place:
![Image](https://github.com/user-attachments/assets/e5e1720c-e75b-4a1e-be4c-66cd8c2e8af3)

I also got an email:
![Image](https://github.com/user-attachments/assets/2dfe7861-eca2-4dd9-b9b4-3c7d127953e6)
And the link from the email also takes you to the same right place ✅

Issue #9230, opened 2024-11-03
Handle alerts after publishing a new PP version (including in-product notifications and emails)


PazYanoverr:

mulygottlieb:
Passed QA ✅

Created a new version of a plugin that is already installed on a service, got an alert as soon as publishing it:
![Image](https://github.com/user-attachments/assets/c70b4342-7037-44b2-b5f1-e97784b10616)
The button in the alert takes you to the correct place:
![Image](https://github.com/user-attachments/assets/786b5a14-639f-4d07-9102-f657438d4d61)


I also got an email alert:
![Image](https://github.com/user-attachments/assets/d4d149d7-39b6-4b59-a5ec-3ec42558bc99)
And the button on the email alert also takes you to the correct page as expected

PR #9231, opened 2024-11-03
feat(server/notification-service): add novu notifications

morhag90:
Close: #9229 #9230 

## PR Details

 Add novu notifications for tech-debt alerts: 
- templates
- private-plugins 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/outdatedVersionAlert/outdatedVersionAlert.service.ts
+++ b/packages/amplication-server/src/core/outdatedVersionAlert/outdatedVersionAlert.service.ts
@@ -12,6 +12,12 @@ import { EnumResourceType } from "../resource/dto/EnumResourceType";
 import { UpdateOutdatedVersionAlertArgs } from "./dto/UpdateOutdatedVersionAlertArgs";
 import { User } from "../../models";
 import { PluginInstallationService } from "../pluginInstallation/pluginInstallation.service";
+import { KafkaProducerService } from "@amplication/util/nestjs/kafka";
+import { KAFKA_TOPICS, TechDebt } from "@amplication/schema-registry";
+import { ConfigService } from "@nestjs/config";
+import { Env } from "../../env";
+import { encryptString } from "../../util/encryptionUtil";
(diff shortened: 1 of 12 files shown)


mulygottlieb (approved):

packages/amplication-server/src/core/outdatedVersionAlert/outdatedVersionAlert.service.ts
```diff
@@ -152,6 +170,33 @@ export class OutdatedVersionAlertService {
...
+              techDebtId: alert.id,
+              envBaseUrl: this.configService.get<string>(Env.CLIENT_HOST),
+              externalId: encryptString(userId),
+              resourceType: service.resourceType,
+              projectName: project.name,
+              alertType: alert.type,
+              alertInitiator: template.name,
+            },
+          })
+          .catch((error) =>
```
yuval-hazaz, line 194: move the code that raise the kafka event into a new funciton 'raiseNotifications' and call this function from the this.create() function.

this way all types of alerts will eventually raise the notifications in the same way

also, the alerts should be sent to all users in the current workspace (and not just the current user)
later, we will add logic to select specific users based on permissions and ownership

packages/amplication-server/src/core/resourceVersion/resourceVersion.service.ts
```diff
@@ -37,7 +37,10 @@ export class ResourceVersionService {
    * create function creates a new resourceVersion for given resource in the DB
    * @returns the resourceVersion object that return after prisma.resourceVersion.create
    */
-  async create(args: CreateResourceVersionArgs): Promise<ResourceVersion> {
+  async create(
+    args: CreateResourceVersionArgs,
+    userId: string
```
yuval-hazaz, line 42: since we send the notifications for all users in the workspace and not just the current user, we dont need this parameter. 
we can leave it if you think there is another use for it

yuval-hazaz (changes requested):
great work!
see my comments

yuval-hazaz (approved):

Issue #9006, opened 2024-10-01
PP: New option- “In development”


PazYanoverr:
An “In development” option should be added per private plugin.

A few comments to take into considerations while implementing the capability- 
1. The “in-dev” version can be chosen in a service only (for the specific private plugin), and not in a service template. The “in-dev” should NOT be treated as the “latest” version.
2. When building such an “in-dev” plugin (as part of a service), the latest plugin code should be fetched from the branch

mulygottlieb:
Passed QA!

New plugins get created with dev version by default:
![Image](https://github.com/user-attachments/assets/33ba11f1-64e7-4fd4-91fc-ae63dd4562b0)

Existing plugins have "add dev version button"
![Image](https://github.com/user-attachments/assets/aa8b08ce-63f0-4292-8d22-4a55a5881820)

Add dev version works and then the button disappears:
![Image](https://github.com/user-attachments/assets/c5cd54f8-b28c-45de-bfc0-46edd417ac33)

New dev version for existing plugins gets created as the "next version-beta":
![Image](https://github.com/user-attachments/assets/cd8e33c4-157e-4df1-9dc0-43a8408b0eb5)

Dev version appears in version list to install:
![Image](https://github.com/user-attachments/assets/dc9cac94-7380-4a7f-8e5c-c2faff0d06bb)

Choosing specific version to install (1.0.0) works as expected:
![Image](https://github.com/user-attachments/assets/2fb202e3-a65d-45cd-a978-e117a3b71796)

Choosing "latest" still chooses 1.0.0 which is the latest (not taking the 1.1.0-dev as latest):
![Image](https://github.com/user-attachments/assets/21eeef23-2e41-47b4-927a-0ca2db4e5794)

Choosing the dev version downloads the default branch:
![Image](https://github.com/user-attachments/assets/0ebfd30b-2a29-478a-a4fd-357b7ec59331)
There is a minor glitch in the reporting of the name to the log (should be pluginName@main and not just "main") FYI @PazYanoverr to open a separate ticket for that.

PR #9233, opened 2024-11-04
feat(client/server): add support to pp dev version

morhag90:
Close: #9006 

## PR Details

Add support to pp dev version. 

## PR Checklist

### DIFF
--- a/packages/amplication-server/src/core/build/build.service.ts
+++ b/packages/amplication-server/src/core/build/build.service.ts
@@ -702,7 +702,7 @@ export class BuildService {
 
       const sortedEnabledVersions = privatePluginBlock.versions
         .filter(
-          (version) => version.enabled && !version.version.includes("dev")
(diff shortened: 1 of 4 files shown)


packages/amplication-server/src/core/build/build.service.ts
```diff
@@ -688,7 +688,9 @@ export class BuildService {
       if (privatePlugin.version !== "latest") {
         pluginsToDownload.push({
           pluginId: privatePlugin.pluginId,
-          pluginVersion: privatePlugin.version,
+          pluginVersion: privatePlugin.version.includes("dev")
+            ? null
+            : privatePlugin.version,
```
yuval-hazaz, line 693: to make things more clear - it is better to explicitly tell the service to use the dev version - and keep the logic for selecting the branch etc internally for the service

mulygottlieb (approved):

PR #9259, opened 2024-11-08
feat/allow import schema with existing entities

yuval-hazaz:
Close: https://github.com/amplication/amplication/issues/9260

## PR Details

When importing a schema, do the following: 

1. if an entity with the same name already exists, use it
2. for each field of the existing entity
  1.  if the field exists and is identical - do nothing 
  2. if the field does not exist on the entity, create it (only on the schema)
  3. if the field exits but is not identical - delete it and recreate it 
  4. if the field exists in the entity and not in the new schema - delete it
3. for all other entities and fields - import normally

This change does not delete entities that do not exist in the imported schema 

  

 

## PR Checklist


packages/amplication-server/src/core/prismaSchemaParser/schema-utils.ts
```diff
@@ -404,14 +412,13 @@ export function handleModelNamesCollision(
   let counter = 0;
 
   do {
+    isFormattedModelNameAlreadyTaken ||=
```
mulygottlieb, line 415: I think you have a bug here with using the ||= before you assign it directly in line 418

mulygottlieb (changes requested):
See my comment about the ||= before direct assignment

mulygottlieb (approved):

PR #9494, opened 2024-12-30
feat(server):Team Assignment to resource

yuval-hazaz:
Close:https://github.com/amplication/amplication/issues/9495

## PR Checklist

### DIFF
--- a/packages/amplication-prisma-db/prisma/schema.prisma
+++ b/packages/amplication-prisma-db/prisma/schema.prisma
@@ -127,37 +127,52 @@ model User {
 }
 
 model Role {
-  id          String    @id @default(cuid())
-  createdAt   DateTime  @default(now())
-  updatedAt   DateTime  @updatedAt
-  name        String
-  key         String
-  deletedAt   DateTime?
-  description String?
-  permissions String[]
-  workspaceId String
-  workspace   Workspace @relation(fields: [workspaceId], references: [id], onDelete: Cascade, onUpdate: Cascade)
-  teams       Team[]
+  id              String           @id @default(cuid())
+  createdAt       DateTime         @default(now())
+  updatedAt       DateTime         @updatedAt
+  name            String
+  key             String
+  deletedAt       DateTime?
+  description     String?
+  permissions     String[]
+  workspaceId     String
+  workspace       Workspace        @relation(fields: [workspaceId], references: [id], onDelete: Cascade, onUpdate: Cascade)
+  teams           Team[]
+  teamAssignments TeamAssignment[]
 
   @@unique([workspaceId, key], map: "Role.workspaceId_key_unique")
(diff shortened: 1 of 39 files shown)


packages/amplication-prisma-db/prisma/schema.prisma
```diff
@@ -127,39 +127,54 @@ model User {
...
 
+model TeamAssignment {
+  id         String   @id @default(cuid())
+  createdAt  DateTime @default(now())
+  updatedAt  DateTime @updatedAt
+  resourceId String
+  resource   Resource @relation(fields: [resourceId], references: [id], onDelete: Cascade)
+  teamId     String
+  team       Team     @relation(fields: [teamId], references: [id], onDelete: Cascade)
+  roles      Role[]
```
tommagni, line 173: a bit weird to me why we have roles on team assignments and team assignments on roles. i guess that prisma's way of allowing you to have this bidirectional access?

working in django (python), they have this same concept but you only need to put the field once and you can access from both sides.

packages/amplication-server/src/core/team/dto/WhereTeamAssignmentInput.ts
```diff
@@ -0,0 +1,16 @@
+import { Field, InputType } from "@nestjs/graphql";
+
+@InputType({
+  isAbstract: true,
+})
+export class WhereTeamAssignmentInput {
```
tommagni, line 6: WhereTeamAssignmentInput sounds like you're trying to find a team assignment or so, but it's not receiving a team assignment id, so it's a bit confusing in first read. then i figured it's likely a prisma thing, and it really is after checking. so it now makes sense. leaving the comment here so you can see the line of thought.

packages/amplication-server/src/core/team/team.service.ts
```diff
@@ -307,21 +445,74 @@ export class TeamService {
...
+    }
+  }
+
+  async getTeamAssignmentRoles(args: WhereTeamAssignmentInput) {
+    return this.prisma.teamAssignment
       .findUnique({
         where: {
-          id: teamId,
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          teamId_resourceId: {
```
tommagni, line 512: i don't like to see this in js teamId_resourceId :) but i guess it's some convention/prisma thing so i'm ok with it

packages/generator-blueprints/src/models.ts
```diff
@@ -1825,6 +1828,12 @@ export type MutationAddRolesToTeamArgs = {
 };
 
 
+export type MutationAddRolesToTeamAssignmentArgs = {
```
tommagni, line 1831: there seems to be a duplication between this file and: data-service-generator, packages/amplication-client/src/models.ts and the cli etc, i'd consider sharing them in one place.

tommagni:
great quality! see comments :)

